PR #96, opened 2025-02-03
Add initial type-C data structures and messages

RobertZ2011:
Copy of https://github.com/OpenDevicePartnership/embedded-services/pull/32/, needed to fix a conflict with main but the web-editor wouldn't work and I couldn't pull after the rename.

### DIFF
--- a/embedded-service/src/type_c/controller.rs
+++ b/embedded-service/src/type_c/controller.rs
@@ -1,0 +1,233 @@
+//! PD controller related code
+use embassy_sync::blocking_mutex::raw::NoopRawMutex;
+use embassy_sync::channel::Channel;
+use embassy_sync::once_lock::OnceLock;
+use embassy_time::{with_timeout, Duration};
+
+use super::ucsi::lpm;
+use super::{ControllerId, Error, PortId};
+use crate::intrusive_list;
+
+/// PD controller command-specific data
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum InternalCommandData {
+    /// Reset the PD controller
+    Reset,
+}
+
+/// PD controller command
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Command {
+    /// Controller specific command
+    Controller(InternalCommandData),
+    /// UCSI command passthrough
+    Lpm(lpm::Command),
+}
+
+/// Controller-specific response data
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum InternalResponseData {
+    /// Command complete
+    Complete,
+}
+
+impl Into<Result<InternalResponseData, Error>> for InternalResponseData {
+    fn into(self) -> Result<InternalResponseData, Error> {
+        Ok(self)
+    }
+}
+
+/// Response for controller-specific commands
+pub type InternalResponse = Result<InternalResponseData, Error>;
+
+/// PD controller command response
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Response {
+    /// Controller response
+    Controller(InternalResponse),
+    /// UCSI response passthrough
+    Lpm(lpm::Response),
+}
+
+/// PD controller
+pub struct Controller<'a> {
+    node: intrusive_list::Node,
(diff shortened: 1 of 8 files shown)


embedded-service/src/type_c/controller.rs
```diff
@@ -0,0 +1,205 @@
...
+            ports,
+            command: Channel::new(),
+            response: Channel::new(),
+        }
+    }
+
+    /// Send a command to this controller
+    pub async fn send_command(&self, command: Command) -> Response {
+        self.command.send(command).await;
+        self.response.receive().await
```
gjpmsft, line 85: should response have timeout?

gjpmsft (approved):

embedded-service/src/type_c/controller.rs
```diff
@@ -0,0 +1,205 @@
...
+pub enum InternalResponseData {
+    /// Command complete
+    Complete,
+}
+
+impl Into<Result<InternalResponseData, Error>> for InternalResponseData {
+    fn into(self) -> Result<InternalResponseData, Error> {
+        Ok(self)
+    }
+}
```
felipebalbi, line 40: Looks a bit hackish. `Error` will never be used, maybe it should be `Infallible`. Also, `Into` is not supposed to produce a `Result`, if you need `Result`, you should use `TryInto` instead. Could you clarify why this is needed?

embedded-service/src/type_c/controller.rs
felipebalbi: sounds like a policy decision that can be left up to the upper layer? Any future can be wrapped with `with_timeout()` to get what you want.

embedded-service/src/type_c/controller.rs
```diff
@@ -0,0 +1,205 @@
...
+                controller.has_port(port_id)
+            } else {
+                false
+            }
+        })
+        .map_or(Error::InvalidPort.into(), Ok)?;
+
+    match node
+        .data::<Controller>()
+        .unwrap()
```
felipebalbi, line 190: this can panic. How about using `?` instead?

embedded-service/src/type_c/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Invalid parameters
+    InvalidParams,
+    /// Incompatible partner
+    IncompatiblePartner,
+    /// CC communication error,
+    CcCommunication,
+    /// Failed due to dead battery condition
+    DeadBattery,
+    /// Contract negociation failed
+    ContratctNegociation,
```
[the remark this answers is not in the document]
RobertZ2011: Done.

embedded-service/src/type_c/controller.rs
RobertZ2011: I think it's a good idea to introduce a timeout argument here so that it's not possible to accidentally forget it. I added a default timeout constant as well and we can refine values for the individual commands as the rest of the code gets fleshed out.

embedded-service/src/type_c/controller.rs
RobertZ2011: Done

embedded-service/src/type_c/controller.rs
RobertZ2011: Yeah, I can see how weird this looks. This was meant to be the counterpart to the `Into<Result<T, Error>>` implementation. I added that implementation because I think that `Error::Value.into()` looks cleaner than `Err(Error::Value)`, particularly when combined with things like `map_or`. But using `value.into()` instead of `Ok(value)` isn't as useful as I thought it was. I've removed this implementation.

felipebalbi: It's still in the `diff`, no?

RobertZ2011: Oops, the change must've gotten lost. Should be gone now.

embedded-service/src/type_c/controller.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// UCSI response passthrough
+    Lpm(lpm::Response),
+}
+
+/// PD controller
+pub struct Controller<'a> {
+    node: intrusive_list::Node,
+    id: ControllerId,
+    ports: &'a [PortId],
+    command: Channel<NoopRawMutex, Command, 1>,
```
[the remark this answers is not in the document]
RobertZ2011: Ultimately only the type-c service will end up directly sending commands to a controller. Everything else will go through the type-c service which will handle the actual communication with the controllers.

felipebalbi (approved):

PR #100, opened 2025-02-05
Create initial power policy implementation

RobertZ2011:
This PR defines the power policy service and provides an initial implementation that handles selecting the highest power sink.


embedded-service/src/power/policy/device.rs
```diff
@@ -0,0 +1,318 @@
...
+        self.response.receive().await
+    }
+
+    /// Connect this device as a sink
+    pub async fn connect_sink(&self, capability: PowerCapability) -> Result<(), Error> {
+        info!("Device {} connecting sink", self.id.0);
+
+        let _ = self
+            .execute_device_request(RequestData::ConnectSink(capability))
+            .await?;
```
felipebalbi, line 138: Why are you ignoring the `ResponseData`?

embedded-service/src/power/policy/device.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        {
+            let mut lock = self.state.lock().await;
+            let state = lock.deref_mut();
+            if state.state != State::Attached {
+                warn!("Received connect sink request for device that is not attached");
+            }
+
+            state.state = State::Sink(capability);
+        }
```
felipebalbi, line 148: The extra scope seems unnecessary. Additionally, if the current state is incorrect, should you produce an `Err` variant?

embedded-service/src/power/policy/device.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        {
+            let mut lock = self.state.lock().await;
+            let state = lock.deref_mut();
+            if state.state != State::Attached {
+                warn!("Received connect source request for device that is not attached");
+            }
+
+            state.state = State::Source(capability);
+        }
```
felipebalbi, line 168: ditto.

embedded-service/src/power/policy/device.rs
```diff
@@ -0,0 +1,318 @@
...
+
+            if !matches!(state.state, State::Sink(_)) && !matches!(state.state, State::Source(_)) {
+                warn!(
+                    "Disconnect request for device that is not sourcing or sinking, state: {:?}",
+                    state.state
+                );
+            }
+
+            state.state = State::Attached;
+        }
```
felipebalbi, line 190: ditto, and because there's so much duplication, it seems like you could extract into a helper that takes a closure argument. Something along the lines of:

```rust
    fn validate<F>(self, state: State, f: F) -> T
    where
        F: FnOnce(State) -> T,
    {
        f(state)
    }

    fn foo() {
        let mut lock = self.state.lock().await;
        let state = lock.deref_mut();

        state.state = self.validate(state, |state| {
            if !matches!(state.state, State::Sink(_)) && !matches!(state.state, State::Source(_)) {
                warn!(
                    "Disconnect request for device that is not sourcing or sinking, state: {:?}",
                    state.state
                );
                Err(Error::InvalidStateTransition(state.state))
            } else {
                Ok(State::Attached);
            }
        })?;
    }
```

embedded-service/src/power/policy/device.rs
```diff
@@ -0,0 +1,318 @@
...
+    /// Provide access to the power policy service
+    pub fn policy<'a>(&'a self) -> PolicyInterface<'a> {
+        PolicyInterface(self)
+    }
+}
+
+/// Struct that provides functions to send requests to the power policy from a specific device
+pub struct PolicyInterface<'a>(&'a Device);
+
+impl PolicyInterface<'_> {
```
felipebalbi, line 213: I see an option to encode the state as typestate, then compiler can prove valid state transitions for you. That being said, it might be difficult to work with resulting API, because you will need to consume `self` in order to change state.

embedded-service/src/power/policy/device.rs
RobertZ2011: Done. Added a `complete_or_err` function to `ResponseData` to make it easy to convert it into a result.

embedded-service/src/power/policy/device.rs
RobertZ2011: I created two wrapper types: `state_machine::Device` and `state_machine::Policy` that use type states to control the actions and transitions available both to the power policy and individual devices. Wrappers are produced by the `try_*_state_machine` functions which return results to allow the caller to decide how to handle encountering an unexpected state. This should make working with each side less error prone without needing to transform the type of the overall device.

embedded-service/src/power/policy/device.rs
RobertZ2011: This was spillover from the implementation of the `notify_*` functions. Since those are intended to be called by the device driver I made those checks warnings on the assumption that we should just trust the driver and accept what it thinks the current state should be. But that doesn't apply to the policy-initiated request. Superseded by the type state changes.

power-policy/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// Send a notification with the comms service
+    async fn comms_notify(&self, message: CommsMessage) {
+        let _ = self
+            .tp
+            .send(comms::EndpointID::Internal(comms::Internal::Battery), &message)
+            .await;
+    }
+
+    /// Determines and connects the best sink
+    async fn select_sink(&self) -> Result<(), Error> {
```
jerrysxie, line 122: Would it be possible to refactor this function a bit to make it more clear? Maybe break it up into sub-functions? It is very tough to read.

Let me see if I grasp what it is trying to do:
1. Searching through existing devices to find the best sink if it is the disconnecting sink ignore
2. If the best sink is same and it is still available, then okay. Else disconnect the sink?
3. Connect the new best sink?

The second part doesn't make sense to me. We are not considering the sink being disconnected in the first search. Then it wouldn't be possible for the best sink to be the device being disconnected, no?

Wouldn't the logic be simpler to record the current best sink, if it is being removed, then find the best sink in the remaining sink, disconnect the current sink and connect the new best sink. Else just keep the best sink as it is.

jerrysxie: It seems like the logic in this function is this way, because it is trying to handle lots of different cases. Can we simplify this a bit? It might be more verbose to handle different case separately, but it might be more readable.

power-policy/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+
+        info!("Best sink: {:#?}", best_sink);
+        if let Some(best_sink) = best_sink {
+            // See if we need to disconnect the current sink
+            if let Some(current_sink) = state.current_sink_state {
+                if let Ok(sink) = self
+                    .context
+                    .get_device(current_sink.device_id)
+                    .await?
+                    .try_policy_state_machine::<sm::Sink>()
```
jerrysxie, line 170: Why would this state transition fail if it is the current sink?

embedded-service/src/power/policy/policy.rs
```diff
@@ -0,0 +1,169 @@
...
+pub enum ResponseData {
+    /// The request was completed successfully
+    Complete,
+}
+
+impl ResponseData {
+    /// Returns an InvalidResponse error if the response is not complete
+    pub fn complete_or_err(self) -> Result<(), Error> {
+        match self {
+            ResponseData::Complete => Ok(()),
```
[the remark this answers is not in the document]
RobertZ2011: Yes if more values are added to `ResponseData`. No policy requests return data so this is currently the only value in the enum. I did try marking the enum `#[non_exhaustive]` and adding a wildcard match arm, but I was still getting warnings about the wildcard being unreachable so I decided to leave it as is for now.

power-policy/src/lib.rs
RobertZ2011: I split the function into two subfunctions and did some unnesting of braces to clarify things. Disconnecting the current sink is done because it may still be present but we're switching to a higher power charger.

power-policy/src/lib.rs
RobertZ2011: If the current sink has been unplugged then we won't need to disconnect it.

embedded-service/src/power/policy/device.rs
```diff
@@ -0,0 +1,424 @@
...
+            self.disconnect_internal().await?;
+            Ok(Device::new(self.device))
+        }
+    }
+
+    /// Policy state machine control
+    pub struct Policy<'a, S: Kind> {
+        device: &'a super::Device,
+        _state: core::marker::PhantomData<S>,
+    }
```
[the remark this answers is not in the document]
RobertZ2011: I spent some time trying out this approach, but I think the current method is better. We'd need at least three power role values (NoRole, Source, Sink) and three power state values (Detached (not present), Idle (present but not in use), Connected (present and in use)). But this allows for states that are valid to the compiler, but not to the application (e.g. (Detached, Source) and (Detached, Sink)).  I also don't think the (Idle, Source) and (Idle, Sink) states make sense because the device doesn't really have a role until the power policy decides what it wants. It would be possible to use those states to indicate a device that can only act as a source and a device that can only act as a sink, but we'd also need a way to indicate dual-role devices and things start getting really complicated.

RobertZ2011:
Also updated the documentation to explain the design of the service better and point out the core structs there. I also did a rename. The source and sink nomenclature was borrowed from the PD standard and I realized that it can be easy to confuse things. Renamed source->provider, sink->consumer and attached->idle.

jerrysxie (approved):

embedded-service/src/power/policy/action/device.rs
```diff
@@ -0,0 +1,97 @@
...
+        Self {
+            device,
+            _state: core::marker::PhantomData,
+        }
+    }
+
+    /// Detach the device
+    pub async fn detach(self) -> Result<Device<'a, Detached>, Error> {
+        info!("Received detach from device {}", self.device.id().0);
+        self.device.set_state(device::State::Detached).await;
```
felipebalbi, line 24: This line implies that you're tracking state via typestate and via a dedicated field within `Device`. Why the duplication?

embedded-service/src/power/policy/action/policy.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+impl<'a> Policy<'a, ConnectedProvider> {
+    /// Disconnect this device
+    pub async fn disconnect(self) -> Result<Policy<'a, Idle>, Error> {
+        self.disconnect_internal().await?;
+        Ok(Policy::new(self.device))
+    }
+}
```
felipebalbi, line 78: Duplicated code here. Maybe add a small trait that both `ConnectedConsumer` and `ConnectedProvider` implement?

power-policy-service/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub fn create() -> Option<Self> {
+        Some(Self {
+            context: policy::ContextToken::create()?,
+            state: Mutex::new(InternalState::new()),
+            tp: comms::Endpoint::uninit(comms::EndpointID::Internal(comms::Internal::Power)),
+        })
+    }
+
+    async fn process_notify_attach(&self) -> Result<(), Error> {
+        self.context.send_response(Ok(policy::ResponseData::Complete)).await;
```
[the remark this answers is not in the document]
RobertZ2011: Yeah, the provider logic will fill out the other empty handler. This one will remain just because there's not anything to do aside from acknowledging the notification.

power-policy-service/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+            {
+                info!(
+                    "Device {}, disconnecting current consumer",
+                    current_consumer.device_id.0
+                );
+                consumer.disconnect().await?;
+            }
+
+            self.comms_notify(CommsMessage {
+                data: CommsData::ConsumerConnected(current_consumer.device_id, new_consumer.power_capability),
```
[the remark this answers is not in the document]
RobertZ2011: Done, good catch!

embedded-service/src/power/policy/action/policy.rs
RobertZ2011: I tried this, but the traits end up a bit more complicated than I'd like. Given that the duplication is small I'm going to leave it as is.

embedded-service/src/power/policy/action/device.rs
RobertZ2011: This is because we can't change the overall type of `Device` since it needs to live in something like a `OnceLock` or `StaticCell` and because the power policy implementation needs a fixed type to work with. So these action types provide checking on a local level and the field keeps track of the state on a global level without having to put the type state on `Device` itself.

embedded-service/src/power/policy/action/device.rs
```diff
@@ -0,0 +1,97 @@
...
+        self.device.update_consumer_capability(None).await;
+        policy::send_request(self.device.id(), policy::RequestData::NotifyDetached)
+            .await?
+            .complete_or_err()?;
+        Ok(Device::new(self.device))
+    }
```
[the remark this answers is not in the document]
RobertZ2011: This transition is valid. It simplifies the device recover logic a bit to always have the `detach` action available. I updated the docs to include this transition.

embedded-service/src/power/policy/action/device.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub async fn request_provider_power_capability(&self, capability: PowerCapability) -> Result<(), Error> {
+        info!("Request provide from device {}, {:#?}", self.device.id().0, capability);
+        policy::send_request(
+            self.device.id(),
+            policy::RequestData::RequestProviderCapability(capability),
+        )
+        .await?
+        .complete_or_err()
+    }
+}
```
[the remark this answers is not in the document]
RobertZ2011: The transition from `Idle` is driven from the policy side not the device side. The corresponding policy `Idle` action contains those functions.

RobertZ2011:
Fixed a few small bugs and added types to contain any action after my experiences with the PD controller integration.

power-policy-service/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+                        Some(ConsumerState {
+                            device_id: device.id(),
+                            power_capability: available,
+                        })
+                    } else {
+                        best_consumer
+                    }
+                }
+            };
+        }
```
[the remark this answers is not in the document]
RobertZ2011: Definitely a fold, but I don't think it's worth it in this case. We'd need space to store a collection of the available power numbers but the amount of devices isn't fixed and I'd like to avoid limiting it with a constant in order to use something like a heapless vec.

PR #101, opened 2025-02-06
Cfu Service

madeleyneVaca:
Defines a Cfu Client service crate, integrates the necessary hooks in the embedded-service crate, and creates an example for a cfu-client.

This PR also stubs out a Cfu host service based on the host traits from the pulled in embedded-cfu-protocol crate. Most EC's will be clients and not hosts for CFU offers.

### DIFF
--- a/cfu-service/src/lib.rs
+++ b/cfu-service/src/lib.rs
@@ -1,0 +1,88 @@
+#![no_std]
+use embassy_sync::once_lock::OnceLock;
+use embedded_cfu_protocol::client::CfuReceiveContent;
+use embedded_cfu_protocol::protocol_definitions::*;
+use embedded_services::cfu::component::*;
+use embedded_services::cfu::{CfuError, ContextToken};
+use embedded_services::{comms, error, info};
+
+pub mod host;
+
+pub struct CfuClient {
+    /// Cfu Client context
+    context: ContextToken,
+    /// Comms endpoint
+    tp: comms::Endpoint,
+}
+
+/// use default "do-nothing" implementations
+impl<T, C, E: Default> CfuReceiveContent<T, C, E> for CfuClient {}
+
+impl CfuClient {
+    /// Create a new Cfu Client
+    pub fn create() -> Option<Self> {
+        Some(Self {
+            context: ContextToken::create()?,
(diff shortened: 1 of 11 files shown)


cfu-service/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// use impl Future or Send so that these can be async?
+pub trait Cfu<W, I>: CfuUpdateContent<W> + CfuStates<W>
+where
+    W: CfuWriter,
+    I: CfuImage,
+{
+    /// Goes through the offer list and returns a slice of offer responses
+    fn process_cfu_offers(
```
RobertZ2011, line 20: Can you add an example?

felipebalbi (approved):

embedded-service/src/cfu/component.rs
```diff
@@ -0,0 +1,315 @@
...
+/// CFU Request types and necessary data
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum RequestData {
+    /// Request for component's current FW version
+    FwVersionRequest,
+    /// Contains an offer for the component to evaluate
+    GiveOffer(FwUpdateOfferCommand),
+    /// Contains bytes for an accepted fw offer
+    GiveContent(FwUpdateContentCommand<&'static [u8]>),
```
RobertZ2011, line 68: This is where you'll need to start using `SharedRef`. But I'm not sure what the best way to make this generic with other slice is. The protocol crate might have to define its own trait that `SharedRef` implements. I think we could also provide an implementation of that trait for `RefCell` too.

madeleyneVaca: for this we know that the data length is going to be up to 52 bytes and any unused bytes are zero-padded so how about I change from using the generic T: AsRef<[u8]> to this:


```suggestion
    pub struct FwUpdateContentCommand> {
        pub header: FwUpdateContentHeader,
        pub data: [u8; DEFAULTDATALENGTH],
    }
```

RobertZ2011: Sounds good to me.

RobertZ2011 (approved):

PR #105, opened 2025-02-13
embedded-service: refactor logging macros

RobertZ2011:
Split up logging macros implementations along feature lines. Using these macros as is in dependent crates can leak the `defmt` and `log` features resulting in warnings if it only supports one of the features.


jerrysxie (approved):

felipebalbi (changes requested):
You need to update `doc` workflow too.

felipebalbi (approved):

PR #107, opened 2025-02-18
Initial eSPI service using the eSPI HAL + initial eSPI service example

jerrysxie:
Adding initial eSPI service:

- Only peripheral channel reads are supported

- Directly using the embassy-imxrt HAL because there not really a standard behavior for an eSPI device. The current driver behavior seems to be strictly particular to NXP hardware.


espi-service/src/espi_service.rs
```diff
@@ -0,0 +1,322 @@
...
+            super::TimeAlarmMessage::AcTimeVal(ac_time_val) => memory_map.tas.ac_time_val = *ac_time_val,
+            super::TimeAlarmMessage::DcTimeVal(dc_time_val) => memory_map.tas.dc_time_val = *dc_time_val,
+        }
+    });
+}
+
+use embassy_imxrt::espi;
+
+#[embassy_executor::task]
+pub async fn espi_service(mut espi: espi::Espi<'static>, memory_map_buffer: &'static mut [u8]) {
```
RobertZ2011, line 262: Definitely make an espi trait and change this function to take `&mut impl Espi`. Unfortunately, embassy tasks cannot be generic, but you should still be able to put this in a function and either use a macro to paste in the types or just leave it to the user to create their own embassy task that calls `espi_service`.

jerrysxie: Unfortunately there is not really a standard eSPI behavior API. This eSPI implementation seems to be very NXP specific; that is why I did not create a generic trait. Until we get smarter about eSPI, we might have to live with linking directly to the hardware specific HAL for now.

.github/workflows/check.yml
```diff
@@ -150,6 +150,31 @@ jobs:
...
+        workdir: [ "espi-service" ]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+          ref: ${{ matrix.commit }}
+      - name: Install nightly
+        uses: dtolnay/rust-toolchain@nightly
+      - name: cargo doc
+        run: cargo doc --target thumbv8m.main-none-eabihf --no-deps --all-features
```
jerrysxie, line 173: @felipebalbi it seems like for these few workflows, we have to manually specify the target for them to work correctly. Is this the best way to do it?

espi-service/src/espi_service.rs
felipebalbi: The trait will be hard for this. NXP's implementation handles a lot of the communication in HW, while Nuvoton doesn't. Nuvoton's implementation seems more natural and more friendly to a trait, IMO.

RobertZ2011 (approved):

.github/workflows/check.yml
felipebalbi: odd, doc should compile without a target. Then again, I guess the HAL is thumbv8m only?

jerrysxie: Yes, HAL is arch specific due to its dependency on cortex-m crate.

.github/workflows/check.yml
```diff
@@ -234,3 +285,36 @@ jobs:
...
+                ref: ${{ matrix.commit }}
+            - name: Install ${{ matrix.msrv }}
+              uses: dtolnay/rust-toolchain@master
+              with:
+                toolchain: ${{ matrix.msrv }}
+            - name: cargo +${{ matrix.msrv }} check
+              run: |
+                cargo check -F log --target thumbv8m.main-none-eabihf
+                cargo check -F defmt --target thumbv8m.main-none-eabihf
+              working-directory: ${{ matrix.workdir }}
```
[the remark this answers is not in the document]
jerrysxie: Can you clarify? Are you thinking that we don't need to run MSRV workflow for eSPI service or why do we need a MSRV workflow for specific for espi-service?

If it is the second question, it is due to the dependency on cortex-m crate. It gets a compile error without specifying the target.

If it is the first question, since it is depending on embassy-imxrt directly, it would be nice if they have the same minimum version of rustc.

espi-service/src/espi_service.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn receive(&self, message: &comms::Message) {
+        if let Some(msg) = message.data.get::<ec_type::message::CapabilitiesMessage>() {
+            update_capabilities_section(msg);
+        } else if let Some(msg) = message.data.get::<ec_type::message::BatteryMessage>() {
+            update_battery_section(msg);
+        } else if let Some(msg) = message.data.get::<ec_type::message::ThermalMessage>() {
+            update_thermal_section(msg);
+        } else if let Some(msg) = message.data.get::<ec_type::message::TimeAlarmMessage>() {
+            update_time_alarm_section(msg);
+        }
```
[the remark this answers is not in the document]
jerrysxie: That is how `comms` is implemented to rely on polymorphism to handle any kind of message between the different endpoints. I don't know what the code size impact this is, but we have to been taking advantage of this in multiple EC services.

```
/// Message to receive
#[derive(Copy, Clone, Debug)]
pub struct Message<'a> {
    /// where this message came from
    pub from: EndpointID,

    /// where this message is going
    pub to: EndpointID,

    /// message content
    pub data: Data<'a>,
}

/// Data reference -- generalized such that any stack variable can be transmitted "in place" as needed
#[derive(Copy, Clone, Debug)]
pub struct Data<'a> {
    contents: &'a dyn Any,
}

impl<'a> Data<'a> {

...

    /// Attempt to retrieve data as type T -- None if incorrect type
    pub fn get<T: Any>(&self) -> Option<&T> {
        self.contents.downcast_ref()
    }

```

RobertZ2011 (approved):

espi-service/src/espi_service.rs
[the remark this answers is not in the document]
felipebalbi: I'm pretty certain that `dyn Any` makes everything larger :-) But okay, we can change it later, I guess.

espi-service/src/espi_service.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+static ESPI_SERVICE: OnceLock<Service> = OnceLock::new();
+static MEMORY_MAP: OnceLock<Mutex<ThreadModeRawMutex, RefCell<&mut ec_type::structure::ECMemory>>> = OnceLock::new();
```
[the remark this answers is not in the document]
felipebalbi: But in that case, do you need interior mutability at all? Couldn't you write it like so:

```rust
pub struct Service<'a> {
    pub endpoint: comms::Endpoint,
    pub ec_memory: &'a mut ec_type::structure::ECMemory,
}

impl Service<'_> {
    pub fn new(ec_memory: &'static mut ec_type::structure::ECMemory) -> Self {
        Service {
            endpoint: comms::Endpoint::uninit(EndpointID::External(External::Host)),
            ec_memory),
        }
    }

    fn update_battery_section(&mut self, msg: &ec_type::message::BatteryMessage) {
        // ...
    }
}
```

And, now that I look at it in a simplified context, do we need `endpoint` and `ec_memory` to be publicly visible?

felipebalbi (commented):
Left a few extra questions.

.github/workflows/check.yml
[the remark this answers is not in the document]
jerrysxie: The only reason we needed to create a special one for msrv-espi-service, was due to the need for `--target thumbv8m.main-none-eabihf`. Without it, we got a compile error due to dependency on cortex-m crate.

examples/rt633/README.md
```diff
@@ -0,0 +1,21 @@
...
+Add uniquely named example to `src/bin` like `adc.rs`
+
+## Build
+
+`cd` to examples folder
+`cargo build --bin <example_name>` for example, `cargo build --bin adc`
+
+## Run
+
+Assuming RT685 is powered and connected to Jlink debug probe and the latest probe-rs is installed via
```
tullom, line 18: Nit: RT633s

espi-service/src/espi_service.rs
```diff
@@ -0,0 +1,214 @@
...
+            ec_type::message::BatteryMessage::RemainCap(remain_cap) => memory_map.batt.remain_cap = *remain_cap,
+            ec_type::message::BatteryMessage::PresentVolt(present_volt) => memory_map.batt.present_volt = *present_volt,
+            ec_type::message::BatteryMessage::PsrState(psr_state) => memory_map.batt.psr_state = *psr_state,
+            ec_type::message::BatteryMessage::PsrMaxOut(psr_max_out) => memory_map.batt.psr_max_out = *psr_max_out,
+            ec_type::message::BatteryMessage::PsrMaxIn(psr_max_in) => memory_map.batt.psr_max_in = *psr_max_in,
+            ec_type::message::BatteryMessage::PeakLevel(peek_level) => memory_map.batt.peak_level = *peek_level,
+            ec_type::message::BatteryMessage::PeakPower(peek_power) => memory_map.batt.peak_power = *peek_power,
+            ec_type::message::BatteryMessage::SusLevel(sus_level) => memory_map.batt.sus_level = *sus_level,
+            ec_type::message::BatteryMessage::SusPower(sus_power) => memory_map.batt.sus_power = *sus_power,
+            ec_type::message::BatteryMessage::PeakThres(peek_thres) => memory_map.batt.peak_thres = *peek_thres,
```
tullom, line 39: Nit: peak misspelled as peek

tullom (approved):
Couple of nit comments, once those are resolved it looks good to go

.github/workflows/check.yml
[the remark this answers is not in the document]
jerrysxie: I figured out the root cause of my issues with workflows for espi-service. I needed to add `.cargo/config.toml` to ensure that it compile for `thumbv8m.main-none-eabihf` target.

I did that and removed all the special espi-service workflows.

PR #110, opened 2025-02-18
First keyboard service commits

RobertZ2011:
Create a few basic message types for the keyboard service.

### DIFF
--- a/embedded-service/src/keyboard.rs
+++ b/embedded-service/src/keyboard.rs
@@ -1,0 +1,120 @@
+//! Keyboard service data types and common functionality
+use core::cell::Cell;
+
+use embassy_sync::once_lock::OnceLock;
+
+use crate::buffer::SharedRef;
+use crate::comms::{self, EndpointID, External, Internal};
+
+/// Keyboard device ID
+#[derive(Debug, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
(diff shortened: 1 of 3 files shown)


RobertZ2011:
984e1d97eff7da82618631648970048e98671339 is present in commit history just to allow checks to run until https://github.com/OpenDevicePartnership/embedded-services/pull/109 is merged.

embedded-service/src/keyboard.rs
```diff
@@ -0,0 +1,54 @@
...
+/// Top-level message struct
+#[derive(Clone)]
+pub struct Message<'a> {
+    /// Target/source device ID
+    pub device_id: DeviceId,
+    /// Message data
+    pub data: MessageData<'a>,
+}
+
+/// Broadcast a keyboard message
```
JamesHuard, line 50: would it make sense to have some kind of keyboard service configuration or provider-based registry, that automatically broadcasts to a set of specific targets, such as Host or HID depending on the config?

felipebalbi (approved):

embedded-service/src/keyboard.rs
RobertZ2011: Good idea. I've changed `broadcast_message` to use a global config to determine which endpoints are enabled as well as functions to enable specific endpoints. It should also be very easy expand this to use features if this is something we want to control at compile time.

JamesHuard (approved):

PR #112, opened 2025-02-24
Initial type-C service implementation and integration with power policy

RobertZ2011:


RobertZ2011:
Checks are currently failing due to an unrelated conflict with the `log` and `defmt` features, currently working on resolving that.

type-c-service/src/wrapper.rs
```diff
@@ -0,0 +1,279 @@
...
+use embassy_futures::select::{select, select_array, Either};
+use embedded_services::power::policy::device::{RequestData, StateKind};
+use embedded_services::power::policy::{self, action};
+use embedded_services::type_c::controller::{self, Contract, PortStatus};
+use embedded_services::type_c::event::{PortEventFlags, PortEventKind};
+use embedded_services::{error, info, intrusive_list, trace, warn};
+use embedded_usb_pd::{Error, PdError, PortId as LocalPortId};
+
+/// PD controller trait for use with wrapper struct
+pub trait Controller {
```
jerrysxie, line 17: Shouldn't these generic traits be exported into a separate create like usb-pd and be implemented in the individual driver crate instead of in the service itself? Having tps6699x implemented this there creates a hard link to the driver from the service. That kind of breaks the service hardware agnostic abstraction.

jerrysxie (changes requested):

embedded-service/src/type_c/controller.rs
```diff
@@ -127,101 +129,124 @@ pub async fn register_controller(controller: &'static impl ControllerContainer)
...
-
-    match node
-        .data::<Controller>()
-        .ok_or(Error::InvalidController)?
-        .send_command(Command::Controller(command))
-        .await
-    {
-        Response::Controller(response) => response,
-        _ => Error::InvalidResponse.into(),
+const DEFAULT_TIMEOUT: Duration = Duration::from_millis(2500);
```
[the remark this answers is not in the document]
RobertZ2011: Updated the commit message with more details. This ensures exclusive access to the internal communicating with PD controllers. So if two different type-C service implementations are included then at least one of them would error at runtime.

embedded-service/src/type_c/controller.rs
```diff
@@ -97,6 +101,12 @@ impl<'a> Device<'a> {
     pub async fn send_response(&self, response: Response) {
         self.response.send(response).await;
     }
+
+    /// Notify of a port event
+    pub async fn notify_ports(&self, events: PortEventFlags) {
+        let context = CONTEXT.get().await;
+        context.port_events.set(context.port_events.get() | events);
```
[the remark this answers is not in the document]
RobertZ2011: Nothing special needs to be done here. Each controller will notify the type-C service when they have events on any of their ports. Since there's no await once we get the context then there's no way for separate tasks to interfere with each other.

embedded-service/src/type_c/controller.rs
```diff
@@ -75,31 +75,36 @@ pub enum Response {
     Lpm(lpm::Response),
 }
 
+/// Maximum number of controller ports
+pub const MAX_CONTROLLER_PORTS: usize = 2;
```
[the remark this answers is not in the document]
RobertZ2011: I was trying to avoid having to manage slice lifetimes by statically limiting the number of ports. 2 is a common top number on the types of PD controllers we tend to use. But this code is meant to be generic as possible so I've moved back to using slices to sort port IDs.

embedded-service/src/type_c/controller.rs
```diff
@@ -75,31 +75,36 @@ pub enum Response {
...
+impl Device {
     /// Create a new PD controller struct
-    pub fn new(id: ControllerId, ports: &'a [GlobalPortId]) -> Self {
-        Self {
+    pub fn new(id: ControllerId, ports: &[GlobalPortId]) -> Result<Self, PdError> {
+        Ok(Self {
             node: intrusive_list::Node::uninit(),
             id,
-            ports,
+            ports: ports.try_into().map_err(|_| PdError::InvalidParams)?,
```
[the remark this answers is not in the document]
RobertZ2011: Yeah, this was meant to cast to the right array size. Removed as part of the refactor back to using port ID slices.

type-c-service/src/wrapper.rs
RobertZ2011: Abstraction over the hardware is provided by the service messages. This trait just provides a standard way to integrate with those messages, but it's not required. You could write a driver that uses messages directly and it would work with this service implementation. I do think this trait can be used in a more generic manner so I've moved it into embedded-service. Embedded-usb-pd is meant to contain things related to the PD spec. But this trait is specific to what we're doing with PD and not related to the PD spec.

embedded-service/src/type_c/controller.rs
```diff
@@ -79,32 +79,29 @@ pub enum Response {
...
     fn get_node(&self) -> &intrusive_list::Node {
         &self.node
     }
 }
 
-impl Device {
+impl<'a> Device<'a> {
     /// Create a new PD controller struct
-    pub fn new(id: ControllerId, ports: &[GlobalPortId]) -> Result<Self, PdError> {
+    pub fn new(id: ControllerId, ports: &'a [GlobalPortId]) -> Result<Self, PdError> {
```
jerrysxie, line 101: Do we still need to return a result here? It does not seem like we can run into an error here.

type-c-service/src/wrapper.rs
jerrysxie: I like that we have moved the trait to embedded-service, but I still feel that it makes more sense for the tps6699x driver crate to implement the `Controller` trait. I would like the service to be hardware agnostic as much as possible.

If we add support for a new USB PD driver, I would assume that we would need to implement the `Controller` trait and then we would not need any change in `type-c-service` for it to work. Then in our actual service instantiation, we can just instantiate the new driver as a `Controller` and pass it to the service. Please correct me if I am missing something.

jerrysxie (changes requested):

embedded-service/src/type_c/controller.rs
RobertZ2011: Done.

type-c-service/src/wrapper.rs
RobertZ2011: >I like that we have moved the trait to embedded-service, but I still feel that it makes more sense for the tps6699x driver crate to implement the Controller trait. I would like the service to be hardware agnostic as much as possible.

I'll move this to the tps6699x crate and place it behind a feature since the core driver functionality can be available with few dependencies. But I want this batch of PRs to land first so I'm tracking it under https://github.com/OpenDevicePartnership/embedded-services/issues/169.

>If we add support for a new USB PD driver, I would assume that we would need to implement the Controller trait and then we would not need any change in type-c-service for it to work. Then in our actual service instantiation, we can just instantiate the new driver as a Controller and pass it to the service. Please correct me if I am missing something.

Correct, this would be the easiest way. Though I wouldn't say that the controller is being passed to the service implementation. The service implementation is running in its own task and each controller instance is running in its own task. Communication between the service task and controller tasks is done through messages. The driver data flow here is `Controller` implementation -> `Wrapper` (translates between messages and trait functions) -> per-device task (processes messages and events in a loop).

jerrysxie (approved):

RobertZ2011:
Current check failures are due to a dependency of `embassy-imxrt` being unmaintained.

felipebalbi (approved):

PR #113, opened 2025-02-27
Add standard EC types

jerrysxie:
- EC internal messages

- EC internal structures

### DIFF
--- a/embedded-service/src/ec_type/message.rs
+++ b/embedded-service/src/ec_type/message.rs
@@ -1,0 +1,87 @@
+//! EC Internal Messages
+
+#[allow(missing_docs)]
+#[derive(Clone, Copy, Debug)]
+pub enum CapabilitiesMessage {
+    Events(u32),
+    FwVersion(super::structure::Version),
+    SecureState(u8),
+    BootStatus(u8),
+    FanMask(u8),
+    BatteryMask(u8),
+    TempMask(u16),
+    KeyMask(u16),
+    DebugMask(u16),
+}
+
+#[allow(missing_docs)]
+#[derive(Clone, Copy, Debug)]
+pub enum TimeAlarmMessage {
+    Events(u32),
+    Capability(u32),
+    Year(u16),
+    Month(u8),
+    Day(u8),
+    Hour(u8),
+    Minute(u8),
+    Second(u8),
+    Valid(u8),
+    Daylight(u8),
+    Res1(u8),
+    Milli(u16),
+    TimeZone(u16),
+    Res2(u16),
+    AlarmStatus(u32),
+    AcTimeVal(u32),
+    DcTimeVal(u32),
+}
+
+#[allow(missing_docs)]
+#[derive(Clone, Copy, Debug)]
+pub enum BatteryMessage {
+    Events(u32),
+    LastFullCharge(u32),
+    CycleCount(u32),
+    State(u32),
+    PresentRate(u32),
+    RemainCap(u32),
+    PresentVolt(u32),
+    PsrState(u32),
+    PsrMaxOut(u32),
+    PsrMaxIn(u32),
+    PeakLevel(u32),
+    PeakPower(u32),
+    SusLevel(u32),
+    SusPower(u32),
+    PeakThres(u32),
+    SusThres(u32),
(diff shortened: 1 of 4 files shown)


embedded-service/src/ec_type/structure.rs
```diff
@@ -0,0 +1,112 @@
+//! EC Internal Data Structures
+
+#[allow(missing_docs)]
+#[repr(C, packed)]
+#[derive(Clone, Copy)]
```
tullom, line 5: for all these structs, should we derive more default traits? Like Debug or Eq?

tullom (approved):

embedded-service/src/ec_type/message.rs
```diff
@@ -0,0 +1,87 @@
+//! EC Internal Messages
```
JamesHuard, line 1: does EC type here correspond to the eSPI/direct memory access structures-- meaning the messages and abstractions over whatever data path we expect that to be eventually (in theory any pipe)?

Trying to get a handle on if this is specific to the ACPI bounds, or if this is meant to be even more generic for messaging within the EC itself

jerrysxie: Yes, for now, it is mostly messages and structures related to eSPI/direct memory access structures for now. We are still talking to different functional subsystem teams on what other messages are needed. At for battery + MPTF + alarm, the feedback is that this is all they need for now.

The messages are intended to be more generic but it is only covered the messages needed for eSPI; the structures are only for the EC memory map.

embedded-service/src/ec_type/structure.rs
jerrysxie: I tried to avoid adding all the compiler derived traits by default due to code size concern; I typically only add them as needed. But I can be convinced.

@RobertZ2011 @felipebalbi @JamesHuard What are your thoughts on this?

philgweber (approved):
Thanks for pulling in these structure updates. Will let you know as we get closer we can start updating version as well.

embedded-service/src/ec_type/structure.rs
RobertZ2011: I'm fine adding these in as needed. Though I do also feel that these are the types of structs that will need to be printed out at some point during development so it might be worth it to do it now.

embedded-service/src/ec_type/structure.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct Version {
+    pub major: u8,
+    pub minor: u8,
+    pub spin: u8,
+    pub res0: u8,
+}
+
+#[allow(missing_docs)]
+#[repr(C, packed)]
```
RobertZ2011, line 14: I've started preferring using bincode's derivable `Encode`/`Decode traits because it'll handle endianness and we won't have to write any unsafe code.

RobertZ2011 (approved):

embedded-service/src/ec_type/message.rs
JamesHuard: Got it, thanks for the clarification.

Do we expect this to grow large enough to the point where we might want to split out message types based on target? Like battery versus thermal as their own mod (message::battery::XYZ etc.)?

embedded-service/src/ec_type/structure.rs
JamesHuard: Generally agree with Robert here-- I would err on the side of adding more debuggability up front while we're learning, and pairing down on derive traits as part of an optimization pass when needed (after we light up more features and we need the space)

JamesHuard (approved):

embedded-service/src/ec_type/message.rs
jerrysxie: That might be a possibility, but there is not enough data to say right now.

For the map structure, the plan is to auto-generate it based on a YAML file eventually.

embedded-service/src/ec_type/structure.rs
jerrysxie: Played around with bincode `Encode` and `Decode` for a bit and currently don't see the value of pulling that in for this. We will add it later if we need it.

PR #116, opened 2025-02-28
Power policy source

RobertZ2011:


power-policy-service/src/provider.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Current system provider power state
+pub enum PowerState {
+    Low,
+    High,
+}
+
+impl PowerPolicy {
+    /// Computes the total requested power considering all current providers
+    async fn compute_total_provider_power(&self, new_request: bool) -> Result<PowerState, Error> {
```
jerrysxie, line 13: I am a little bit confused by the logic here.

If there is a new request which means a new provider, we are using the low max power of all the providers and see if it is higher than high power threshold?

But then if we are in the high power state, then we have to force all the providers to high power. That seems backwards. If there is a new provider, doesn't that mean there is new power capability coming online? Is there is no additional consumer, why would that force the system into a higher power state?

jerrysxie (changes requested):

power-policy-service/src/provider.rs
```diff
@@ -0,0 +1,79 @@
...
+
+            if let Ok(action) = self
+                .context
+                .try_policy_action::<action::ConnectedProvider>(device.id())
+                .await
+            {
+                if action.power_capability().await != target_power {
+                    if let Err(_) = action.connect_provider(target_power).await {
+                        error!("Device{}: Failed to connect provider", device.id().0);
+                        // Don't return to update other devices
```
jerrysxie, line 51: Can this actually fail? Do we need to do anything when this fails?

power-policy-service/src/lib.rs
```diff
@@ -80,8 +66,9 @@ impl PowerPolicy {
         Ok(())
     }
 
-    async fn process_request_provider_power_capabilities(&self) -> Result<(), Error> {
+    async fn process_request_provider_power_capabilities(&self, device: DeviceId) -> Result<(), Error> {
         self.context.send_response(Ok(policy::ResponseData::Complete)).await;
+        self.update_providers(Some(device)).await?;
```
jerrysxie, line 71: Does `process_notify_disconnect` need to call `update_providers()` as well? From the state diagram it seems like we should.

power-policy-service/src/provider.rs
```diff
@@ -0,0 +1,79 @@
...
+
+        for device in self.context.devices().await {
+            let device = device.data::<device::Device>().ok_or(Error::InvalidDevice)?;
+
+            if device.is_provider().await {
+                num_providers += 1;
+            }
+        }
+
+        let request_low_power = num_providers * self.config.provider_unlimited.max_power_mw();
```
jerrysxie, line 31: Can we rename request_low_power to total_provided_power to match the comment?

power-policy-service/src/provider.rs
```diff
@@ -0,0 +1,79 @@
...
+            PowerState::Unlimited => self.config.provider_unlimited,
+            PowerState::Limited => self.config.provider_limited,
+        };
+
+        self.update_provider_capability(target_power).await?;
+        if let Some(new_provider) = new_provider {
+            if let Ok(action) = self.context.try_policy_action::<action::Idle>(new_provider).await {
+                action.connect_provider(target_power).await?;
+            } else {
+                error!("Device {}: Failed to connect provider", new_provider.0);
```
jerrysxie, line 73: Same here, under what circumstances can this actually fail?

jerrysxie (changes requested):

power-policy-service/src/provider.rs
jerrysxie: Thanks for the renaming and added comment, that really helps to understand this code. 👍

RobertZ2011:
Current check failures are due to a dependency of `embassy-imxrt` being unmaintained.

power-policy-service/src/provider.rs
```diff
@@ -46,34 +67,108 @@ impl PowerPolicy {
...
+                            "Device{}: Failed to connect provider, attempting to disconnect",
+                            device.id().0
+                        );
+
+                        if let Err(_) = action.disconnect().await {
+                            error!(
+                                "Device{}: Failed to disconnect provider, entering recovery mode",
+                                device.id().0
+                            );
+                            recovery = true;
```
jerrysxie, line 83: So if one device fails, then that might the entire system goes into a recovery state?

power-policy-service/src/lib.rs
```diff
@@ -121,6 +132,14 @@ impl PowerPolicy {
             }
         }
     }
+
+    /// Top-level event loop function
+    pub async fn process_request(&self) -> Result<(), Error> {
+        match select(self.process_message(), self.attempt_provider_recovery()).await {
```
jerrysxie, line 138: I am concerned about the dropped future consequence here. Even if `attempt_provider_recovery()` has nothing to do, it will tick and be completed. What happens if the tick happened in the middle of `process_message()` and completed first which causes `process_message()` to be dropped? That is a lot of corner cases to think about.

jerrysxie (changes requested):

power-policy-service/src/provider.rs
jerrysxie: Then does other provider need to go into a more limited state as well? Can you add `recovery` to the state diagram in the doc?

docs/power-policy.md
```diff
@@ -27,6 +27,10 @@ stateDiagram-v2
     Provider --> Detached : NotifyDetach
     Idle --> Detached : NotifyDetach
 ```
+
+#### Device recovery
+Requests to a device can fail for any number of reasons (bus communication issues, deadlock in driver code, etc). In most cases this won't result in any issues as most failed requests result in the system being in a safe state. E.g. a failed transition from `Idle` to `ConnectedConsummer` will result in the system just not drawing power. However, a failed request in the `ConnectedProvider` state can leave the system providing more power than intended. The device state isn't enough to capture this situation as the device must be assumed to still be the in the ConnectedConsummer` state. The `Device` struct contains a `recovery` member to track this situation. The specifics of the recovery process are implementation defined.
```
jerrysxie, line 32: > However, a failed request in the `ConnectedProvider` state can leave the system providing more power than intended. The device state isn't enough to capture this situation as the device must be assumed to still be the in the `ConnectedConsummer` state.

A failed state transition in `ConnectedProvider` state means the device would still in `ConnectedProvider` state, right? Can we clarify this? Or did I misunderstand the intention?

jerrysxie (dismissed):

RobertZ2011:
So far all failures seem to be due to `embedded-cfu-protocol` not being public. Will let everything run and double check before merging.

PR #117, opened 2025-02-28
Add peripheral channel write support to eSPI service

### LINKED ISSUE #181: Add peripheral channel write support for eSPI service

jerrysxie:
Implement peripheral channel write support for eSPI service. On host writes to battery, thermal, and time alarm section of the EC memory map, a message will be generated to be forwarded to the underlying service with the updated.

Going through the fields in each section, it does not seem like host would update some of them. But for now, forward any writes to the service for now. Later on with more input from subsystem teams, we can remove the forwarding for fields that should never be updated by the host.

Closes #181

### DIFF
--- a/embedded-service/src/ec_type/mod.rs
+++ b/embedded-service/src/ec_type/mod.rs
@@ -11,6 +11,98 @@ pub enum Error {
     InvalidLocation,
 }
 
+/// Update battery section of memory map based on battery message
+pub fn update_battery_section(msg: &message::BatteryMessage, memory_map: &mut structure::ECMemory) {
+    match msg {
+        message::BatteryMessage::Events(events) => memory_map.batt.events = *events,
+        message::BatteryMessage::LastFullCharge(last_full_charge) => {
+            memory_map.batt.last_full_charge = *last_full_charge
+        }
+        message::BatteryMessage::CycleCount(cycle_count) => memory_map.batt.cycle_count = *cycle_count,
+        message::BatteryMessage::State(state) => memory_map.batt.state = *state,
+        message::BatteryMessage::PresentRate(present_rate) => memory_map.batt.present_rate = *present_rate,
+        message::BatteryMessage::RemainCap(remain_cap) => memory_map.batt.remain_cap = *remain_cap,
+        message::BatteryMessage::PresentVolt(present_volt) => memory_map.batt.present_volt = *present_volt,
+        message::BatteryMessage::PsrState(psr_state) => memory_map.batt.psr_state = *psr_state,
+        message::BatteryMessage::PsrMaxOut(psr_max_out) => memory_map.batt.psr_max_out = *psr_max_out,
+        message::BatteryMessage::PsrMaxIn(psr_max_in) => memory_map.batt.psr_max_in = *psr_max_in,
+        message::BatteryMessage::PeakLevel(peak_level) => memory_map.batt.peak_level = *peak_level,
+        message::BatteryMessage::PeakPower(peak_power) => memory_map.batt.peak_power = *peak_power,
+        message::BatteryMessage::SusLevel(sus_level) => memory_map.batt.sus_level = *sus_level,
+        message::BatteryMessage::SusPower(sus_power) => memory_map.batt.sus_power = *sus_power,
+        message::BatteryMessage::PeakThres(peak_thres) => memory_map.batt.peak_thres = *peak_thres,
+        message::BatteryMessage::SusThres(sus_thres) => memory_map.batt.sus_thres = *sus_thres,
+        message::BatteryMessage::TripThres(trip_thres) => memory_map.batt.trip_thres = *trip_thres,
+        message::BatteryMessage::BmcData(bmc_data) => memory_map.batt.bmc_data = *bmc_data,
+        message::BatteryMessage::BmdData(bmd_data) => memory_map.batt.bmd_data = *bmd_data,
+        message::BatteryMessage::BmdFlags(bmd_flags) => memory_map.batt.bmd_flags = *bmd_flags,
+        message::BatteryMessage::BmdCount(bmd_count) => memory_map.batt.bmd_count = *bmd_count,
+        message::BatteryMessage::ChargeTime(charge_time) => memory_map.batt.charge_time = *charge_time,
+        message::BatteryMessage::RunTime(run_time) => memory_map.batt.run_time = *run_time,
+        message::BatteryMessage::SampleTime(sample_time) => memory_map.batt.sample_time = *sample_time,
+    }
+}
+
+/// Update capabilities section of memory map based on battery message
+pub fn update_capabilities_section(msg: &message::CapabilitiesMessage, memory_map: &mut structure::ECMemory) {
+    match msg {
+        message::CapabilitiesMessage::Events(events) => memory_map.caps.events = *events,
+        message::CapabilitiesMessage::FwVersion(fw_version) => memory_map.caps.fw_version = *fw_version,
+        message::CapabilitiesMessage::SecureState(secure_state) => memory_map.caps.secure_state = *secure_state,
+        message::CapabilitiesMessage::BootStatus(boot_status) => memory_map.caps.boot_status = *boot_status,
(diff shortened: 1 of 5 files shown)


espi-service/src/espi_service.rs
```diff
@@ -0,0 +1,412 @@
...
+            && offset < offset_of!(ec_type::structure::ECMemory, alarm)
+        {
+            //route_to_thermal_service(&mut offset, &mut length).await;
+        } else if offset >= offset_of!(ec_type::structure::ECMemory, alarm) {
+            //route_to_time_alarm_service(&mut offset, &mut length).await;
+        }
+    }
+}
+
+async fn route_to_battery_service(offset: &mut usize, length: &mut usize) {
```
jerrysxie, line 207: @felipebalbi @RobertZ2011 This is one way I can see translating from offset to struct member. But it is extremely verbose.

espi-service/src/espi_service.rs
```diff
@@ -0,0 +1,412 @@
...
+        } else if local_offset == offset_of!(ec_type::structure::Battery, run_time) {
+            let value = memory_map.batt.run_time;
+            *offset += size_of_val(&value);
+            *length -= size_of_val(&value);
+            return Some(ec_type::message::BatteryMessage::RunTime(value));
+        } else if local_offset == offset_of!(ec_type::structure::Battery, sample_time) {
+            let value = memory_map.batt.sample_time;
+            *offset += size_of_val(&value);
+            *length -= size_of_val(&value);
+            return Some(ec_type::message::BatteryMessage::SampleTime(value));
```
jerrysxie, line 330: @philgweber @tullom Does the underlying service need to be notified of each host write?

espi-service/src/espi_service.rs
RobertZ2011: I'm not currently aware of a better way to handle offsets.

RobertZ2011 (approved):

embedded-service/src/ec_type/mod.rs
```diff
@@ -1,4 +1,1123 @@
...
+/// Modifies offset and length
+pub fn mem_map_to_battery_msg(
+    memory_map: &structure::ECMemory,
+    offset: &mut usize,
+    length: &mut usize,
+) -> Result<message::BatteryMessage, Error> {
+    let local_offset = *offset - offset_of!(structure::ECMemory, batt);
+    let mut message: Option<message::BatteryMessage> = None;
+
+    if local_offset == offset_of!(structure::Battery, events) {
```
felipebalbi, line 116: Couldn't this be a `match` statement?

```rust
match local_offset {
    offset_of!(structure::Battery, events) => { ... }
}

File: embedded-service/src/ec_type/mod.rs [hunk 4]
```diff
@@ -1,4 +1,1123 @@
...
+        let value = memory_map.batt.events;
+        *offset += size_of_val(&value);
+        *length -= size_of_val(&value);
+        message = Some(message::BatteryMessage::Events(value));
```
felipebalbi, line 120: given the amount of duplicated code here, perhaps a small macro ought to help?

```rust
macro_rules! into_message! {
    ($member:expr, $msg:expr) => {
        let value = $member;
        *offset += size_of_val(&value);
        *length -= size_of_val(&value);
        message = Some($msg);
    }
}
```

embedded-service/src/ec_type/mod.rs
```diff
@@ -1,4 +1,1123 @@
...
+        *offset += size_of_val(&value);
+        *length -= size_of_val(&value);
+        message = Some(message::BatteryMessage::SampleTime(value));
+    }
+
+    if let Some(msg) = message {
+        Ok(msg)
+    } else {
+        Err(Error::InvalidLocation)
+    }
```
felipebalbi, line 242:
```rust
message.ok_or(Error::InvalidLocation)
````

File: embedded-service/src/ec_type/mod.rs [hunk 6]
```diff
@@ -1,4 +1,1123 @@
...
+
+/// Convert from memory map offset and length to battery message
+/// Modifies offset and length
+pub fn mem_map_to_battery_msg(
+    memory_map: &structure::ECMemory,
+    offset: &mut usize,
+    length: &mut usize,
+) -> Result<message::BatteryMessage, Error> {
+    let local_offset = *offset - offset_of!(structure::ECMemory, batt);
+    let mut message: Option<message::BatteryMessage> = None;
```
felipebalbi, line 114: this can be made immutable if you make the `if` that follows evaluate to the message. It would look something like:

```rust
let message = match local_offset {
    offset_of!(structure::Battery, events) => {
        // ...
    }

    // ...

    _ => None
};
```

But if you do that, then it makes sense to do away with the `Option<T>` and just evaluate to the `Result<T, E>` at the end, thus removing the local variable:

```rust
match local_offset {
    offset_of!(structure::Battery, events) => {
        // ...
        Some(message)
    }

    // ...

    _ => Err(Error::InvalidLocation)
}
```

felipebalbi (changes requested):

.github/workflows/check.yml
```diff
@@ -150,6 +150,31 @@ jobs:
...
+          submodules: true
+          ref: ${{ matrix.commit }}
+      - name: Install nightly
+        uses: dtolnay/rust-toolchain@nightly
+      - name: cargo doc
+        run: cargo doc --target thumbv8m.main-none-eabihf --no-deps --all-features
+        env:
+          RUSTDOCFLAGS: --cfg docsrs
+        working-directory: ${{ matrix.workdir }}
+
```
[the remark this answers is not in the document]
jerrysxie: This has been removed in the latest round of changes. I figured out why it was unhappy. I needed to add a `.cargo/config.toml` to force it to compile for a particular target.

embedded-service/src/ec_type/mod.rs
jerrysxie: Great idea, that is what macro are for. I was using copilot to generate most of the boiler template code.

embedded-service/src/ec_type/mod.rs
felipebalbi: fjust tested, you're right. We would need to make it `offset if offset == offset_of!(...)`. Oh well 👎

espi-service/src/espi_service.rs
```diff
@@ -19,112 +20,94 @@ impl Service<'_> {
...
+        ec_type::update_time_alarm_section(msg, &mut memory_map);
+    }
+
+    async fn route_to_service(&self, offset: usize, length: usize) -> Result<(), ec_type::Error> {
+        let mut offset = offset;
+        let mut length = length;
+
+        while length > 0 {
+            if offset >= offset_of!(ec_type::structure::ECMemory, batt)
+                && offset < offset_of!(ec_type::structure::ECMemory, therm)
```
philgweber, line 47: Can we use size of the batt items for end offset rather than using start address of therm in case the ordering is different?

philgweber (approved):

embedded-service/src/ec_type/mod.rs
jerrysxie: Due to Rust macro hygiene it, local identifiers cannot be used outside of the macro and outside identifiers cannot be directly used inside the macro.

So it will have to be something like this

```
/// Helper macro to simplify the conversion of memory map to message
macro_rules! into_message {
    ($offset:ident, $length:ident, $message:ident, $member:expr, $msg:expr) => {
        let value = $member;
        *$offset += size_of_val(&value);
        *$length -= size_of_val(&value);
        $message = Some($msg(value));
    };
}
```

```
        into_message!(
            offset,
            length,
            message,
            memory_map.alarm.events,
            message::TimeAlarmMessage::Events
        );
```

What do you think?

felipebalbi: I think it's okay

espi-service/src/espi_service.rs
jerrysxie: Good point, since we might be still moving things around. Add more checking for writes to invalid locations and use size to scope the search.

rafiaaa456 (approved):

PR #119, opened 2025-03-03
Espi + Battery example rt633

tullom:


battery-service/src/charger/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    pub async fn rx_msg_from_service(&self) {
+        let rx_message = self.rx.receive().await;
+        match rx_message {
+            BatteryMsgs::Acpi(msg) => match msg {
+                crate::BatteryMessage::CycleCount(cycles) => {
+                    self.tx
+                        .send(Ok(BatteryMsgs::Acpi(crate::BatteryMessage::CycleCount(cycles + 1))))
```
jerrysxie, line 35: Should CycleCount be forwarded to the charger?

battery-service/src/charger/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+impl<SmartCharger: embedded_batteries_async::charger::Charger> Charger<SmartCharger> {
+    pub fn new(smart_charger: SmartCharger) -> Self {
+        Charger {
+            device: RefCell::new(smart_charger),
+            rx: Channel::new(),
+            tx: Channel::new(),
+        }
+    }
+
+    pub async fn rx_msg_from_service(&self) {
```
jerrysxie, line 29: Calling this `rx_msg_from_service()` is a little misleading, it kind of implies there should be a `tx` equivalent. But this is processing of service message, is that correct?

battery-service/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+                    .rx
+                    .try_send(BatteryMsgs::Oem(msg))
+                    .map_err(|_| BatteryServiceErrors::BufferFull),
+                _ => todo!(),
+            },
+        }
+    }
+
+    /// Function that handles responses from the charger and fuel gauge and sends messages over the endpoint.
+    pub async fn handle_charger_fuel_gauge_msg(&self) -> Result<(), BatteryServiceErrors> {
```
jerrysxie, line 89: Will the service need to do any housekeeping here like passing message between charger and fuel gauge?

Can we compose this in a more structured way? Or document this in a README?

So there is the charger service and fuel gauge service that each take message from the battery service which forwards message from transport to them and coordinating between them. Charger and fulel gauge services are responsible for talking with the underlying hardware. Is that the relationship?

jerrysxie (changes requested):

RobertZ2011 (approved):

battery-service/src/lib.rs
tullom: This function receives messages from the underlying charger or fuel gauge services. The battery-service, which sits a layer up from the charger/fuel gauge, is where the FSMs that OEMs can build will live. OEMs can customize the messages received in `handle_charger_fuel_gauge_msg()` to route a response to another service, or back into the charger if the message originated from the fuel gauge. This is also where state will be handled.

Edit: Now that we are officially pulling charging functionality out of battery-service, I envision this function as a "first pass" routing, which routes the standard messages and returns the msg to the app level so the implementer can build their state machines on top of it.

battery-service/src/charger/mod.rs
tullom: Good catch, removed.

battery-service/src/charger/mod.rs
tullom: Correct. Changed to `process_service_message()`

jerrysxie (approved):
I think the first commit is the same as [Battery service skeleton by tullom · Pull Request #118 · OpenDevicePartnership/embedded-services](https://github.com/OpenDevicePartnership/embedded-services/pull/118)? This example looks good.

battery-service/src/fuel_gauge/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+                        })
+                        .map_err(|_| FuelGaugeError::Bus);
+                    if res.is_ok() {
+                        info!(
+                            "Remaining cap = {}",
+                            if let BatteryMsgs::Acpi(crate::BatteryMessage::RemainCap(cap)) = res.unwrap() {
+                                cap
+                            } else {
+                                1234
+                            }
```
felipebalbi, line 111: if `res` is not `Ok()`, this will `panic!()` before that `1234`. How about:

```rust
res.unwrap_or(1234)
```

ditto to all other instances of this

tullom: technically the 1234 branch will never get hit actually, since this logic is wrapped in an `if res.is_ok()`. I think i tried to get it to print out 1234 if it error'd (for some espi testing) but after taking a look it should actually be `unreachable!()`. I have updated the PR.

tullom: note that I also need to have the else branch because when i extract the inner value from the enum, i technically need to handle the cases for all the BatteryMsgs. But since I precisely know the type of the message since i construct it in the line above, i know the code will never take the else branch and thus will never panic.

felipebalbi (approved):

PR #121, opened 2025-03-03
Type-C service source implementation

RobertZ2011:


type-c-service/src/driver/tps6699x.rs
```diff
@@ -100,50 +73,217 @@ impl<'a, const N: usize, M: RawMutex, B: I2c> Controller for Tps6699x<'a, N, M,
...
+
+        self.port_status[port.0 as usize].set(port_status);
+        Ok(events)
+    }
+
+    /// Wait for an event on any port
+    async fn wait_interrupt_event(&self, tps6699x: &mut tps6699x::Tps6699x<'a, M, B>) -> Result<(), Error<B::Error>> {
+        let interrupts = tps6699x.wait_interrupt(false, |_, _| true).await;
+
+        for (interrupt, cell) in zip(interrupts.iter(), self.port_events.iter()) {
```
asasine, line 142: Are `interrupts` and `self.port_events` guaranteed to be ordered the same?

type-c-service/src/wrapper/mod.rs
```diff
@@ -6,14 +6,20 @@ use core::cell::{Cell, RefCell};
...
+use embedded_usb_pd::{type_c::Current as TypecCurrent, Error, PdError, PortId as LocalPortId};
 
 mod pd;
 mod power;
 
+/// Default current to source
+const DEFAULT_SOURCE_CURRENT: TypecCurrent = TypecCurrent::Current1A5;
+/// Threshold power capability before we'll attempt to sink from a dual-role supply
+/// This ensures we don't try to sink from something like a phone
+const DUAL_ROLE_CONSUMER_THRESHOLD: u32 = 15000;
```
asasine, line 21: "power" has a unit but `u32` us unitless. What unit is `15000` and if a newtype doesn't exist for it, should one? If not, can the unit be specified in the doc comment?

embedded-service/src/type_c/controller.rs
```diff
@@ -244,6 +269,25 @@ pub trait Controller {
         port: LocalPortId,
         enable: bool,
     ) -> impl Future<Output = Result<(), Error<Self::BusError>>>;
+    /// Enable or disable sourcing
+    fn enable_source(
```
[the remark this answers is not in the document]
RobertZ2011: Done. Changing the `tps6699x` crate will be done as a follow-up

embedded-service/src/type_c/event.rs
```diff
@@ -26,6 +28,11 @@ impl PortEventKind {
         Self(PortEventKindRaw(0))
     }
 
+    /// Returns the union of self and other
+    pub fn union(self, other: PortEventKind) -> PortEventKind {
+        PortEventKind(PortEventKindRaw(self.0 .0 | other.0 .0))
```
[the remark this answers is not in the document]
RobertZ2011: This seems to be what `rustfmt` wants. Added a comment so we don't have to fight it on each commit.

type-c-service/src/wrapper/mod.rs
RobertZ2011: Renamed to `DUAL_ROLE_CONSUMER_THRESHOLD_MW`. Anything further should be a larger discussion on if we're going to use a crate like `uom`.

type-c-service/src/driver/tps6699x.rs
RobertZ2011: They are. Everything related to a local port is stored and ordered by the local port ID. I'm not sure if there's a way to enforce some kind of compile time checking of this but I'm open to suggestions.

asasine: I think only redesigning to arrays of structs rather than structs of arrays.

asasine (approved):

embedded-service/src/power/policy/action/device.rs
```diff
@@ -85,6 +85,12 @@ impl<'a, S: Kind> Device<'a, S> {
 
     /// Request the given power from the power policy service
     async fn request_provider_power_capability_internal(&self, capability: PowerCapability) -> Result<(), Error> {
+        if self.device.provider_capability().await == Some(capability) {
```
jerrysxie, line 88: So if I am understanding this correctly, if it is already set, then it has been connected already?

examples/rt685s-evk/src/bin/type_c.rs
```diff
@@ -145,11 +145,15 @@ async fn main(spawner: Spawner) {
...
     let controller = CONTROLLER.init(Controller::new_tps66994(device, ADDR0).unwrap());
-    let (tps6699x, interrupt) = controller.make_parts();
+    let (mut tps6699x, interrupt) = controller.make_parts();
 
     info!("Spawining interrupt task");
     spawner.must_spawn(interrupt_task(int_in, interrupt));
 
+    info!("Resetting PD controller");
+    let mut delay = Delay;
+    tps6699x.reset(&mut delay).await.unwrap();
```
jerrysxie, line 157: The commit message does not seem to match what this commit is doing.

jerrysxie (dismissed):

embedded-service/src/power/policy/action/device.rs
RobertZ2011: It's more that if the power policy receives a request that matches what it already has then there's nothing to do since the power policy state is already up to date. I added a comment and changed the commit message to make it clearer.

jerrysxie (approved):

felipebalbi (approved):

PR #122, opened 2025-03-05
Create main type-C service task

RobertZ2011:
This PR creates the framework for the main type-C task. This task will respond to events on each port. Currently there is no logic for the type-C task to implement, but this will be fleshed out by subsequent PRs.


RobertZ2011:
Check failures are due to known embedded-cfu-protocol not being public issue.

embedded-service/src/type_c/controller.rs
```diff
@@ -184,8 +187,21 @@ impl<'a> Device<'a> {
     /// Notify that there are pending events on one or more ports
     /// Each bit corresponds to a global port ID
     pub async fn notify_ports(&self, events: PortEventFlags) {
```
RobertZ2011, line 189: @asasine What do you think of the event flags being used here? This one is just a bitfield to indicate a pending event on the corresponding port. Then each port uses bit flags to report the various events it's signaling (status changed, new provider contract, new consumer contract), but each of those has a dedicated check function like `fn status_changed() -> bool`.

felipebalbi (approved):

embedded-service/src/type_c/controller.rs
asasine: Using a newtype here is definitely an improvement over passing a `u32` around. I think a safer version would be a struct over `[bool; N]` instead of bitpacking into an integer. You already know how many flags there should be based `self.ports.len()` or `self.num_ports`, you know which port based on the index, and you're less likely to mistakenly interpret this as an actual numeric value.

For the actual events, can you point me to where this is used?

asasine: Ah, found it. Since the meaning of the bits are known, I also think writing `PortEventKind` as a more explicit `struct PortEventKind { pub plug_inserted_or_removed: bool, pub new_power_contract_as_consumer: bool, }` has similar benefits.

This all being said, writing things in this explicit way is also more verbose and it can be time-consuming to write the tests to verify that it deserializes correctly from the bits coming from remote sources (e.g., i2c registers). Bitfields are a common enough pattern in embedded that I'd definitely look the other way and accept the current implementation.

type-c-service/src/wrapper/power.rs
```diff
@@ -0,0 +1,114 @@
...
+        port: LocalPortId,
+    ) -> Result<&'b policy::device::Device, Error<C::BusError>> {
+        if port.0 > N as u8 {
+            return PdError::InvalidPort.into();
+        }
+        Ok(&self.power[port.0 as usize])
+    }
+
+    /// Handle a new consumer contract
+    /// None of the event processing functions return errors to allow processing to continue for other ports on a controller
```
asasine, line 19: Should the error still be returned but the caller decide to ignore and continue processing other ports?

asasine (approved):
Looks functional enough to me. All of my comments are subjective.

jerrysxie (dismissed):

embedded-service/src/type_c/event.rs
```diff
@@ -1,58 +1,95 @@
...
+    pub u8, plug_inserted_or_removed, set_plug_inserted_or_removed: 0, 0;
+    /// New power contract as consumer
+    pub u8, new_power_contract_as_consumer, set_new_power_contract_as_consumer: 3, 3;
+}
+
+/// Type-safe wrapper for the raw port event kind
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub struct PortEventKind(pub u32);
+pub struct PortEventKind(pub PortEventKindRaw);
```
asasine, line 22: Does the field need to be public? I think the only public use is for converting to a `u32` for the sake of formatting in a log, and `PortEventKind` already implements `Debug` and sometimes `defmt::Format`

asasine (approved):
Looks good to me. The `PortEventKind` abstraction can be tightened up some but not a big deal.

embedded-service/src/type_c/controller.rs
RobertZ2011: Made this more safe. `PortEventFlags` is now implemented on top of `bitvec` so we don't have to roll our own. `PortEventKind` is now implemented on top of `bitfield`. I moved away from `bitflags` because it felt like a C-ism.

embedded-service/src/type_c/event.rs
```diff
@@ -1,58 +1,95 @@
...
-use bitfield::{Bit, BitMut};
-use bitflags::bitflags;
+use super::GlobalPortId;
 
-/// Port event kind
+bitfield! {
+    /// Raw bitfield of possible port events
+    #[derive(Copy, Clone, PartialEq, Eq)]
+    #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+    pub struct PortEventKindRaw(u32);
```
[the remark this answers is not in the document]
RobertZ2011: Done.

embedded-service/src/type_c/event.rs
RobertZ2011: Done.

type-c-service/src/wrapper/power.rs
RobertZ2011: Done, threaded error types through for both currently implemented event processing functions.

felipebalbi (approved):

asasine (approved):

type-c-service/src/wrapper/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+        Self {
+            pd_controller,
+            power,
+            controller: RefCell::new(controller),
+            active_events: [const { Cell::new(PortEventKind::none()) }; N],
+        }
+    }
+
+    /// Handle a plug event
+    /// None of the event processing functions return errors to allow processing to continue for other ports on a controller
```
asasine, line 40: FYI this is an outdated comment now

RobertZ2011: I'll update this in a follow-up PR.

jerrysxie (approved):

Issue #174, opened 2025-03-11
Storage-Bus: Refine the trait methods to include some generic APIs like Read, Write, Lock, etc


Shaibal-Microsoft:

Shaibal-Microsoft:
@JamesHuard  @jerrysxie  FYI

PR #176, opened 2025-03-11
Adding storage bus traits

Shaibal-Microsoft:
This PR adds generic storage bus traits to be used by the NOR flash device drivers. Device drivers can use the traits to access the lower level storage busses like FlexSPI, SPI, etc.
PR also adds traits for NAND storage bus for future support

### DIFF
--- a/storage-bus/Cargo.toml
+++ b/storage-bus/Cargo.toml
@@ -1,0 +1,33 @@
+[package]
+name = "storage_bus"
+version = "0.1.1"
+edition = "2021"
+description = "Storage Bus interfaces"
+repository = "https://github.com/OpenDevicePartnership/embedded-services"
+rust-version = "1.83"
+license = "MIT"
+
+[dependencies]
+embedded-services = { path = "../embedded-service" }
+embassy-time = { git = "https://github.com/embassy-rs/embassy" }
+embassy-sync = { git = "https://github.com/embassy-rs/embassy" }
+embassy-executor = { git = "https://github.com/embassy-rs/embassy" }
+defmt = { version = "0.3", optional = true }
+log = { version = "0.4.14", optional = true }
+
+[features]
+default = []
+defmt = [
+    "dep:defmt",
+    "embedded-services/defmt",
+    "embassy-time/defmt",
+    "embassy-sync/defmt",
+    "embassy-executor/defmt",
+]
+log = [
+    "dep:log",
+    "embedded-services/log",
+    "embassy-time/log",
+    "embassy-sync/log",
+    "embassy-executor/log",
+]
--- a/storage-bus/src/lib.rs
+++ b/storage-bus/src/lib.rs
@@ -1,0 +1,4 @@
+#![no_std]
+
+pub mod nand;
+pub mod nor;
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -59,7 +59,7 @@ jobs:
       fail-fast: false
       matrix:
         commit: ${{ fromJSON(needs.commit_list.outputs.commits) }}
-        workdir: [ "embedded-service", "espi-service", "hid-service", "power-button-service"]
+        workdir: [ "embedded-service", "espi-service", "hid-service", "power-button-service", "storage-bus"]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -86,7 +86,7 @@ jobs:
         # Get early warning of new lints which are regularly introduced in beta channels.
         toolchain: [stable, beta]
         commit: ${{ fromJSON(needs.commit_list.outputs.commits) }}
-        workdir: [ "embedded-service", "espi-service", "hid-service", "power-button-service"]
+        workdir: [ "embedded-service", "espi-service", "hid-service", "power-button-service", "storage-bus"]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -136,7 +136,7 @@ jobs:
       fail-fast: false
       matrix:
         commit: ${{ fromJSON(needs.commit_list.outputs.commits) }}
-        workdir: [ "embedded-service", "espi-service", "hid-service", "power-button-service"]
+        workdir: [ "embedded-service", "espi-service", "hid-service", "power-button-service", "storage-bus"]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -162,7 +162,7 @@ jobs:
       fail-fast: false
       matrix:
         commit: ${{ fromJSON(needs.commit_list.outputs.commits) }}
-        workdir: [ "embedded-service", "espi-service", "hid-service", "power-button-service"]
+        workdir: [ "embedded-service", "espi-service", "hid-service", "power-button-service", "storage-bus"]
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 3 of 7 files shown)


Shaibal-Microsoft:
@jerrysxie @JamesHuard  Created issue https://github.com/OpenDevicePartnership/embedded-services/issues/174 to track the discussion

storage-bus/src/storage_bus.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Nor Storage Command upper byte                       
+    pub cmd_ub: Option<u8>,
+    /// Address of the command
+    pub addr: Option<u32>,
+    /// Address width in bytes              
+    pub addr_width: Option<u8>,
+    /// DDR or SDR mode             
+    pub mode: NorStorageCmdMode,
+    /// Number of Dummy clock cycles. Assuming max 256 dummy cycles beyond which its impractical           
+    pub dummy: Option<u8>,
```
JamesHuard, line 45: consider using a unit type for making it clear whether the delay is in clocks or bytes
```[rust]
// enum approach
pub enum DummyCycles {
    Clocks(u8),
    Bytes(u8),
}

// struct approach
pub struct DummyClocks(u8);
impl SealedUnit for DummyClocks {}
pub struct DummyBytes(u8);
impl SealedUnit for DummyBytes {}
// helper functions, into etc
```

Shaibal-Microsoft: took the enum approach

storage-bus/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "storage_bus"
+version = "0.1.1"
+edition = "2021"
+description = "Storage Bus interfaces"
+repository = "https://github.com/OpenDevicePartnership/embedded-services"
+rust-version = "1.83"
+license = "MIT"
```
jerrysxie, line 8: Add these features and dependencies to avoid workflow failures for now:

```
[dependencies]
embedded-services = { path = "../embedded-service" }
embassy-time = { git = "https://github.com/embassy-rs/embassy" }
embassy-sync = { git = "https://github.com/embassy-rs/embassy" }
embassy-executor = { git = "https://github.com/embassy-rs/embassy" }
defmt = { version = "0.3", optional = true }
log = { version = "0.4.14", optional = true }

[features]
default = []
defmt = [
    "dep:defmt",
    "embedded-services/defmt",
    "embassy-time/defmt",
    "embassy-sync/defmt",
    "embassy-executor/defmt",
]
log = [
    "dep:log",
    "embedded-services/log",
    "embassy-time/log",
    "embassy-sync/log",
    "embassy-executor/log",
]
```

storage-bus/src/storage_bus.rs
```diff
@@ -0,0 +1,129 @@
...
+    StorageBusNotAvailable,
+    /// Bus IO error while sending command
+    /// Could be used for example
+    /// 1 - Bus read error
+    /// 2 - Bus write error
+    StorageBusIoError,
+    /// Bus transfer timed out. Timeout could happen due to
+    /// 1. Bus is not responding
+    /// 2. External device did not respond in time
+    StorageBusTimeout,
```
jerrysxie, line 81: Should the underlying flash part driver handle timeout or should the application layer? If we expected the lower layer to handle timeout, then shouldn't a timeout be passed in as well.

For async, we can use `with_timout()`, https://docs.rs/embassy-time/latest/embassy_time/fn.with_timeout.html to drop the async future when the timeout expires.

For blocking, I would assume we would just block unless the client passes in a timeout.

jerrysxie (changes requested):

storage-bus/src/storage_bus.rs
Shaibal-Microsoft: Agree. Timeout should be detected by flash part driver. Otherwise we are inviting complexity at lower level.
I will remove Timeout error.

storage-bus/src/storage_bus.rs
```diff
@@ -0,0 +1,129 @@
...
+/// Async NOR Storage Driver. The trait introduces a method to send command to the bus
+/// The NOR device driver should implement this trait to send command to the bus
+/// Parameters:
+///   cmd - Command to be sent to the bus
+///   read_buf - Read buffer to store the data read from the bus
+///   write_buf - Write buffer to write the data to the bus
+/// Returns:
+///   Result<(), StorageBusError> - Result of the command sent to the bus
+///   StorageBusError - Error code if the command failed
+pub trait AsyncNorStorageBusDriver {
```
Shaibal-Microsoft, line 115: Do I need to add Send super trait to make it work for multi-threaded executor? @JamesHuard  @jerrysxie

JamesHuard: https://doc.rust-lang.org/nomicon/send-and-sync.html

In short, I don't think you need to manually enforce Send here. 

Ultimately the send requirement would be imposed by the ownership "stack" that the struct impl'ing AsyncNorStorageBusDriver is part of. For a multi-executor model, it's unlikely we'd want to share the NOR storage directly across threads anyway (neither Send nor Sync) because that would be a priority inversion between the higher pri executor and the lower one if both need to perform writes to said device at the same time. If we had a pre-emptive scheduler in any form, we'd likely isolate the ownership of this resource to a single (likely lower priority) task that requests are queued to. 

Since you have `send_command(&mut self, ...` here, the mut keyword will already gate multiple contexts calling at the same time -- since you cannot have multiple mutable references to the same object at the same time.

storage-bus/src/storage_bus.rs
```diff
@@ -0,0 +1,129 @@
...
+/// Blocking NOR Storage Driver. The trait introduces a method to send command to the bus
+/// The NOR device driver should implement this trait to send command to the bus
+/// Parameters:
+///    cmd - Command to be sent to the bus
+///    read_buf - Read buffer to store the data read from the bus
+///    write_buf - Write buffer to write the data to the bus
+/// Returns:
+///    Result<(), StorageBusError> - Result of the command sent to the bus
+///    StorageBusError - Error code if the command failed
+pub trait BlockingNorStorageBusDriver {
```
JamesHuard, line 95: do we need to consider any sort of power management abstractions, or is that something that's solely tied to the chip implementation? Thinking about an example such as if SPI low power looks different to FlexSPI or some other chip-specific implementation

storage-bus/src/storage_bus.rs
Shaibal-Microsoft: Thanks Jimi.

storage-bus/src/storage_bus.rs
Shaibal-Microsoft: Good point. I have added this point in the following issue to track after this PR

https://github.com/OpenDevicePartnership/embedded-services/issues/174

storage-bus/src/storage_bus.rs
Shaibal-Microsoft: fixed

storage-bus/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+#![no_std]
+pub mod storage_bus;
```
JamesHuard, line 2: nit: why do we need a separate mod for this? Without a `pub use::storage_bus::*;` this potentially makes every reference to these traits "double scoped": `storage_bus::storage_bus::NorStorageCmdMode` for example.

Should the code in storage_bus.rs go here directly? Or is there some expansion we might expect and we should rename storage_bus.rs to something else (like nor.rs? and expect a nand.rs eventually? `storage_bus::nor::BlockingStorageDriver` for example

jerrysxie (approved):

storage-bus/src/lib.rs
Shaibal-Microsoft: Actually I like idea of having separate module for nor storage and nand storage bus traits for future extensibility.
I have made the changes and it will look like below.

use storage_bus::nor::BlockingNorStorageBusDriver

Shaibal-Microsoft: fixed

PR #184, opened 2025-03-14
Update eSPI memory map layout to the latest and version it as 0.1.0

jerrysxie:
- Add EC memory layout YAML file to source control
- Add Python generator to generate both C and Rust header files for the EC memory layout
- Update the EC memory layout to the latest generated from YAML file
- Version the EC memory layout as 0.1.0
- Update eSPI service to set the version on initialization

### DIFF
--- a/embedded-service/src/ec_type/generator/ec-memory-generator.py
+++ b/embedded-service/src/ec_type/generator/ec-memory-generator.py
@@ -1,0 +1,98 @@
+import sys,yaml
+
+# Function to convert YAML data to Rust structures
+def yaml_to_rust(data):
+    rust_code = "//! EC Internal Data Structures\n\n"
+    rust_code += "pub const EC_MEMMAP_VERSION: Version = Version {major: 0, minor: 1, spin: 0, res0: 0};\n\n"
+    for key, value in data.items():
+        rust_code += "#[allow(missing_docs)]\n"
+        rust_code += "#[repr(C, packed)]\n"
+        rust_code += "#[derive(Clone, Copy, Debug, Default)]\n"
+        rust_code += f"pub struct {key} {{\n"
+        for sub_key, sub_value in value.items():
+            if isinstance(sub_value, dict) and 'type' in sub_value:
+                rust_code += f"    pub {sub_key}: {sub_value['type']},\n"
+            else:
+                rust_code += f"    pub {sub_key}: {sub_value},\n"
+        rust_code += "}\n\n"
+
+    return rust_code
+
+# Function to convert YAML data to C structures
+def yaml_to_c(data):
+    c_code = "#pragma once\n\n"
+    c_code += "#include <stdint.h>\n\n"
+    c_code += "#pragma pack(push, 1)\n\n"
+    for key, value in data.items():
+        c_code += "typedef struct {\n"
+        for sub_key, sub_value in value.items():
+            if isinstance(sub_value, dict) and 'type' in sub_value:
+                c_code += f"    {type_to_c_type(sub_value['type'])} {sub_key};\n"
+            else:
+                c_code += f"    {sub_value} {sub_key};\n"
+        c_code += f"}} {key};\n\n"
+
+    c_code += "#pragma pack(pop)\n\n"
+
+    c_code += "const Version EC_MEMMAP_VERSION = {0x00, 0x01, 0x00, 0x00};\n"
+    return c_code
+
+def type_to_c_type(type_str):
+    if type_str == 'u32':
+        return 'uint32_t'
+    elif type_str == 'u16':
+        return 'uint16_t'
+    elif type_str == 'u8':
+        return 'uint8_t'
+    elif type_str == 'i32':
+        return 'int32_t'
+    elif type_str == 'i16':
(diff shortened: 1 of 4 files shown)


embedded-service/src/ec_type/structure.rs
```diff
@@ -23,6 +30,7 @@ pub struct Capabilities {
     pub temp_mask: u16,
     pub key_mask: u16,
     pub debug_mask: u16,
+    pub res0: u16,
```
jerrysxie, line 33: @philgweber this was in the C version.

embedded-service/src/ec_type/structure.rs
```diff
@@ -53,6 +61,7 @@ pub struct TimeAlarm {
 #[derive(Clone, Copy, Debug, Default)]
 pub struct Battery {
     pub events: u32,
+    pub status: u32,
```
jerrysxie, line 64: @philgweber this was in the C version.

embedded-service/src/ec_type/structure.rs
```diff
@@ -100,12 +109,21 @@ pub struct Thermal {
...
+    pub event: u16,
+}
+
 #[allow(missing_docs)]
 #[repr(C, packed)]
 #[derive(Clone, Copy, Debug, Default)]
 pub struct ECMemory {
     pub ver: Version,
     pub caps: Capabilities,
+    pub notif: Notifications,
```
jerrysxie, line 126: @philgweber added for notification.

tullom (changes requested):

embedded-service/src/ec_type/generator/ec-memory-generator.py
```diff
@@ -0,0 +1,102 @@
...
+    rust_code = "//! EC Internal Data Structures\n\n"
+    rust_code += "pub const EC_MEMMAP_VERSION: Version = Version {major: 0, minor: 1, spin: 0, res0: 0};\n\n"
+    for key, value in data.items():
+        rust_code += "#[allow(missing_docs)]\n"
+        rust_code += "#[repr(C, packed)]\n"
+        rust_code += "#[derive(Clone, Copy, Debug, Default)]\n"
+        rust_code += f"pub struct {key} {{\n"
+        for sub_key, sub_value in value.items():
+            if isinstance(sub_value, dict) and 'type' in sub_value:
+                rust_code += f"    pub {sub_key}: {sub_value['type']},\n"
```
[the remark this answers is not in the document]
jerrysxie: In the war between tabs and spaces, I am firmly on the side of spaces, LOL. I prefer space as tab will have different width depending on the editor.

embedded-service/src/ec_type/generator/ec-memory-generator.py
```diff
@@ -0,0 +1,102 @@
...
+    c_code = yaml_to_c(data)
+
+    rust_output_filename = "structure.rs"
+    c_output_filename = "ecmemory.h"
+
+    try: 
+      with open(rust_output_filename, "w") as output_file:
+        output_file.write(rust_code)
+      print(f"Rust code has been written to {rust_output_filename}")
+    except Exception as e:
```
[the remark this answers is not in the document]
jerrysxie: Good point, I have not written Python for a while, bad form.

tullom (approved):
approved pending Phil confirms struct fields

embedded-service/src/ec_type/structure.rs
philgweber: The status was requested by surface so they could return this value from _STA in the OS. Lets leave this in for now and if we get pushed on space we can try to remove or compress this.

embedded-service/src/ec_type/structure.rs
philgweber: Yes from the service side we get aborts if the accesses are not naturally aligned. This makes rounds out the size of the structure to be 64-bit aligned.

embedded-service/src/ec_type/structure.rs
philgweber: We may need to actually add res for u32 here to keep the next structure 64-bit aligned so service in Hafnium doesn't barf. Otherwise we need to access on byte at a time in Hafnium.

philgweber (changes requested):
can we just add a reserve field in Notification to keep the start of the  next structure 64-bit aligned?

embedded-service/src/ec_type/structure.rs
jerrysxie: I see, so each struct needs to be 64-bit aligned. I will add logic to the script to check for that.

philgweber: I see all our values are only 32-bit access so keeping 32-bit aligned is fine. If we bring in 64-bit values we may need to change access on the apps side to prevent alignment faults.

philgweber (approved):

embedded-service/src/ec_type/structure.rs
jerrysxie: Gotcha, I added checking for 32-bit alignment of the struct to the script.

felipebalbi (approved):

PR #185, opened 2025-03-18
espi-service: update due to imxrt breakage

jerrysxie:
- Fix breakage due to embassy-imxrt
- Lock embassy-imxrt to a commit to avoid accidental breakage in the future
   - https://github.com/OpenDevicePartnership/embassy-imxrt/commit/35d430666574379ef2186c25172ca848ae19d11b

### DIFF
--- a/espi-service/src/espi_service.rs
+++ b/espi-service/src/espi_service.rs
@@ -213,7 +213,7 @@ pub async fn espi_service(mut espi: espi::Espi<'static>, memory_map_buffer: &'st
             Ok(espi::Event::Port80) => {
                 info!("eSPI Port 80");
             }
-            Ok(espi::Event::WireChange) => {
+            Ok(espi::Event::WireChange(_)) => {
                 info!("eSPI WireChange");
             }
             Err(_) => {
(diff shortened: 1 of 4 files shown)


espi-service/src/espi_service.rs
```diff
@@ -213,7 +213,7 @@ pub async fn espi_service(mut espi: espi::Espi<'static>, memory_map_buffer: &'st
             Ok(espi::Event::Port80) => {
                 info!("eSPI Port 80");
             }
-            Ok(espi::Event::WireChange) => {
+            Ok(espi::Event::WireChange(_)) => {
```
jerrysxie, line 216: @felipebalbi We should consider versioning embassy-imxrt so we can lock to version and signal breaking changes.

sukomath (approved):

espi-service/src/espi_service.rs
felipebalbi: We can use the git SHA. Problem with versioning it now is that the crate is not in crates.io.

felipebalbi (approved):

espi-service/src/espi_service.rs
jerrysxie: Done, lock to a git SHA and only update with intention.

felipebalbi (approved):

tullom (approved):

PR #186, opened 2025-03-20
Add Power Policy message handling to Battery service

tullom:
Tested in hardware, works.

### DIFF
--- a/battery-service/src/lib.rs
+++ b/battery-service/src/lib.rs
@@ -2,6 +2,9 @@
 
 use embassy_futures::select::select;
 use embassy_futures::select::Either::{First, Second};
+use embassy_sync::blocking_mutex::raw::NoopRawMutex;
+use embassy_sync::signal::Signal;
+use embassy_time::Timer;
 use embedded_batteries_async::charger::{MilliAmps, MilliVolts};
 use embedded_services::comms::{self, External};
 use embedded_services::ec_type::message::BatteryMessage;
@@ -42,6 +45,7 @@ pub struct Service<
     pub endpoint: comms::Endpoint,
     pub charger: charger::Charger<SmartCharger>,
     pub fuel_gauge: fuel_gauge::FuelGauge<SmartBattery>,
+    policy_signal: Signal<NoopRawMutex, MilliAmps>,
 }
 
 impl<
@@ -54,6 +58,7 @@ impl<
             endpoint: comms::Endpoint::uninit(comms::EndpointID::Internal(comms::Internal::Battery)),
             charger: charger::Charger::new(smart_charger),
             fuel_gauge: fuel_gauge::FuelGauge::new(fuel_gauge),
+            policy_signal: Signal::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


felipebalbi (dismissed):

battery-service/src/lib.rs
```diff
@@ -122,10 +128,24 @@ impl<
...
+        let current = self.policy_signal.wait().await;
+        Timer::after_millis(1000).await;
+        let res = self
+            .charger
+            .rx
+            .try_send(BatteryMsgs::Oem(OemMessage::ChargeCurrent(current)));
+        if res.is_err() {
+            error!("Charger buffer full");
+        }
+        info!("SENT NEW CHARGE CURRENT");
```
asasine, line 147: If the rx buffer is full, this will log both the error and info. Consider pattern matching to only log one or the other
```suggestion
        let res = self
            .charger
            .rx
            .try_send(BatteryMsgs::Oem(OemMessage::ChargeCurrent(current)));

        match res {
            Ok(_) => info!("SENT NEW CHARGE CURRENT"),
            Err(_) => error!("Charge buffer full"),
        }
```

battery-service/src/lib.rs
```diff
@@ -41,6 +45,7 @@ pub struct Service<
     pub endpoint: comms::Endpoint,
     pub charger: charger::Charger<SmartCharger>,
     pub fuel_gauge: fuel_gauge::FuelGauge<SmartBattery>,
+    policy_signal: Signal<NoopRawMutex, MilliAmps>,
```
asasine, line 48: Would it make sense to signal with the entire `PowerCapability` since the charger can set both current and voltage?

battery-service/src/lib.rs
```diff
@@ -134,18 +154,33 @@ impl<
...
+                embedded_services::power::policy::CommsData::ConsumerDisconnected(device_id) => {
+                    info!("Consumer disconnected: {}", device_id.0);
+                    let res = self.charger.rx.try_send(BatteryMsgs::Oem(OemMessage::ChargeCurrent(0)));
+                    if res.is_err() {
+                        error!("Charger buffer full");
+                    }
+                }
+                embedded_services::power::policy::CommsData::ConsumerConnected(device_id, power_capability) => {
+                    info!("Consumer connected: {} {:?}", device_id.0, power_capability);
+                    info!("BATTERY: WAITING HALF SECOND");
```
asasine, line 180: Is this a leftover debugging comment or is there a missing sleep?

battery-service/src/lib.rs
```diff
@@ -122,10 +128,24 @@ impl<
                     .await
                     .unwrap();
             }
+            BatteryMsgs::Oem(_) => (),
             _ => todo!(),
```
asasine, line 132: With the additional `BatteryMsgs::Oem(_)` arm above, this catch-all is now unreachable.

asasine (commented):
I'm getting up to speed on the project and noticed some things of minor importance.

battery-service/src/lib.rs
```diff
@@ -134,18 +154,33 @@ impl<
...
-
-        if let Some(msg) = message.data.get::<OemMessage>() {
+        } else if let Some(msg) = message.data.get::<OemMessage>() {
             // Todo: Handle case where buffer is full.
             if self.handle_transport_msg(BatteryMsgs::Oem(*msg)).is_err() {
                 error!("Buffer full");
             }
+        } else if let Some(msg) = message.data.get::<CommsMessage>() {
+            match msg.data {
+                embedded_services::power::policy::CommsData::ConsumerDisconnected(device_id) => {
```
jerrysxie, line 171: Trying to understand the states here:

- When any consumer is disconnected, we set the charge current to 0. Do we need to check the device ID?
- When consumer is connected, then we set charge current to half of the consumer's current. Why can't we send it call into the charger.rx directly in this other case. Why did we need to signal a different Embassy task to do so? Because of the 1 second wait?

Is this is a simplified state machine just for the demo? It would be helpful to put in the PR description what this PR is trying to do.

jerrysxie (changes requested):

RobertZ2011:
Should this be closed? I think the refactor to place the charger functionality into the power policy supersedes this.

PR #187, opened 2025-03-20
Make `receive()` produce a `Result`

felipebalbi:
In this big PR, we're making sure `receive()` produces a `Result` that
can be checked by the caller. Ideally, caller would propagate the error
all the way up the call chain, but we're not there yet.


felipebalbi:
Fails because some dependencies are still private.

embedded-service/src/hid/mod.rs
```diff
@@ -221,6 +221,22 @@ impl MailboxDelegate for Device {
...
+            .data
+            .get::<Message>()
+            .ok_or(comms::MailboxDelegateError::MessageNotFound)?;
+
+        match message.data {
+            MessageData::Request(ref request) => {
+                self.request.signal(request.clone());
+                Ok(())
+            }
+            _ if message.id != self.id => Err(comms::MailboxDelegateError::InvalidId),
```
jerrysxie, line 224: I believe this case is not an error, it is a linked list of devices that we are iterating over. The device with the matched ID acts on the message; the unmatched device will ignore the message.

examples/std/src/bin/hid.rs
```diff
@@ -45,6 +45,20 @@ impl comms::MailboxDelegate for Device {
...
         info!("{:?} got message", self.id);
     }
+
+    fn receive2(&self, message: &comms::Message) -> Result<(), comms::MailboxDelegateError> {
+        let message = message
+            .data
+            .get::<hid::Message>()
+            .ok_or(comms::MailboxDelegateError::MessageNotFound)?;
+
+        if message.unwrap().id != self.id {
```
jerrysxie, line 44: Not matching id just means skip over this element in the linked list.

hid-service/src/i2c/host.rs
```diff
@@ -305,4 +305,23 @@ impl<B: I2cSlaveAsync> MailboxDelegate for Host<B> {
...
+            .data
+            .get::<hid::Message>()
+            .ok_or(comms::MailboxDelegateError::MessageNotFound)?;
+
+        match hid_msg.data {
+            hid::MessageData::Response(ref response) => {
+                self.response.signal(response.clone());
+                Ok(())
+            }
+            _ if message.to != EndpointID::External(External::Host) => {
```
jerrysxie, line 304: I don't think this is possible, unless we register this EP with the wrong ID.

hid-service/src/i2c/host.rs
```diff
@@ -305,4 +305,23 @@ impl<B: I2cSlaveAsync> MailboxDelegate for Host<B> {
...
+                Err(comms::MailboxDelegateError::InvalidDestination)
+            }
+            _ if hid_msg.id != self.id => Err(comms::MailboxDelegateError::InvalidId),
```
jerrysxie, line 307: Not matching ID should mean just skip this entry in the linked list.

jerrysxie (changes requested):

embedded-service/src/comms.rs
```diff
@@ -151,7 +151,34 @@ pub struct Message<'a> {
...
+
+    /// Invalid ID
+    InvalidId,
+
+    /// Invalid data
+    InvalidData,
+
+    /// Other error. Usually related to the underlying device or
+    /// transport.
+    Other,
```
JamesHuard, line 181: should this be parameterized over something like a u32/usize for error code prop?

embedded-service/src/hid/mod.rs
felipebalbi: Isn't the top-level router already making sure that only properly routed messages reach the destination? In that case, this would be an error, indeed. It would mean the router failed?

jerrysxie: No, the top-level router just make sure that we are routing the messages to right type of endpoint: https://github.com/OpenDevicePartnership/embedded-services/blob/0e41d0ba8b01c71306376564326aeb33ce6b6347/embedded-service/src/comms.rs#L280

For this instance, the id is added as part of the HID message: https://github.com/OpenDevicePartnership/embedded-services/blob/0e41d0ba8b01c71306376564326aeb33ce6b6347/embedded-service/src/hid/mod.rs#L279

The upper router only knows endpoint type, not sub-id within each endpoint type: https://github.com/OpenDevicePartnership/embedded-services/blob/0e41d0ba8b01c71306376564326aeb33ce6b6347/embedded-service/src/comms.rs#L19, we can consider adding an ID to each of these endpoint type.

felipebalbi: fair enough, I'll update.

embedded-service/src/comms.rs
felipebalbi: I'd rather not propagate raw `u32` values. It could be parameterized with some other Error, but not a raw `u32`. Let me review the users, maybe there's something we can use for this.

hid-service/src/i2c/host.rs
felipebalbi: and because we don't check this in compile time, there's a possibility that someone would get this wrong. Additionally, this is a direct conversion of the original code without error:

https://github.com/OpenDevicePartnership/embedded-services/blob/aac544e4281ac0a2c6ee12f9fece9957e52f46cc/hid-service/src/i2c/host.rs#L293-L307

embedded-service/src/hid/mod.rs
jerrysxie: @RobertZ2011 I believe in the HID service, if we are not matching the HID ID, it is not considered to be an error, correct?

So maybe we need a different kind of error that the router layer can ignored, like NotMyRequestError?

RobertZ2011: The HID code maintains a list of HID devices and the function to look-up a device by its ID does return an error if no device is found. However, that's different than the message passing. There's nothing in the HID code that will result in an error if you broadcast to a non-existent device, each device is responsible for listening for messages addressed to it. Maybe we could change the result type to `Result<bool, Infallible>` and return true/false based on if any handlers responded so the caller can determine if it wants to consider that an error.

felipebalbi: It's not `Infallible`, though. Many services are printing error messages in various conditions. Instead of printing an error message, we should produce the error. Caller can still parse the error type to make decisions on what to do next, but I don't think the service's `receive()` method should be the one making the decision, we can leave that to the caller, which is $this PR does.

Currently, `route()` continues to chug along the list, as it is prior to $this PR, but that decision can change in the future.

RobertZ2011: Talked with Felipe and I'm fine with the changes in this PR. But we're also going to talk with Jimi about adding various generic message handler implementations so that specific services can pick the exact behavior that they want.

RobertZ2011 (approved):

embedded-service/src/comms.rs
felipebalbi: Looked a little closer regarding this one. Currently only HID service and Battery service utilize the `Other` variant, none of which provide extra information to pass along to the caller. I'll take a mental note for this in case we need it in the future. For now, `Other` remains just that.

embedded-service/src/hid/mod.rs
felipebalbi: No, that's incorrect. Currently the error propagation still stops within comms module, it doesn't propagate all the way to caller. Once that part is carried out, then all services will have to be updated.

PR #202, opened 2025-03-27
cfu-service, embedded-service changes to match the updated cfu definitions

ankurung:
cfu-service, embedded-service changes to match the updated cfu definitions

### DIFF
--- a/cfu-service/Cargo.toml
+++ b/cfu-service/Cargo.toml
@@ -9,7 +9,6 @@ embedded-cfu-protocol = { git = "https://github.com/OpenDevicePartnership/embedd
 embedded-services = { path = "../embedded-service" }
 
 # External dependencies
-binary_serde = "1.0.24"
 embassy-time = { git = "https://github.com/embassy-rs/embassy" }
 embassy-sync = { git = "https://github.com/embassy-rs/embassy" }
 embassy-executor = { git = "https://github.com/embassy-rs/embassy" }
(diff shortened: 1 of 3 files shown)


cfu-service/Cargo.toml
```diff
@@ -9,29 +9,9 @@ embedded-cfu-protocol = { git = "https://github.com/OpenDevicePartnership/embedd
...
-embassy-time = { git = "https://github.com/embassy-rs/embassy" }
 embassy-sync = { git = "https://github.com/embassy-rs/embassy" }
 embassy-executor = { git = "https://github.com/embassy-rs/embassy" }
-log = { version = "0.4.14", optional = true }
-defmt = { version = "0.3", optional = true }
 heapless = "0.8.*"
 
 [features]
 default = []
-defmt = [
```
jerrysxie, line 22: We want to keep the defmt and log features for tracing and testing.

jerrysxie (changes requested):

cfu-service/Cargo.toml
ankurung: Curious, machete flaged as not used.  Still keep them?

jerrysxie: Not familiar with `machete` but tracing macro are definitely in use: https://github.com/OpenDevicePartnership/embedded-services/blob/0e41d0ba8b01c71306376564326aeb33ce6b6347/cfu-service/src/lib.rs#L36

Though it is not enabled by default, and both features are optional. Does `machete` only check default features?

felipebalbi (approved):

JamesHuard (approved):

PR #204, opened 2025-03-27
Fix unit test failure and clippy warning

jerrysxie:
Due to previous broken workflows, missed a few things.

- Fix clippy warning due to not documented version construct
- Fix unit test due to updated battery struct
- Fix missing license field in Cargo.toml

### DIFF
--- a/embedded-service/src/ec_type/generator/ec-memory-generator.py
+++ b/embedded-service/src/ec_type/generator/ec-memory-generator.py
@@ -3,6 +3,7 @@ import sys,yaml
 # Function to convert YAML data to Rust structures
 def yaml_to_rust(data):
     rust_code = "//! EC Internal Data Structures\n\n"
+    rust_code += "#[allow(missing_docs)]\n"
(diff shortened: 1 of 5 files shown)


jerrysxie:
Single failure due to missing license field in `cfu-service` Cargo.toml, fixed in later commit.

embedded-service/src/ec_type/generator/ec-memory-generator.py
```diff
@@ -3,6 +3,7 @@
 # Function to convert YAML data to Rust structures
 def yaml_to_rust(data):
     rust_code = "//! EC Internal Data Structures\n\n"
+    rust_code += "#[allow(missing_docs)]\n"
```
felipebalbi, line 6: You could just add a simple documentation:

```rust
/// Embedded Controller Memory Map Version
```

felipebalbi (approved):
Approving, but left a comment. Safe to ignore.

tullom (approved):

PR #207, opened 2025-03-27
Set up Cargo workspace and standardize dependencies

dymk:
### What changes does this PR introduce?
This PR sets up a Cargo workspace at the repository root and standardizes dependencies across all service crates. It unifies dependency management and makes it easier to maintain consistent versions across crates within the project.

### Key changes:
1. Added a root `Cargo.toml` with workspace configuration
2. Moved common dependency specifications to workspace-level
3. Updated all service crates to use workspace dependencies
4. Added `cargo test` to GitHub Actions
5. Check `no_std` targets compile
6. Check `rt633` example compiles - the other examples are broken, and can be fixed in a future PR.
7. Cleaned up individual `Cargo.toml` files for consistent formatting
8. Exclude `example` directories from the workspace as they are effectively separate projects
9. Commit `Cargo.lock` to prevent future breakages due to upstream crates
 
### Why?
- Simplifies dependency management across all crates
- Ensures consistent dependency versions
- Makes it easier to update dependencies in one place
- Improves build times through shared target directories
- Ensure tests and builds do not break if changes are made cross-package via CI/CD


.github/workflows/check.yml
```diff
@@ -53,25 +53,23 @@ jobs:
 
   fmt:
     runs-on: ubuntu-latest
-    name: nightly / fmt
+    name: stable / fmt
```
felipebalbi, line 56: rustfmt on nightly is a concious decisions because of our of module level `use`

Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+embedded-storage = "0.3"
+embedded-storage-async = "0.4.1"
+embedded-usb-pd = { git = "https://github.com/OpenDevicePartnership/embedded-usb-pd", default-features = false }
+fixed = "1.23.1"
+heapless = "0.8.*"
+log = "0.4"
+postcard = "1.*"
+rand_core = "0.6.4"
+serde = { version = "1.0.*", default-features = false }
+tps6699x = { git = "https://github.com/asasine/tps6699x", branch = "remove-controller-trait" }
```
asasine, line 55: One https://github.com/OpenDevicePartnership/tps6699x/pull/17 merges, this can be the default branch.

embedded-service/Cargo.toml
```diff
@@ -5,53 +5,42 @@ edition = "2021"
...
+embedded-storage.workspace = true
+embedded-usb-pd.workspace = true
+fixed.workspace = true
+heapless.workspace = true
+log = { workspace = true, optional = true }
+postcard.workspace = true
+rand_core.workspace = true
+serde.workspace = true
+
+[target.'cfg(target_os = "none")'.dependencies]
```
asasine, line 37: This is a nice solution for the linking problem.

examples/rt633/Cargo.toml
```diff
@@ -1,3 +1,5 @@
+[workspace]
```
asasine, line 1: Why is this a workspace? It's the only crate in this directory. Same Q for the other example crates.

rustfmt.toml
```diff
@@ -1,3 +1 @@
-group_imports = "StdExternalCrate"
-imports_granularity = "Module"
```
asasine, line 2: I think these should be added back once .github/workflows/check.yml goes back to using nightly. Same for the other rustfmt.toml

type-c-service/src/driver/tps6699x.rs
```diff
@@ -40,7 +40,7 @@ impl<'a, const N: usize, M: RawMutex, B: I2c> Controller for Tps6699x<'a, N, M,
         let interrupts = self.tps6699x.wait_interrupt(false, |_, _| true).await;
 
         for (i, (interrupt, event)) in zip(interrupts.iter(), self.port_events.iter_mut()).enumerate() {
-            trace!("Interrupt {}: {:#X}", i, interrupt);
+            trace!("Interrupt {}: {:#?}", i, interrupt);
```
asasine, line 43: I believe that printing this as hex was intentional since `interrupt` is a bitflag over a `u32`. Using debug would likely print the integer as decimal, which is less useful for bitflags.

asasine (commented):
Some changes would be good but not blocking on them.

examples/rt633/Cargo.toml
dymk: This and the other crates in `example` otherwise think they're part of the parent workspace, which causes issues as they define binaries with the same names (e.g. both this and `std` define a `keyboard` binary). This forces them to be in their own separate workspace, which isolates their respective `target` directories and avoids the conflict. I do suspect there's a better way to restructure this so they can be part of the parent workspace, but that's a more invasive change.

.github/workflows/check.yml
dymk: Is that particularly important to us? A dependency on `nightly` for day to day devex tasks should be a high bar.

type-c-service/src/driver/tps6699x.rs
dymk: The print was broken in `~main` due to the type of `interrupt` not implementing `UpperHex` for all targets, likely due to a feature that would be enabled sometimes and not other times. There probably is a way to ensure that trait exists for the type in all build configurations, but this was the least-invasive change in the meantime.

rustfmt.toml
dymk: If they're nightly features, I suggest we don't keep them. There's recent activity on the PRs tracking i.e. the imports_granularity feature, and our code formatters changing a rule out from under us will cause future devex and CI woes.

jerrysxie (approved):

asasine (approved):

jerrysxie (approved):

.github/workflows/check.yml
felipebalbi: Only really important for embassy HALs, really.

felipebalbi (approved):

PR #208, opened 2025-03-28
type-c-service: fix build errors

felipebalbi:
Trivial fix for build errors.

### DIFF
--- a/type-c-service/src/wrapper.rs
+++ b/type-c-service/src/wrapper.rs
@@ -200,7 +200,7 @@ impl<'a, const N: usize, C: Controller> ControllerWrapper<'a, N, C> {
     /// Process a power command
     /// Returns no error because this is a top-level function
     async fn process_power_command(&self, controller: &mut C, port: LocalPortId, command: RequestData) {
-        trace!("Processing power command: device{} {:#?}", port, command);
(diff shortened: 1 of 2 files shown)


RobertZ2011 (approved):

type-c-service/src/wrapper.rs
```diff
@@ -200,7 +200,7 @@ impl<'a, const N: usize, C: Controller> ControllerWrapper<'a, N, C> {
     /// Process a power command
     /// Returns no error because this is a top-level function
     async fn process_power_command(&self, controller: &mut C, port: LocalPortId, command: RequestData) {
-        trace!("Processing power command: device{} {:#?}", port, command);
+        trace!("Processing power command: device{} {:#?}", port.0, command);
```
asasine, line 203: `LocalPortId` derives `defmt::Format` and I don't see any circumstances where the `embedded-usb-pd/defmt` feature isn't set. What am I missing?

asasine (approved):

type-c-service/src/wrapper.rs
felipebalbi: you're missing the fact that `defmt` is not the only logging crate. Disable `defmt` and compile with `log`. ;)

jerrysxie (approved):

PR #211, opened 2025-04-01
Move charging logic into power-policy service

### LINKED ISSUE #179: battery: move charger functionality to power-policy-service
For OEMs, the charger acts more like a power delivery device rather than a battery charger. Therefore, it makes more sense to move the charger to the power policy service so it can more easily interface with the power policy state machines. This also allows for devices without a battery, which usually still have a charging IC to regulate system voltage, to completely omit the battery service.

tullom:
Move charging logic out of battery service and into power policy service.

Basic device state machine structure is created. Drivers need to impl the `ChargeController` trait to define state logic and get `embedded_batteries` support.

Closes #179


felipebalbi (changes requested):
Check clippy :)

embedded-service/src/power/policy/charger.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn init_charger(&mut self) -> impl Future<Output = Result<(), Self::BusError>>;
+    fn is_psu_attached(&mut self) -> impl Future<Output = Result<bool, Self::BusError>>;
+}
+
+/// Charger Device ID new type
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct ChargerId(pub u8);
+
+/// Charger Device ID new type
```
asasine, line 21: Copy pasta

embedded-service/src/power/policy/charger.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Charger finished initialization sequence
+    Initialized,
+    /// PSU attached and we want to switch to it
+    PsuAttached(PowerCapability),
+    /// PSU detached
+    PsuDetached,
+    /// A timeout of some sort was detected
+    Timeout,
+    /// An error occured on the bus
+    BusError,
```
asasine, line 39: Should the bus error be wrapped so the caller can inspect?

embedded-service/src/power/policy/policy.rs
```diff
@@ -105,6 +108,16 @@ pub async fn register_device(device: &'static impl device::DeviceContainer) -> R
...
 }
 
+/// Register a charger with the power policy service
+pub async fn register_charger(device: &'static impl charger::ChargerContainer) -> Result<(), intrusive_list::Error> {
+    let device = device.get_charger();
+    if get_charger(device.id()).await.is_some() {
+        return Err(intrusive_list::Error::NodeAlreadyInList);
+    }
+
+    CONTEXT.get().await.devices.push(device)
```
asasine, line 118: Should this be `chargers.push(device)`?

power-policy-service/src/wrapper.rs
```diff
@@ -4,48 +4,79 @@ use embassy_futures::select::select;
...
-        self.charger_state.set_state(new_state).await
+    pub async fn set_state(&self, new_state: charger::InternalState) {
+        self.charger_policy_state.set_state(new_state).await
+    }
+
+    async fn wait_policy_command(&self) -> PolicyEvent {
+        self.charger_policy_state.wait_event().await
+    }
+
+    async fn process_controller_event(&self, controller: &mut C, event: ChargerEvent) {
```
asasine, line 38: `controller` isn't modified. Why is it a mutable reference? It isn't even accessed so it could be removed.

power-policy-service/src/wrapper.rs
```diff
@@ -0,0 +1,137 @@
...
+                }
+                State::PsuDetached => {
+                    let res = controller.charging_current(0).await;
+                    if res.is_err() {
+                        error!("Error writing charging current to charger");
+                        self.charger_policy_state.send_event(ChargerEvent::BusError).await
+                    }
+                }
+                State::Unpowered => {
+                    self.set_state(State::Init).await;
```
sukomath, line 116: Will there be more actions in Unpowered state? If not, can we remove 'Unpowered' and go to init state directly when there is timeout or bus error?

embedded-service/src/power/policy/charger.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Intrusive list node
+    node: intrusive_list::Node,
+    /// Device ID
+    id: ChargerId,
+    /// Current state of the device
+    state: Mutex<NoopRawMutex, State>,
+    /// Channel for requests to the device
+    events: Channel<NoopRawMutex, ChargerEvent, CHARGER_CHANNEL_SIZE>,
+    // /// Channel for responses from the device
+    // response: Channel<NoopRawMutex, InternalResponseData, CHARGER_CHANNEL_SIZE>,
```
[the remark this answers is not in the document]
tullom: Makes sense, implemented.

power-policy-service/src/wrapper.rs
```diff
@@ -0,0 +1,106 @@
...
+};
+
+pub struct ControllerWrapper<C: ChargeController> {
+    charger_state: charger::Device,
+    controller: RefCell<C>,
+}
+
+const STATE_MACHINE_TIMEOUT_SECS: u64 = 10;
+
+impl<C: ChargeController> ControllerWrapper<C> {
```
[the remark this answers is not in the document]
tullom: Done. Renamed module to `charger` and struct to `Wrapper`

embedded-service/src/power/policy/charger.rs
```diff
@@ -0,0 +1,136 @@
...
+    PsuDetached,
+    /// Device is discharging battery
+    Unpowered,
+    // TODO: Dead battery revival?
+    /// OEM specific state(s)
+    Oem(OemStateId),
+}
+
+/// Channel size for device requests
+pub const CHARGER_CHANNEL_SIZE: usize = 2;
```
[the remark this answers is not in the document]
tullom: Done.

embedded-service/src/power/policy/charger.rs
```diff
@@ -0,0 +1,136 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ChargerEvent {
+    /// Charger finished initialization sequence
+    Initialized,
+    /// PSU attached and we want to switch to it
+    PsuAttached(PowerCapability),
+    /// PSU detached
+    PsuDetached,
+    /// A timeout of some sort was detected
+    Timeout,
```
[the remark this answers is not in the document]
tullom: Its technically also an event that can occur from the device side so i think it should stay here.

embedded-service/src/power/policy/charger.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// Charger Device ID new type
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct OemStateId(pub u8);
+
+/// Data for a device request
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ChargerEvent {
```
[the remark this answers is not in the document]
tullom: As discussed offline, the state machine implemented here is managing the hardware device moreso than policy state. So it should stay.

embedded-service/src/power/policy/charger.rs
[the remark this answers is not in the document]
tullom: Found a use for it, used it 😉

embedded-service/src/power/policy/charger.rs
```diff
@@ -0,0 +1,136 @@
+use core::{future::Future, ops::DerefMut};
+
+use embassy_sync::{blocking_mutex::raw::NoopRawMutex, channel::Channel, mutex::Mutex};
+
+use crate::intrusive_list;
+
+use super::PowerCapability;
+
+pub trait ChargeController: embedded_batteries_async::charger::Charger {
```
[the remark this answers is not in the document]
tullom: Done.

embedded-service/src/power/policy/policy.rs
tullom: Great catch, saved me an hour of debug time probably LOL.

power-policy-service/src/wrapper.rs
tullom: The plan is to interface with the charger more fully here, especially if OEM partners want to do something on the hardware. This state machine is incomplete and more functionality will be added as the charger requirements become clearer.

power-policy-service/src/wrapper.rs
tullom: Yeah good catch. I removed Unpowered and instead made the state machine go directly to Init,

embedded-service/src/power/policy/charger.rs
RobertZ2011: One of the goals of using messaging is to erase generic types. There isn't much that the power policy can do with an actual error object and it would lock the power policy into a specific bus error type so we couldn't have the power policy simultaneously use an I2C device and a SPI device. This also means that the of handling and logging errors is left to the driver which can handle any the device-specific functionality.

embedded-service/src/power/policy/charger.rs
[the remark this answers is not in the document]
RobertZ2011: If it's not an error then it can stay here, but if it's also an error then it should be present separately in the error type.

RobertZ2011 (approved):
General approval pending unresolved comments and clippy warnings.

embedded-service/src/power/policy/charger.rs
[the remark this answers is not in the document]
tullom: Done. Added a `ChargerError` type that gets returned with device responses to the main power policy state machine.

embedded-service/src/power/policy/charger.rs
tullom: Done.

embedded-service/src/power/policy/charger.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/// Data for a device request
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ChargerResponse {
+    /// Command completed
+    Ack,
+    /// Something went wrong
+    Nack(ChargerError),
```
RobertZ2011, line 81: Instead of putting this variant here I would instead use a `Result<ChargerResponse, ChargerError>` where it makes sense to. Then I would also provide a type alias for `Result<ChargerResponse, ChargerError>` to use with the response channel to make things a little clearer. See https://github.com/OpenDevicePartnership/embedded-services/blob/5e91a91b6ea7010bcb8694ae1f15f1db9f201072/embedded-service/src/type_c/controller.rs#L114 for an example in the type-C code.

RobertZ2011 (changes requested):

RobertZ2011 (dismissed):
Approved pending clippy fixes.

felipebalbi (approved):

RobertZ2011 (approved):

PR #212, opened 2025-04-02
Add send_device_request method on CfuDevice + FinalizeUpdate enum content

Ctru14:
Adding send_device_request() method on CfuDevice in order to send a message to itself without blocking and waiting for a response. This is an alternative method to execute_device_request(). Where execute_device_request should be used for requesting data from other devices, send_device_request selects the next execution path from the same object.

Along with this, the FinalizeUpdate entry for RequestData also takes in a FwUpdateContentCommand for routing and replying after a content message.

### DIFF
--- a/embedded-service/src/cfu/component.rs
+++ b/embedded-service/src/cfu/component.rs
@@ -70,7 +70,7 @@ pub enum RequestData {
     /// Request for component to prepare itself for an update
     PrepareComponentForUpdate,
     /// Request for component to execute any logic needed to finalize update
-    FinalizeUpdate,
+    FinalizeUpdate(FwUpdateContentCommand),
 }
 
 /// CFU Response types and necessary data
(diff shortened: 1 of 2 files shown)


JamesHuard (dismissed):

embedded-service/src/cfu/component.rs
```diff
@@ -135,15 +135,23 @@ impl CfuDevice {
...
     }
+    
     /// Setter for component state
     /// Intended to be used to auto-block updates if one is in-progress
     pub async fn state(&self) -> InternalState {
         *self.state.lock().await
     }
+    
+    /// Sends a request to this device without waiting for a response
+    pub async fn send_device_request(&self, request: RequestData) {
```
RobertZ2011, line 151: What's the use case for this function? My initial concern is that the receiver doesn't know that this request isn't waiting for a response so will still produce one. Then the response will stay in the channel until a subsequent wait for a response will get the stale response. A subsequent non-waiting request will also get blocked because the response channel will be full. If the goal is to perform FW updates across components simultaneously (e.g. send a request to each component and then check until they're all complete) then we should use something like `join` to execute those futures concurrently. If the goal is to have a request that immediately completes without an error then we should introduce an `Ack` or `Complete` variant to the response type and have the receiver immediately return that.

RobertZ2011 (changes requested):

Ctru14:
Due to above thoughts, went a different route for implementation and abandoning this PR in favor of https://github.com/OpenDevicePartnership/embedded-services/pull/220

PR #223, opened 2025-04-07
Add Basic NVRAM Platform Service Abstraction

JamesHuard:
Adds a preliminary platform services crate for managing and abstracting common mechanisms like NVRAM without imposing knowledge of underlying implementer on users of said features.

Has a temporary work around for missing GPREG management APIs in embassy-imxrt

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "embedded-service",
     "espi-service",
     "hid-service",
+    "platform-service",
     "power-button-service",
     "power-policy-service",
     "storage-bus",
--- a/examples/rt685s-evk/src/bin/nvram.rs
+++ b/examples/rt685s-evk/src/bin/nvram.rs
@@ -1,0 +1,47 @@
+#![no_std]
+#![no_main]
+
+extern crate rt685s_evk_example;
+
+use {defmt_rtt as _, panic_probe as _};
+
+#[embassy_executor::main]
+async fn main(_spawner: embassy_executor::Spawner) {
+    let _p = embassy_imxrt::init(Default::default());
+
+    use platform_service::nvram;
+
+    #[repr(usize)]
+    enum Entries {
+        // These sections are unusable:
+        //     0,
+        //     1,
+        //     2,
+        General = 3,
+        Special = 4,
+    }
+
+    static NVRAM_TABLE: nvram::Table<2> = nvram::Table::new(&[Entries::General as usize, Entries::Special as usize]);
+
+    embedded_services::init().await;
+    nvram::init(&NVRAM_TABLE).await.unwrap();
+
+    let mut general_section = nvram::lookup_section(NVRAM_TABLE.get_index(Entries::General as usize).unwrap())
+        .await
+        .unwrap();
+
+    general_section.write(0);
+    general_section.write(1);
+
+    use defmt::info;
+
+    info!("general_section = {:?}", general_section.read());
+
+    let special = nvram::lookup_section(NVRAM_TABLE.get_index(Entries::Special as usize).unwrap())
+        .await
+        .unwrap();
+    info!("special = {:?}", special.read());
+
+    let untouchable: Option<nvram::ManagedSection> = nvram::lookup_section(10).await;
+    info!("Attempted invalid section is_none = {:?}", untouchable.is_none());
+}
--- a/platform-service/src/imxrt/mod.rs
+++ b/platform-service/src/imxrt/mod.rs
@@ -1,0 +1,28 @@
+use core::ops::Range;
+use embassy_imxrt::pac;
+
+pub(crate) fn nvram_read(address: usize) -> u32 {
+    // TODO introduce RTC GPReg interface directly from embassy_imxrt
+
+    // SAFETY: safe from single executor
+    let rtc = unsafe { &*pac::Rtc::ptr() };
+
+    rtc.gpreg(address).read().bits()
+}
+
+pub(crate) fn nvram_write(address: usize, value: u32) {
+    // TODO introduce RTC GPReg interface directly from embassy_imxrt
+
+    // SAFETY: safe from single executor
+    let rtc = unsafe { &*pac::Rtc::ptr() };
+
+    rtc.gpreg(address).write(|w|
+        // SAFETY: safe from single executor 
+        unsafe { w.bits(value) });
+}
+
+pub(crate) fn nvram_valid_range() -> Range<usize> {
+    // TODO introduce RTC GPReg interface directly from embassy_imxrt
+    // indices 0, 1, 2 are utilized by timer_driver in embassy_imxrt
+    3..8
+}
--- a/platform-service/src/lib.rs
+++ b/platform-service/src/lib.rs
@@ -1,0 +1,32 @@
+#![no_std]
+
+/// NVRAM platform service abstraction
+pub mod nvram;
+
+#[cfg(any(feature = "imxrt", feature = "imxrt685"))]
+pub(crate) mod imxrt;
+
+#[cfg(any(feature = "imxrt", feature = "imxrt685"))]
+pub(crate) use imxrt::*;
+
+#[cfg(not(any(feature = "imxrt", feature = "imxrt685")))]
+mod defaults {
+    use core::ops::Range;
+
+    pub(crate) fn nvram_read(_address: usize) -> u32 {
+        0
+    }
+    pub(crate) fn nvram_write(_address: usize, _value: u32) {}
+
+    pub(crate) fn nvram_valid_range() -> Range<usize> {
+        0..0
+    }
+}
+
+#[cfg(not(any(feature = "imxrt", feature = "imxrt685")))]
+pub(crate) use defaults::*;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+}
--- a/platform-service/src/nvram.rs
+++ b/platform-service/src/nvram.rs
@@ -1,0 +1,106 @@
+use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
+use embassy_sync::blocking_mutex::Mutex;
+use embassy_sync::once_lock::OnceLock;
+
+use crate::nvram_valid_range;
+
+// Describes a u32 section of non-volatile RAM
+struct Info {
+    // the starting address/offset and length in u32 words of the section
+    offset: usize,
+
+    guard: Mutex<CriticalSectionRawMutex, ()>,
+}
+
+/// Section descriptor table for linking indices to offsets
+pub struct Table<const N: usize> {
+    sections: [Info; N],
+}
+
+impl<const N: usize> Table<N> {
+    /// Generate a Section descriptor table of length N
+    pub const fn new(valid_offsets: &[usize; N]) -> Self {
+        let mut as_info: [Info; N] = [const { Info::new(0) }; N];
+
+        // have to while loop here for const fn
+        let mut i = 0;
+        while i < N {
+            as_info[i].offset = valid_offsets[i];
+
+            i += 1;
+        }
+
+        Self { sections: as_info }
+    }
+
+    /// Get the index associated with given offset, for passing context key around as needed
+    pub fn get_index(&self, offset: usize) -> Option<usize> {
+        self.sections
+            .iter()
+            .enumerate()
+            .find_map(|(index, info)| if info.offset == offset { Some(index) } else { None })
+    }
+}
+
+/// Table initialization errors
+#[derive(Copy, Clone, Debug)]
+pub enum TableError {
+    /// Offset constructed is not accessible per NVRAM implementation
+    InvalidOffset(usize),
+}
+
+impl Info {
+    const fn new(offset: usize) -> Self {
+        Self {
+            offset,
+            guard: Mutex::new(()),
+        }
+    }
+}
+
+/// Guarded handle to section
+pub struct ManagedSection {
+    info: &'static Info,
+}
+
+impl ManagedSection {
+    const fn new(info: &'static Info) -> Self {
+        Self { info }
+    }
+
+    /// Attempt to read an offset from a section. Returns None if the offset is invalid or the section is marked inaccessible
+    pub fn read(&self) -> u32 {
+        self.info.guard.lock(|_| crate::nvram_read(self.info.offset))
+    }
+
+    /// Attempt to write value to offset within section. Returns read(offset) on success, None on failure
+    pub fn write(&mut self, value: u32) {
+        // mutex guard ensures only one writer to this region at a time
+        self.info.guard.lock(|_| crate::nvram_write(self.info.offset, value));
+    }
+}
+
+static LAYOUT: OnceLock<&'static [Info]> = OnceLock::new();
+
+/// Initialize the NVRAM service with a given table of named sections
+pub async fn init<const N: usize>(table: &'static Table<N>) -> Result<(), TableError> {
+    for entry in &table.sections {
+        if !nvram_valid_range().contains(&entry.offset) {
+            return Err(TableError::InvalidOffset(entry.offset));
+        }
+    }
+
+    LAYOUT.get_or_init(|| &table.sections);
+    Ok(())
+}
+
+/// General API for attempting to interact with named NVRAM sections
+pub async fn lookup_section(index: usize) -> Option<ManagedSection> {
+    let layout = LAYOUT.get().await;
+
+    if layout.len() <= index {
+        None
+    } else {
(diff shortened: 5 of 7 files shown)


Cargo.toml
```diff
@@ -5,7 +5,7 @@ members = [
     "cfu-service",
     "embedded-service",
     "espi-service",
-    "hid-service",
+    "hid-service", "platform-service",
```
JamesHuard, line 8: I've run cargo fmt multiple times, no idea why it's like this

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,76 @@
+use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
+use embassy_sync::blocking_mutex::Mutex;
+use embassy_sync::once_lock::OnceLock;
+
+pub struct Info {
+    pub name: &'static str,
+    pub start: usize,
+    pub range: usize,
```
asasine, line 8: These look a lot like a [`core::ops::Range<usize>`](https://doc.rust-lang.org/stable/core/ops/struct.Range.html)

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,76 @@
...
+        if self.info.is_accessible && offset < self.info.range {
+            self.info
+                .guard
+                .lock(|_| Some(crate::nvram_read(self.info.start + offset)))
+        } else {
+            None
+        }
+    }
+
+    pub fn write(&mut self, offset: usize, value: u32) -> Option<u32> {
```
asasine, line 44: Can you add docs on what this returns and in what circumstances it returns `None`? Same for `read`

platform-service/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+#[cfg(not(feature = "imxrt"))]
+mod defaults {
+    pub(crate) fn nvram_read(_address: usize) -> u32 {
+        0
+    }
+    pub(crate) fn nvram_write(_address: usize, _value: u32) {}
+}
+
+#[cfg(not(feature = "imxrt"))]
+pub(crate) use defaults::*;
```
asasine, line 20: Are these the description's mentioned workaround? If so, could you add a comment/TODO to the respective issue number so it can be tracked?

platform-service/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[cfg(not(feature = "imxrt"))]
+mod defaults {
+    pub(crate) fn nvram_read(_address: usize) -> u32 {
+        0
+    }
+    pub(crate) fn nvram_write(_address: usize, _value: u32) {}
+}
+
+#[cfg(not(feature = "imxrt"))]
```
asasine, line 19: This includes the feature `imxrt685`. Is that intentional or does there need to be a third mutually-exclusive module pulled in for that?

JamesHuard: ah yea good catch, I did this before fighting with the CI to make it compile without a binary specifying the chip feature

platform-service/src/nvram.rs
JamesHuard: docs added

platform-service/src/lib.rs
JamesHuard: should be fixed now. imxrt685 was introduced because of a CI gate that specifies imxrt without chip feature

platform-service/src/lib.rs
JamesHuard: Task #224

asasine (approved):

platform-service/src/imxrt/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+pub(crate) fn nvram_read(address: usize) -> u32 {
+    // TODO introduce RTC GPReg interface directly from embassy_imxrt
+
+    // SAFETY: safe from single executor
+    let rtc = unsafe { &*pac::Rtc::ptr() };
+
+    rtc.gpreg(address).read().bits()
+}
+
+pub(crate) fn nvram_write(address: usize, value: u32) {
```
madeleyneVaca, line 12: we'll always intend to fully write, never modify, correct?

madeleyneVaca (dismissed):

platform-service/src/imxrt/mod.rs
JamesHuard: Yea, as far as I can tell this is just the one u32 register for each GPREG instance, so we'd want to write the whole word at once

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,88 @@
...
+/// Describes a named section of NVRAM
+pub struct Info {
+    /// the name associated with this NVRAM section, for decoupled lookup
+    pub name: &'static str,
+
+    /// the starting address/offset and length in u32 words of the section
+    pub range: Range<usize>,
+
+    /// whether or not this block is accessible to users of the NVRAM platform service
+    pub is_accessible: bool,
```
jerrysxie, line 15: Is there any reason we would want to expose NVRAM that is not accessible?

platform-service/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[cfg(any(feature = "imxrt", feature = "imxrt685"))]
+pub(crate) mod imxrt;
+
+#[cfg(any(feature = "imxrt", feature = "imxrt685"))]
+pub(crate) use imxrt::*;
+
+#[cfg(not(any(feature = "imxrt", feature = "imxrt685")))]
+mod defaults {
+    pub(crate) fn nvram_read(_address: usize) -> u32 {
```
jerrysxie, line 14: So this is the basis of the NVRAM traits to be implemented by `embassy-imxrt`?

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,88 @@
...
+        if entry.name == name {
+            return Some(entry);
+        }
+    }
+
+    None
+}
+
+/// Initialize the NVRAM service with a given table of named sections
+pub async fn init(nvram_table: &'static [Info]) {
```
jerrysxie, line 80: Do we want to sanity check for no section with duplicated names? And no section will have overlapping regions?

It seems like we are making an assumption that NVRAM is always continuous. In the embassy-imxrt case, if time driver feature is not enabled, we technically can use the first 4 general purpose registers as well.

Perhaps we should add to NVRAM traits to return the number of available slots of u32 (time enabled, size 5, 0-4 mapping to 3-7) + (time disabled, 8, 0-7) and then just allow client to divide up into named sections starting from 0..(num_of_slot - 1).

jerrysxie (commented):
In general this looks fine, added some comments for clarification and improvements.

platform-service/src/lib.rs
JamesHuard: Not necessarily-- this is more a temporary interface until we nail down a proper API for the RTC in embassy-imxrt. After that point, we'll probably want to restructure to have a platform specific init() for nvram that hands over the initialized RTC peripheral, like we've done for other services, and gate that behind a oncelock/mutex in the service here. If those "proper" APIs have better restrictions or patterns, we'll likely want to update this "glue" function to represent that RTC trait/impl.

platform-service/src/nvram.rs
JamesHuard: I was primarily thinking of book-keeping in the product instantiation of the service, but taking up resources for that is technically unnecessary. A comment in the product init is probably as effective

platform-service/src/nvram.rs
JamesHuard: That's fair. I went back and forth on interface in general, opting for string mainly as it wouldn't require subsystems/services that use this interface to "know" what index they need. But that doesn't address the offset into the slot mapping, as you've pointed out. Having a slot->u32 map makes more sense broadly here I think, and also reduces the interface intercept to a single index/enum map we can provide to subsystems/services on init. Will make the change, thanks for the suggestion!

platform-service/src/nvram.rs
JamesHuard: removed

platform-service/src/nvram.rs
JamesHuard: changes made

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,88 @@
...
+async fn get_by_name(name: &'static str) -> Option<&'static Info> {
+    let layout = LAYOUT.get().await;
+    for entry in *layout {
+        if entry.name == name {
+            return Some(entry);
+        }
+    }
+
+    None
+}
```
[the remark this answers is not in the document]
JamesHuard: Removed string and instead using index into table as "hash"

examples/rt685s-evk/src/bin/nvram.rs
```diff
@@ -0,0 +1,47 @@
...
+use {defmt_rtt as _, panic_probe as _};
+
+#[embassy_executor::main]
+async fn main(_spawner: embassy_executor::Spawner) {
+    let _p = embassy_imxrt::init(Default::default());
+
+    use platform_service::nvram;
+
+    #[repr(usize)]
+    enum Entries {
```
jerrysxie, line 15: Might be better to show `nvram_valid_range()` here in the example.

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,114 @@
...
+impl<const N: usize> Table<N> {
+    /// Generate a Section descriptor table of length N
+    pub const fn new(valid_offsets: &[usize; N]) -> Self {
+        let mut as_info: [Info; N] = [const { Info::new(0, 0) }; N];
+
+        // have to while loop here for const fn
+        let mut id = 0;
+        while id < N {
+            as_info[id].id = id;
+            as_info[id].offset = valid_offsets[id];
```
jerrysxie, line 31: I had to stare this for bit to understand what `id` is; it has multiple meanings there. If `id` is always the same as index, do we need to store it as a separate field?

jerrysxie, line 31: We are making the assumption that offsets passed in are valid (no duplicates and no outside of the range). We should leave a comment to assert the precondition for this function or better yet add checking to make sure the offsets passed in are sane.

This can be addressed as a fix-up issue.

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,114 @@
+use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
```
jerrysxie, line 1: The constructs in this case can be unit tested. This can be addressed as a fix-up issue.

jerrysxie (changes requested):

examples/rt685s-evk/src/bin/nvram.rs
JamesHuard: I don't think I understand what you're asking for here

platform-service/src/nvram.rs
JamesHuard: We don't strictly need to store id in the Info struct. I'm not sure I follow where it has multiple meanings, but I can remove it as a struct parameter

JamesHuard: To do this we either need to make the NVRAM valid range MCU call const, or no longer allow direct static storage of this type and remove const from the constructor. I don't think either solution is desirable, hence why this function is infallible but the init() function can perform the validation at runtime over the provided block and returns an error if there's an invalid offset provided in the pre-constructed static table.

platform-service/src/nvram.rs
JamesHuard: Filed: #232

examples/rt685s-evk/src/bin/nvram.rs
jerrysxie: Sorry, might be a misunderstanding on my part.

I thought the usage model here is:
1. Client call `nvram_valid_range()` to get the range of valid slots
2. Client divides up the range and assign the slots by names/id to different sections
3. Client init nvram with this table of the sections
4. Client can look up sections and access the sections safely by name/id

Is that the usage model?

platform-service/src/nvram.rs
JamesHuard: I'll just rename it all to index given there's any confusion here

examples/rt685s-evk/src/bin/nvram.rs
JamesHuard: Ah-- I see the confusion. The intention here is for nvram_valid_range to not be exposed to the client, and instead be validated by the NVRAM service init(). Basically,

1. Client allocates table of "named" offsets
2. Client attempts registration of table with NVRAM service via service::init(&table)
3. If an invalid offset entry is detected by the nvram_valid_range() method, the service::init() will return a TableError
4. If init() passed, client can then pass on the indices to subsystems/services for later usage via construction of the ManagedSection structs (lookup_section(index))

Since we rarely seem to have const interfaces to our hardware peripherals (like those in embassy-imxrt), it seemed prudent to me that the nvram_valid_range() function is not required to be const. Hence the runtime check on init, rather than doing it during const Table construction, which would be preferable. In either case, this would still not be exposed directly to the client.

platform-service/src/nvram.rs
jerrysxie: Gotcha, I didn't notice that this function is `const`. Now I see why we cannot have error checking here. If that is the case, then we have to do the error checking in `init()` to validate the sanity of the table.

JamesHuard: renamed to index and removed from struct

jerrysxie: > We don't strictly need to store id in the Info struct. I'm not sure I follow where it has multiple meanings, but I can remove it as a struct parameter

I was staring at `as_info[id].id = id;` trying to figure out the difference between id and index.

JamesHuard: No worries-- I think it's good that you caught a proverbial logic trap in the naming :D simplifying to a single "index" name should help readability

jerrysxie (approved):

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,106 @@
...
+use embassy_sync::once_lock::OnceLock;
+
+use crate::nvram_valid_range;
+
+// Describes a u32 section of non-volatile RAM
+struct Info {
+    // the starting address/offset and length in u32 words of the section
+    offset: usize,
+
+    guard: Mutex<CriticalSectionRawMutex, ()>,
```
dymk, line 12: If `offset` is the value being protected by the guard, can you use `offset: Mutex<..., usize>` to let the compiler enforce the guard is held when the value is accessed?

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        Self { sections: as_info }
+    }
+
+    /// Get the index associated with given offset, for passing context key around as needed
+    pub fn get_index(&self, offset: usize) -> Option<usize> {
+        self.sections
+            .iter()
+            .enumerate()
+            .find_map(|(index, info)| if info.offset == offset { Some(index) } else { None })
```
dymk, line 41: e.g. there's an unguarded read of `offset` here

platform-service/src/nvram.rs
```diff
@@ -0,0 +1,106 @@
...
+pub async fn init<const N: usize>(table: &'static Table<N>) -> Result<(), TableError> {
+    for entry in &table.sections {
+        if !nvram_valid_range().contains(&entry.offset) {
+            return Err(TableError::InvalidOffset(entry.offset));
+        }
+    }
+
+    LAYOUT.get_or_init(|| &table.sections);
+    Ok(())
+}
```
dymk, line 95: `lazy_static` maybe?

platform-service/src/nvram.rs
madeleyneVaca: alternatively, why not use an AtomicUsize here?

JamesHuard: Offset is not what's being protected here. It's the NVRAM slot at offset. Once offset is set, it shouldn't be touched.

platform-service/src/nvram.rs
JamesHuard: What advantage would that have over OnceLock<> included in embassy?

platform-service/src/nvram.rs
JamesHuard: offset is not changing

JamesHuard: in no place is Info even being made mutable around mutex accesses

platform-service/src/nvram.rs
JamesHuard: This struct isn't even ever made mutable. It's only accessed via a static ref

madeleyneVaca (approved):

jeffglaum:
@williampMSFT FYI

PR #233, opened 2025-04-11
[RFC] ECS Demo

RobertZ2011:
Entity-component systems are designs based on an entity object to which multiple component objects can be attached. Each component adds a specific behavior to the entity. This is common in video games where structurally, the only difference between your character and an enemy might be that one will have a `PlayerInput` component and the other will have a `AiInput` component. This type of design can be very useful in embedded development where it's common for a single device to present many different types of functionalities (e.g. a PD controller could also be a GPIO extender). The message-based approach of services also requires an implementation to bridge from messages to function calls on a hardware device. From an ECS perspective we can interpret the aforementioned PD controller as: an entity (the base driver object), PD message bridge (component), GPIO extender message bridge (component). This allows new and optional functionalities to be added to a device without (ideally) needing to touch existing code.

Using an existing Rust ECS framework in embedded services is difficult because:
* `no-std`
* We're using async
* We're not making a video game
* It needs to integrate with our existing messaging

None of the Rust ECS frameworks I examined (`hecs` being the most prominent light-weight one mentioned). Were not compatible with the above restrictions.

A very simple ECS would look like
```rust
// Normal version
struct Entity {
    components: Vec<&dyn Component>,
}

trait Component {
    fn do_something(&self);
}

impl Entity {
    fn do_something(&self) {
        for component in &self.components {
            component.do_something();
        }
    }
}
```

To remove dynamic allocation and dynamic dispatch we can come up with a basic embedded-friendly version:
```
struct Entity<A: Component, B: Component> {
    components: (A, B),
}

impl <A: Component, B: Component> Entity<A, B> {
    fn do_something(&self) {
        self.components.0.do_something();
        self.components.1.do_something();
    }
}
```

But this doesn't generalize well because Rust doesn't currently support variadic generics. But we can layer things to get the same result:
```
struct ComponentLayer<I: Layer, C: Component> {
    inner: I,
    component: C,
}

trait Layer {
    fn do_something(&self);
}
type Stack = ComponentLayer<ComponentLayer<Entity, A>, B>;

impl<I: Inner, C: Component> Layer for ComponentLayer<I, C> {
    fn do_something(&self) {
        self.component.do_something();
        self.inner.do_something();
    }
}
```

This allows each layer to chain the next one, replacing the loop in the original example. This PR is a rough draft of an ECS that is based on these ideas. Each component is capable of producing a message and then processing that message. Normally, the message produced by each layer is bubbled up through the stack and then is sent down the stack to be processed. **But layers can be inserted anywhere into the stack to allow for filtering/modifying messages and to control how lower layers in the stack process their messages. In addition to composing functionality together this design provides a natural way for custom code to control and modify existing functionality.**

Future improvements:
* Compile-time checking of layer types: It should be possible to define layers that only work on top of specific layers, something like `struct TypedLayer<I: Layer<MessageA, MessageB>>` for a layer that will only work on top of two layers that produce messages of types `MessageA` and `MessageB`.
* Ergonomics around the message value each layer sees: this implementation uses the same type-layering concept to allow each layer to append its message type and pass things up the stack. Need to make this easier to use with something like `fn message<&T>(&self) -> Option<T>` and `fn message_mut(&self) -> Option<&mut T>` to make it easy for layers to inspect/edit the exact messages they're interested in.
* In addition to the message type M, components should have a response type R, to allow more control over exactly how a message is processed and a response is sent back.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,7 +9,7 @@
         "editor.formatOnSave": true
     },
     "rust-analyzer.check.allTargets": false,
-    "rust-analyzer.cargo.target": "thumbv8m.main-none-eabihf",
+    //"rust-analyzer.cargo.target": "thumbv8m.main-none-eabihf",
     "rust-analyzer.cargo.features": [],
     "rust-analyzer.linkedProjects": [
         "examples/rt633/Cargo.toml",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ embedded-storage = "0.3"
 embedded-storage-async = "0.4.1"
 embedded-usb-pd = { git = "https://github.com/OpenDevicePartnership/embedded-usb-pd", default-features = false }
 fixed = "1.23.1"
+frunk = { version = "0.4.3", default-features = false }
 heapless = "0.8.*"
 log = "0.4"
 postcard = "1.*"
--- a/embedded-service/Cargo.toml
+++ b/embedded-service/Cargo.toml
@@ -30,6 +30,7 @@ embedded-storage.workspace = true
 embedded-usb-pd.workspace = true
 embedded-batteries-async.workspace = true
 fixed.workspace = true
+frunk.workspace = true
 heapless.workspace = true
 log = { workspace = true, optional = true }
 postcard.workspace = true
--- a/embedded-service/src/ecs/mod.rs
+++ b/embedded-service/src/ecs/mod.rs
@@ -1,0 +1,179 @@
+//! Embedded Entity-Component System (ECS)
+use core::{
+    cell::{Ref, RefCell, RefMut},
+    future::{self, Future},
+    ops::{Deref, DerefMut},
+};
+
+use embassy_futures::select::{select, Either};
+
+/// Used to chain the results of each layer together
+pub enum List<A, B> {
+    /// Event for the first layer
+    Head(A),
+    /// Events for the rest of the layers
+    Rest(B),
+}
+
+// Type alias that may be helpful using layer results
+/*pub type StackResult1<A> = List<A, ()>;
+pub type StackResult2<A, B> = List<A, StackResult1<B>>;
+pub type StackResult3<A, B, C> = List<A, StackResult2<B, C>>;*/
+
+/// Trait to allow for borrowing a reference to the inner type
+pub trait RefGuard<T>: Deref<Target = T> {}
+
+/// Trait to allow for borrowing a mutable reference to the inner type
+pub trait RefMutGuard<T>: DerefMut<Target = T> {}
+
+/// Core component type
+pub trait Component<T> {
+    /// Event type that we're waiting for
+    type Event;
+
+    /// Wait for an event to occur
+    fn wait_event(&self, entity: &T) -> impl Future<Output = Self::Event>;
+    /// Process the event
+    fn process(&self, entity: &mut T, event: Self::Event) -> impl Future<Output = ()>;
+}
+
+/// Entity trait
+pub trait Entity {
+    /// Underlying type of the entity
+    type Inner;
+
+    /// Get a reference to the inner entity
+    fn get_entity(&self) -> impl RefGuard<Self::Inner>;
+    /// Get a mutable reference to the inner entity
+    fn get_entity_mut(&self) -> impl RefMutGuard<Self::Inner>;
+}
+
+/// Layer trait
+pub trait Layer: Entity + Component<Self::Inner> + Sized {
+    /// Process all events for the layer and layers below it
+    fn process_all(&mut self) -> impl Future<Output = ()>;
+}
+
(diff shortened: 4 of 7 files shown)


RobertZ2011:
@OpenDevicePartnership/uefi-contributors for feedback as well.

Javagedes:
I would like a bit more information / background regarding the reasoning behind the decision that ECS was an appropriate architecture for embedded work (I'm not against it, I think ECS is awesome). Have you investigated the actual need of a complex design like this? I whole heartedly support its ability to force abstraction layers, separate out the different areas of concern, parallelize code, and reduce tight coupling of code), however it has a pretty narrow range for seeing actual performance improvements. Mainly you need to be working on some subset of (1) large datasets, (2) large individual objects in the datasets, (3) Minimal fields in the objects, (4) Shared fields between different objects, (5) running the same code over and over again, like an RTOS.

I actually did a fair amount of research regarding rust and ECS in my college days; I specifically investigated the performance differences of using a data-oriented design (via ECS) over a traditional object-oriented design for usage in real-time simulations (which is arguably very similar to video games, the main purpose of ECS). The main focus was the improvements that using this design had on cache usage and reduction of cache misses when operating on large datasets.

What I learned is that the main performance benefit for ECS comes from operating on huge data sets where objects are large but the elements, we need to operate on are small (say we have an object that is 512 bytes large, but we only care about operating on two fields totaling 128 bytes. We can fit way more "128" bytes in cache than we can "512" bytes, greatly reducing cache misses (assuming large datasets).

Due to this background knowledge, and knowing the usage model regarding ECS, we actually opted *not* to use a full ECS architecture for our outward facing user interface. Instead, we pulled out the only `System` design for dependency injection to allow users to define their requirements, and we execute them if we have all of their dependencies. Feel free to take a look at [our](https://github.com/OpenDevicePartnership/uefi-sdk/tree/main/uefi_sdk/src/component) implementation of dependency injectable systems (confusingly named `component`) which is heavily inspired by [bevy](https://crates.io/crates/bevy). (specifically [bevy_ecs](https://crates.io/crates/bevy_ecs)

RobertZ2011:
@Javagedes

> I would like a bit more information / background regarding the reasoning behind the decision that ECS was an appropriate architecture for embedded work (I'm not against it, I think ECS is awesome). Have you investigated the actual need of a complex design like this? I whole heartedly support its ability to force abstraction layers, separate out the different areas of concern, parallelize code, and reduce tight coupling of code), however it has a pretty narrow range for seeing actual performance improvements. Mainly you need to be working on some subset of (1) large datasets, (2) large individual objects in the datasets, (3) Minimal fields in the objects, (4) Shared fields between different objects, (5) running the same code over and over again, like an RTOS.

I would say that the primary motivation here is that we've already been following an ECS design, just manually implemented. To give an example, take a look [at this struct](https://github.com/OpenDevicePartnership/embedded-services/blob/f50cbc34da00a7e3e7473adbd72669bf00e0f0cf/type-c-service/src/wrapper/mod.rs#L25). This is a part of our type-C controller code. We use a message passing based design and this struct handles the translation between a message and the corresponding calls on the trait object. It has a `controller::Device<'a>` and `power: [policy::device::Device; N]` members. These fields are how the code gets access to messages from the type-C subsystem (manages general type-C port state) and the power policy subsystem (manages external power flowing into and out of the system). You'll also see that there are child modules named `pd` and `power` which are pretty self-contained. This is already an ECS design where this struct attaches those components to an entity (anything that implements our `Controller` trait). The next step is to support FW update functionality here. This is something that I think _could_ be manually integrated into our existing code implementation like the existing components since it doesn't seem like a stretch to require our controllers to support FW update. But this doesn't extend well to optional functionality, like what do we do if we can use a controller as a GPIO extender? How do we support OEM code? My answer is to formalize our existing ECS patterns. To enable GPIO extender functionality all you need to do is add the corresponding component. OEMs can implement basically any custom functionality they want in their components.

As far as running the same code over and over, pretty much everything at a top-level is executing an event processing loop like:
```rust
loop {
    let event = self.wait_event().await;
    self.process_event(event).await;
}
```
So we're a good match there as well.

Performance isn't really a concern here. We don't currently have any performance issues so the motivation here is not that formalizing these designs will solve anything. Nor should there really be much, if any overhead associated with this design over doing it all manually. The bigger focus is providing a way to build modular, extendable code.


> Due to this background knowledge, and knowing the usage model regarding ECS, we actually opted _not_ to use a full ECS architecture for our outward facing user interface. Instead, we pulled out the only `System` design for dependency injection to allow users to define their requirements, and we execute them if we have all of their dependencies. Feel free to take a look at [our](https://github.com/OpenDevicePartnership/uefi-sdk/tree/main/uefi_sdk/src/component) implementation of dependency injectable systems (confusingly named `component`) which is heavily inspired by [bevy](https://crates.io/crates/bevy). (specifically [bevy_ecs](https://crates.io/crates/bevy_ecs)

I think by using message passing we're already committed to a design that is in some sense component driven so it's something that I think should be exposed.

PR #235, opened 2025-04-14
Add embedded_crc platform service with IMXRT hardware acceleration support

Ctru14:
Add embedded_crc platform service with IMXRT hardware acceleration support.

Reorganized the platform_service directory structure with a separate defaults folder with NVRAM service.

Tested via RT685 EVK examples using all compatible algorithms from the Rust CRC library definitions.


platform-service/src/embedded_crc.rs
```diff
@@ -0,0 +1,109 @@
+use crc::Algorithm;
+
+pub struct EmbeddedCrc<W: crc::Width> {
+    algorithm: &'static Algorithm<W>,
+
```
felipebalbi, line 5: I believe `cargo fmt` would remove this extra blank line.

platform-service/src/embedded_crc.rs
```diff
@@ -0,0 +1,109 @@
...
+            current_crc: None,
+        }
+    }
+
+    pub async fn calculate(&mut self, bytes: &[u8]) -> Result<u32, EmbeddedCrcError> {
+        let initial = if self.current_crc.is_none() {
+            self.algorithm.init
+        } else {
+            self.un_finalize(self.current_crc.unwrap())
+        };
```
felipebalbi, line 32: (untested)

```rust
let initial = self.un_finalize(self.current_crc.unwrap_or(self.algorithm.init));
```

Ctru14: Algorithm init shouldn't go through un_finalize when being selected. But I'll keep the unrwap_or function in mind for future scenarios

platform-service/src/embedded_crc.rs
Ctru14: Huh, I'm actually surprised, I double checked and it let it slide

examples/rt685s-evk/src/bin/embedded_crc.rs
```diff
@@ -0,0 +1,211 @@
...
+#![no_main]
+
+extern crate rt685s_evk_example;
+
+use crc::*;
+
+use {defmt_rtt as _, panic_probe as _};
+
+use platform_service::embedded_crc::{self, EmbeddedCrcError};
+// use platform_service::embedded_crc::crc_engine_init;
```
tullom, line 11: Nit: remove comment

platform-service/src/imxrt/embedded_crc.rs
```diff
@@ -0,0 +1,120 @@
+use embassy_imxrt::crc::{Config, Polynomial};
+use embassy_sync::blocking_mutex::raw::ThreadModeRawMutex;
```
tullom, line 2: We are getting build errors with references to this import, as well as linker errors to an undefined critical section. I believe we use NoopRawMutex in the other parts of the codebase, which is only safe in the context of a single executor. Need to investigate how to add support for ThreadModeRawMutex.

Ctru14: Worked with Jerry offline as well. Thanks for the suggestion! Ended up switching to Noop and tested successfully. Updating PR

platform-service/src/embedded_crc.rs
Ctru14: Removed

examples/rt685s-evk/src/bin/embedded_crc.rs
Ctru14: Removed

examples/rt685s-evk/src/bin/embedded_crc.rs
```diff
@@ -0,0 +1,210 @@
...
+        error!(
+            "ALGORITHM CHECK FAILED (CRC: 0x{:08X}, Expected = 0x{:08X})",
+            oneshot_crc, algorithm.check
+        );
+    }
+
+    // Calculate the CRC in two split function calls
+    let mut crc_handle: embedded_crc::EmbeddedCrc<u32> = embedded_crc::EmbeddedCrc::<u32>::new(algorithm);
+
+    let _half_data_crc = match crc_handle.calculate(&DATA[0..(DATA.len() / 2)]).await {
```
madeleyneVaca, line 119: nit but it'd be more straight-forward to call these something like first_half_data and second_half_data instead of an unused variable and split

examples/rt685s-evk/src/bin/embedded_crc.rs
```diff
@@ -0,0 +1,210 @@
...
+        info!("Split calculation test passed (CRC = 0x{:08X})", split_crc);
+    } else {
+        error!(
+            "RESUME TEST FAILED (Oneshot CRC = 0x{:08X}, Two shot CRC = 0x{:08X}, Expected CRC = 0x{:08X})",
+            oneshot_crc, split_crc, algorithm.check
+        );
+    }
+}
+
+fn parse_embeddedcrc_error<'a>(err: EmbeddedCrcError) -> &'a str {
```
madeleyneVaca, line 202: why this (new function) instead of implementing a trait like Format for the enum?

platform-service/src/defaults/embedded_crc.rs
madeleyneVaca, line 1: The defaults folder with these 3 files could probably go in its own commit

platform-service/src/embedded_crc.rs
madeleyneVaca: can you add this as a comment so that's clear to someone reading the code later?

platform-service/src/imxrt/embedded_crc.rs
```diff
@@ -0,0 +1,120 @@
...
+                _ => {
+                    return Err(EmbeddedCrcError::CrcErrorPolynomial);
+                }
+            }
+        }
+        32 => {
+            if algorithm.xorout != 0 && algorithm.xorout != 0xFFFFFFFF {
+                return Err(EmbeddedCrcError::CrcErrorXorOut);
+            }
+            if algorithm.poly == 0x04C11DB7 {
```
madeleyneVaca, line 48: might be useful to make some pub consts for these known algorithm.poly values to avoid future typos

supply-chain/config.toml
```diff
@@ -297,10 +361,18 @@ criteria = "safe-to-deploy"
...
+criteria = "safe-to-deploy"
+
 [[exemptions.mimxrt633s-pac]]
 version = "0.4.0"
 criteria = "safe-to-deploy"
 
+[[exemptions.mimxrt633s-pac]]
+version = "0.4.1"
+criteria = "safe-to-deploy"
+
```
madeleyneVaca, line 375: I'm not the most familiar with this supply chain file but do we need duplicates for newer versions instead of updating the versions listed in here?

platform-service/src/embedded_crc.rs
Ctru14: Comments added

examples/rt685s-evk/src/bin/embedded_crc.rs
Ctru14: Named updated to halves.

examples/rt685s-evk/src/bin/embedded_crc.rs
Ctru14: At first I didn't want to lock the Format dependency to defmt, but on second thought that seems to be what we're doing anyway. Function removed in favor of trait

platform-service/src/imxrt/embedded_crc.rs
Ctru14: Good idea, constants added

platform-service/src/embedded_crc.rs
```diff
@@ -0,0 +1,130 @@
...
+pub enum EmbeddedCrcError {
+    #[default]
+    CrcErrorUnknown,
+    CrcErrorWidth,
+    CrcErrorPolynomial,
+    CrcErrorXorOut,
+    CrcErrorMutexGet,
+}
+
+impl EmbeddedCrc<u32> {
```
jerrysxie, line 32: If we are allowing a partial calculations, shouldn't usage of this top-level object be protected by a mutex, instead of only imxrt specialization having a mutex protecting the access to the hardware crc? It seems like the protection is at the wrong level of abstraction.

supply-chain/imports.lock
```diff
@@ -745,29 +745,6 @@ version = "1.0.7"
 notes = "Simple hasher implementation with no unsafe code."
 aggregated-from = "https://hg.mozilla.org/mozilla-central/raw-file/tip/supply-chain/audits.toml"
 
-[[audits.mozilla.audits.half]]
```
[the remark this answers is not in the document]
Ctru14: I'm not sure, this happened after running `cargo vet regenerate exemptions`

platform-service/src/embedded_crc.rs
Ctru14: Multiple instances of the EmbeddedCrc object can be created to each handle their own calculation. They track their own internal state for partial calculations with the current_crc field, and they reset the hardware (or software) configuration with each calculation.
The Rust software CRC implementation does not need to share resources since each calculation recreates the CRC object. Only the mutex around the IMXRT CRC registers is needed to prevent interruptions on the hardware state

platform-service/src/lib.rs
```diff
@@ -3,25 +3,17 @@
...
-    pub(crate) fn nvram_read(_address: usize) -> u32 {
-        0
-    }
-    pub(crate) fn nvram_write(_address: usize, _value: u32) {}
-
-    pub(crate) fn nvram_valid_range() -> Range<usize> {
-        0..0
-    }
-}
+pub(crate) mod defaults;
```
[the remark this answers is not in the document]
Ctru14: Done https://github.com/OpenDevicePartnership/embedded-services/issues/248

platform-service/src/embedded_crc.rs
jerrysxie: I see thanks, for the explanation, so the usage model is that each client can create an instance of embedded_crc, and then the crc is tracked internally. When we call into the actual crc mechanism, we protect the resource there.

This was not obvious reviewing the example. Can we add a README to the create and demonstrate how this is intended to be used or add doc comment to the embedded_crc interface at the top-level? I surprised the clippy did not complain about all the pub functions not being documented. Perhaps the feature toggle is confusing it.

platform-service/src/imxrt/mod.rs
```diff
@@ -1,28 +1,6 @@
-use core::ops::Range;
-use embassy_imxrt::pac;
+pub(crate) mod nvram;
+pub(crate) use nvram::*;
```
asasine, line 2: Exporting the `nvram` module publically and exporting the items of the module publically is redundant imo. Do you intend for users of the items with the module as part of the path or not? If so, no need to `pub(crate) use nvram::*`. If not, no need to `pub(crate) mod nvram`, just `mod nvram`.

For what it's worth, the backwards-compatible change is to `mod nvram; pub(crate) use nvram::*;`

platform-service/src/imxrt/nvram.rs
```diff
@@ -0,0 +1,28 @@
+use core::ops::Range;
+use embassy_imxrt::pac;
+
+pub(crate) fn nvram_read(address: usize) -> u32 {
```
asasine, line 4: This one is subjective but if the `nvram` module is `pub(crate) mod nvram` or `pub(crate) use nvram::*`, declaring the items within as `pub(crate)` is redundant; they can be just `pub` and the parent module decides exactly how public. `pub(crate)` items within a module is useful if there's mixed public accessibility within the module.

asasine (commented):
Non-blocking suggestions around clearing up what the intended public API of the crate is

platform-service/src/embedded_crc.rs
```diff
@@ -0,0 +1,130 @@
+use crc::Algorithm;
+#[cfg(feature = "defmt")]
+use defmt::Format;
+
+pub struct EmbeddedCrc<W: crc::Width> {
+    algorithm: &'static Algorithm<W>,
```
asasine, line 6: Why does this need to be the static lifetime? In theory, the caller could use any lifetime they want, including static; static feels too restrictive.

platform-service/src/imxrt/embedded_crc.rs
```diff
@@ -0,0 +1,126 @@
...
+pub const CRC_CRC16_POLY_U32: u32 = 0x8005;
+
+// Must be called once to initialize hardware resources, then EmbeddedCrc object can run with hardware-agnostic signatures
+pub fn crc_engine_init(crc: embassy_imxrt::crc::Crc<'static>) {
+    CRC.get_or_init(|| Mutex::new(crc));
+}
+
+pub(crate) async fn crc_calculate_u32(
+    init: u32,
+    algorithm: &'static crc::Algorithm<u32>,
```
asasine, line 26: The way `algorithm` is used in this function and `crc_calcuate_u16`, this could be any `&crc::Algorithm<u32>`; constraining it to `'static` feels odd since you're not actually storing `algorithm`, just reading its fields.

platform-service/src/embedded_crc.rs
Ctru14: A static algorithm is required by the Rust CRC object https://docs.rs/crc/latest/crc/struct.Crc.html, as it's mostly used with the const algorithm definitions https://docs.rs/crc-catalog/2.4.0/src/crc_catalog/algorithm.rs.html

platform-service/src/imxrt/embedded_crc.rs
Ctru14: See above

platform-service/src/embedded_crc.rs
Ctru14: This was written to be compatible with the Rust implementation for easy translation. But even for custom algorithms, as can be seen in the RTK685 EVK example, can be created as a static variable. I'd be surprised if there was a scenario where the algorithm would need to be completely configured dynamically and where this would be a limiting factor.

platform-service/src/embedded_crc.rs
Ctru14: Sure, README added to platform-service to describe structure and usage of EmbeddedCrc object + IMXRT variant

platform-service/src/imxrt/mod.rs
Ctru14: I'm not sure here - these were Jimi's changes initially, I only changed up the folder structure

platform-service/src/imxrt/nvram.rs
Ctru14: Okay good to know! Same as above, NVRAM service was Jimi's contribution before, and I was only changing the file structure to add the default folder. We can ask him once he feels better.

supply-chain/imports.lock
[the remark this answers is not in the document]
jerrysxie: Gotcha, half 2.6.0 has not been audited yet. Since it is added to the exemption list, audits for older versions are removed.

jerrysxie (dismissed):

jerrysxie (approved):

madeleyneVaca (approved):

PR #246, opened 2025-04-16
Avoid vetting embassy dependencies pulled from GitHub

akshat2112:
`cargo vet` is primarily to audit third party dependencies from crates.io
Since embassy crates are also published on crates.io but we're using the GitHub versions, this leads to some issues.

If we mark them to have the same audit criteria as applied to other crates `audit-as-crates-io = true`, it would mean that we need to audit new versions of the dependencies every time cargo.lock is updated, which is pretty frequent for a GitHub based dependency as it uses the commit hash of the HEAD.

image

If we don't include them in config.toml, it results in a vet failure as matches are found on crates.io, but they don't match with what we're using.

image

The alternative is to mark them as not from crates.io, which is what this PR does.
We mark them as `audit-as-crates-io = false`, which is meant to be used for first party dependencies to declare them as inherently vetted, but seems to be the best option in this case.

As we move to using the published versions from crates.io, this would need to be updated again.

### DIFF
--- a/supply-chain/config.toml
+++ b/supply-chain/config.toml
@@ -11,31 +11,31 @@ url = "https://raw.githubusercontent.com/google/rust-crate-audits/main/audits.to
 url = "https://raw.githubusercontent.com/mozilla/supply-chain/main/audits.toml"
 
 [policy.embassy-executor]
-audit-as-crates-io = true
+audit-as-crates-io = false
 
 [policy.embassy-executor-macros]
-audit-as-crates-io = true
+audit-as-crates-io = false
 
 [policy.embassy-futures]
-audit-as-crates-io = true
+audit-as-crates-io = false
 
 [policy.embassy-hal-internal]
-audit-as-crates-io = true
+audit-as-crates-io = false
 
 [policy.embassy-imxrt]
-audit-as-crates-io = true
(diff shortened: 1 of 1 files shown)


felipebalbi (approved):

jerrysxie:
> `cargo vet` is primarily to audit third party dependencies from crates.io Since embassy crates are also published on crates.io but we're using the GitHub versions, this leads to some issues.
> 
> If we mark them to have the same audit criteria as applied to other crates `audit-as-crates-io = true`, it would mean that we need to audit new versions of the dependencies every time cargo.lock is updated, which is pretty frequent for a GitHub based dependency as it uses the commit hash of the HEAD.
> 
> image
> 
> If we don't include them in config.toml, it results in a vet failure as matches are found on crates.io, but they don't match with what we're using.
> 
> image
> 
> The alternative is to mark them as not from crates.io, which is what this PR does. We mark them as `audit-as-crates-io = false`, which is meant to be used for first party dependencies to declare them as inherently vetted, but seems to be the best option in this case.
> 
> As we move to using the published versions from crates.io, this would need to be updated again.

Very detailed PR description, thanks!

jerrysxie (approved):

PR #249, opened 2025-04-16
Overhaul battery service and add battery state machine support

tullom:
Overhaul battery service to remove charger hardware support now that it lives in power policy.

Add support for standard battery state machine.


battery-service/src/context.rs
```diff
@@ -0,0 +1,303 @@
...
+
+// pub trait BatterySequence {
+//     // TODO: make a macro, similar to power sequence
+//     async fn init(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+//     async fn not_present(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+//     async fn present(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+//     async fn not_operational(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+//     async fn operational(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+//     async fn polling(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+// }
```
felipebalbi, line 84: why commented? Maybe add it on another PR when you actually get to use it?

battery-service/src/context.rs
```diff
@@ -0,0 +1,303 @@
...
+        }
+        None
+    }
+
+    pub async fn register_fuel_gauge(&self, device: &'static Device) -> Result<(), intrusive_list::Error> {
+        if self.get_fuel_gauge(device.id()).is_some() {
+            return Err(embedded_services::Error::NodeAlreadyInList);
+        }
+
+        self.fuel_gauges.push(device)
```
felipebalbi, line 274: nit: looks like you want some type of Set datastructure here.

battery-service/src/context.rs
```diff
@@ -0,0 +1,303 @@
...
+//     async fn present(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+//     async fn not_operational(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+//     async fn operational(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+//     async fn polling(devices: &mut [Device], event: BatteryEvent) -> Option<State>;
+// }
+
+pub struct Context {
+    fuel_gauges: IntrusiveList,
+    state: RefCell<State>,
+    battery_event: Channel<NoopRawMutex, (BatteryEvent, u8), 1>,
```
RobertZ2011, line 89: Create a new-type for the ID here and create a new struct instead of using a tuple.

battery-service/src/context.rs
```diff
@@ -0,0 +1,303 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum OperationalSubstate {
+    Init,
+    Polling,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum BatteryEvent {
```
RobertZ2011, line 37: Can you clarify what this struct is? It looks like its events that are coming from a battery, but later you bundle it with a fuel gauge ID? Are we considering a fuel gauge and a battery to be the same thing?

battery-service/src/context.rs
```diff
@@ -0,0 +1,303 @@
...
+        }
+    }
+
+    pub fn get_state_machine_timeout(&self) -> u64 {
+        self.state_machine_timeout_ms
+    }
+
+    pub async fn process(&self) {
+        loop {
+            let (event, id) = self.battery_event.receive().await;
```
RobertZ2011, line 111: You should break this out into separate `wait_event` and `process` functions.

battery-service/src/context.rs
```diff
@@ -0,0 +1,303 @@
...
+                Some(device) => device,
+                None => {
+                    // TODO: Send error response
+                    error!("Fuel gauge with ID {} not found", id);
+                    self.battery_response.send(Err(ContextError::DeviceNotFound)).await;
+                    continue;
+                }
+            };
+            let timeout = self.get_state_machine_timeout();
+            let res = select(self.do_state_machine(device, event), Timer::after_millis(timeout)).await;
```
RobertZ2011, line 123: Using [with_timeout](https://docs.rs/embassy-time/latest/embassy_time/fn.with_timeout.html) would be clearer. But can all the timeout logic be put in `do_state_machine` and have it return a timeout error?

battery-service/src/context.rs
```diff
@@ -0,0 +1,303 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum State {
+    NotPresent,
+
+    Present(PresentSubstate),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum PresentSubstate {
```
RobertZ2011, line 23: In addition to adding some documentation comments here, can you also add a high-level description similar to https://github.com/OpenDevicePartnership/embedded-services/blob/main/docs/power-policy.md?

battery-service/src/context.rs
```diff
@@ -0,0 +1,303 @@
...
+        match self.handle_event(state.deref_mut(), event) {
+            Ok(new_state) => *state = new_state,
+            Err(err) => return Err(err),
+        }
+
+        loop {
+            match *state {
+                State::NotPresent => {
+                    info!("Initializing fuel gauge with ID {}", device.id());
+                    if device.execute_command(device::Command::Ping).await.is_err() {
```
RobertZ2011, line 200: Instead of executing this directly I would create wrapper functions in `Context` that takes the device ID as an argument. [See here for an example](https://github.com/OpenDevicePartnership/embedded-services/blob/5e3c205fb68c9d3fb37500baf3b1d93e02c7125f/embedded-service/src/type_c/controller.rs#L537).

RobertZ2011 (changes requested):

battery-service/src/context.rs
tullom: Done, see `Context::execute_device_command()`

battery-service/src/context.rs
tullom: I kinda took both pieces of advice here. I want to keep the state machine/global timeout as it was an explicit requirement from an OEM. But I agree that there should be more granular timeouts within the state machine to represent a device timeout, so i added `with_timeout()` to the `execute_device_command()` fn.

battery-service/src/context.rs
tullom: Done.

battery-service/src/context.rs
tullom: Yeah `Battery` in this context refers to a _smart battery_, which is another term for a fuel gauge. The fuel gauge and the battery are the same thing.

battery-service/src/context.rs
tullom: Done, see `BatteryEvent`.

battery-service/src/context.rs
tullom: Are you referring to the ID? it is composed within the device struct by the caller.

battery-service/src/context.rs
tullom: Done. Will add high level description to the docs in another PR.

battery-service/src/context.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    /// Get global state machine timeout.
+    fn get_state_machine_timeout(&self) -> Duration {
+        self.state_machine_timeout_ms
+    }
+
+    /// Main processing function.
+    pub async fn process(&self, event: BatteryEvent) {
+        loop {
+            let res = with_timeout(self.get_state_machine_timeout(), self.do_state_machine(event)).await;
```
jerrysxie, line 122: `with_timeout()` is convenient. However, if we timeout, then we would drop the `do_state_machine()`. We might be in the middle of a state transition and leave inconsistent state behind. Have we thought through any recovery step would be needed?

battery-service/src/context.rs
```diff
@@ -0,0 +1,344 @@
...
+        let device = match self.get_fuel_gauge(id) {
+            Some(device) => device,
+            None => {
+                // TODO: Send error response
+                error!("Fuel gauge with ID {:?} not found", id);
+                return Err(ContextError::DeviceNotFound);
+            }
+        };
+
+        match with_timeout(device.get_timeout(), device.execute_command(command)).await {
```
jerrysxie, line 333: Same comment about usage of `with_timeout()` and dropped future.

jerrysxie (commented):
Checking for understanding, in this design
- Controller has the hardware access
- Device is a SW representation of the device with the communication channels
- Wrapper ties controller and device together, takes the command from device and forwards it to controller
- Context is the overall orchestrator of the global state machine
   - Would message from application processor/OS be handled here?

battery-service/src/context.rs
```diff
@@ -0,0 +1,344 @@
...
+                Err(_) => {
+                    error!("Battery state machine timeout!");
+                    // Should be infalliable
+                    let _ = self
+                        .do_state_machine(BatteryEvent {
+                            event: BatteryEventInner::Timeout,
+                            device_id: event.device_id,
+                        })
+                        .await;
+                    self.battery_response.send(Err(ContextError::Timeout)).await;
```
tullom, line 143: @jerrysxie see here

battery-service/src/context.rs
tullom: yeah so if we timeout, we will run the state machine with the `BatteryEventInner::Timeout` command, which should be infalliable and put us back in a recovery/reinit state. See comment below.

battery-service/src/context.rs
tullom: For this future, since it is just a device command, we will stay in the same state and it should be safe from a state machine standpoint. The danger is in the command execution future being dropped, but that's up to the implementer to make sure the OEM device controlled state machine has a recovery path.

jerrysxie: We should adding a comment to make sure device trait implementor is aware of this.

jerrysxie (approved):

battery-service/src/context.rs
RobertZ2011: This code is quadratic over the number of registered devices since you iterate over the list for each registered device. I'm not too worried since I don't think we're ever going to register more than a handful of devices, but it may be a worth it to look into the `heapless` crate and see if they have a set implementation that can hold already inserted device IDs.

RobertZ2011 (approved):

battery-service/src/context.rs
tullom: Ahh i understand now. It should be very rare that the number of fuel gauges (and therefore sealed batteries) on a device will exceed 1, and if it does i don't see it going very high.

That said, [`heapless` does have a set implementation ](https://docs.rs/heapless/latest/heapless/type.FnvIndexSet.html) but that would require a compile time known fixed size. Support for runtime determined size is partially why we are using a linked list to hold the list of fuel gauges to begin with, so i think the cons of the set outweigh the pros.

battery-service/src/context.rs
tullom: Issue tracked in #256

PR #253, opened 2025-04-21
Reduce log spam from type-c service

asasine:
These periodic `INFO` logs show up approximately every second but aren't paticularly useful for most developers, only those actively working on the type-c service. As such, reducing their log levels and recommending developers use defmt's log filtering is more appropriate.

### DIFF
--- a/type-c-service/src/driver/tps6699x.rs
+++ b/type-c-service/src/driver/tps6699x.rs
@@ -69,9 +69,9 @@ impl<'a, const N: usize, M: RawMutex, B: I2c> Tps6699x<'a, N, M, B> {
         if port_status.is_connected() {
             // Determine current contract if any
             let pdo_raw = tps6699x.get_active_pdo_contract(port).await?.active_pdo();
-            info!("Raw PDO: {:#X}", pdo_raw);
+            trace!("Raw PDO: {:#X}", pdo_raw);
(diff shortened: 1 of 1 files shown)


JamesHuard (approved):

Ctru14 (approved):

type-c-service/src/driver/tps6699x.rs
```diff
@@ -69,9 +69,9 @@ impl<'a, const N: usize, M: RawMutex, B: I2c> Tps6699x<'a, N, M, B> {
         if port_status.is_connected() {
             // Determine current contract if any
             let pdo_raw = tps6699x.get_active_pdo_contract(port).await?.active_pdo();
-            info!("Raw PDO: {:#X}", pdo_raw);
+            trace!("Raw PDO: {:#X}", pdo_raw);
```
RobertZ2011, line 72: Would debug be more appropriate here? I tend to use trace for localizes exact paths through code while this is a more general status update. Though I don't know if we have any official guidance as to how we want to use each level. I also created a back log issue to see if we can configure the log level on a per-service basis: https://github.com/OpenDevicePartnership/embedded-services/issues/254.

RobertZ2011 (approved):

type-c-service/src/driver/tps6699x.rs
asasine: I think it's entirely subjective

PR #260, opened 2025-04-22
embedded-service: Create initial OEM message definitions

RobertZ2011:
Create generic message type that can be integrated into other message types to support OEM-specific functionality.


asasine:
To be honest, I don't really understand the utility of these types. The `MessageData` enum, for instance, forces the size of even the most basic primitives to balloon to the size of their larger types, and any structure that uses one would be better suited and more clear just using the appropriate primitive directly.

RobertZ2011:
The idea is to provide a standard, generic way for services to support OEM messages. A service can update its message types to look like
```rust
pub enum Message {
    ...
    Oem(oem::Message),
}
```

This allows OEM messaging to take advantage of our existing code. The OEM doesn't have to do anything to get their messages to a specific device (e.g. send it to a specific PD controller) since there's nothing special about them. These messages would ultimately be handled/produced by OEM-specific code. They can convert the `oem:Message` type to/from their own internal types. That's what `oem::MessageData` is for, it's to allow the OEM to include a single argument along with the message. Worst case scenario: they can include a byte slice and deserialize it on their end. The header is to mark which OEM a message is for and include a field to indicate the desired function/command (might not be needed by everything but seems general enough).

The advantage over doing something like:
```rust
pub enum Message<OEM> {
    ...
    Oem(OEM)
}
```

Is that:

1. One of the goals of using messaging is to erase types and this continues that pattern

2. Any number of messages from different OEM can co-exist simultaneously, with a generic type we'd either be restricted to a single OEM message type or have to deal with nested result types like what the component demo code does. The latter can be a bit inflexible, and I'd like to avoid it whenever possible.

3. Using a generic type means that the type arguments have to trickle down to things that don't need them. E.g. every driver would need to be generic over the OEM type even if they never use it.

This is basically approach two, just with a specific struct that is useful as a general message. I can also add an example to help clarify.

RobertZ2011:
I talked with Jimi a bit and I wanted to clarify the purpose behind this PR. This is not intended as a way to send messages to a specific OEM. This is to allow existing message types to integrate support for custom OEM messages. The idea is to support something like:
```
pd::send_oem_message(PORT0, Message { header: Header { id: OEM_ID, function: OEM_CMD_CUSTOM}, data: None}).await;
```

The underlying PD message types would be modified to allow them to contain an `oem::Message` value.

OEM-defined messages can then be carried through our existing messaging code. Once they arrive at a controller then it can be passed to OEM-specific code in the driver through hooks or however we end up supporting customizations there. The benefit with this approach is that the OEM doesn't have to create their own messaging or have to deal with implementing the logic of figuring out how to get their messages to the right PD controller.

PR #263, opened 2025-04-24
espi-service: add echoing of OOB messages on port1

jerrysxie:


espi-service/src/espi_service.rs
```diff
@@ -202,10 +203,31 @@ pub async fn espi_service(mut espi: espi::Espi<'static>, memory_map_buffer: &'st
...
 
-                espi.complete_port(1).await;
+                // Assume port1 is used for OOB messages and start from 0x2000_0200
+                // If this changes, we need to update the offset in the eSPI driver
+                let data = unsafe {
+                    let start_oob_data = 0x2000_0200 as *const u8 as *mut u8;
+                    let espi_oob_len = 0x200;
+
+                    slice::from_raw_parts_mut(start_oob_data, espi_oob_len)
+                };
```
asasine, line 216: Can these magic numbers be pulled from somewhere shared with the eSPI driver? Otherwise, the opportunity for UB here looks amazingly likely.

jerrysxie: I was thinking along the same lines. We would have to make some updates to the eSPI HAL driver to expose this. Currently we need this for OOB testing so it is a temporary solution for now.

Created an issue to expose API from the eSPI HAL driver: https://github.com/OpenDevicePartnership/embassy-imxrt/issues/405

asasine (approved):
Copilot's suggestion is goodness

RobertZ2011 (approved):

PR #266, opened 2025-04-25
Battery Service bug and QoL fixes, and fix battery examples

tullom:
Battery service examples were broken due to recent battery service overhaul change. This PR fixes the battery service examples and updates them to use the new state machine.

Battery bug fixes and quality of life updates are baked into this PR as a separate commit as well.

### DIFF
--- a/battery-service/src/context.rs
+++ b/battery-service/src/context.rs
@@ -80,7 +80,7 @@ pub enum ContextError {
 }
 
 /// External battery service context response.
-type BatteryResponse = Result<ContextResponse, ContextError>;
+pub type BatteryResponse = Result<ContextResponse, ContextError>;
 
 /// External battery state machine event wrapper.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -107,7 +107,7 @@ impl Context {
             state: RefCell::new(State::NotPresent),
             battery_event: Channel::new(),
             battery_response: Channel::new(),
-            state_machine_timeout_ms: Duration::from_millis(10000),
+            state_machine_timeout_ms: Duration::from_secs(120),
         }
     }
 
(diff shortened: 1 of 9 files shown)


tullom:
PR is ready for review, broken workflows are fixed. Pipelines still failing because the fixes are chunked up into commits. Once code review is finished, I will squash all commits and merge the PR as one commit

battery-service/src/device.rs
```diff
@@ -117,7 +117,19 @@ pub struct Device {
...
 
 impl Device {
-    // Get device ID.
+    pub fn new(id: DeviceId) -> Self {
+        Self {
+            node: embedded_services::Node::uninit(),
+            id,
+            command: Channel::new(),
+            response: Channel::new(),
+            dynamic_battery_cache: Cell::new(DynamicBatteryMsgs { ..Default::default() }),
```
jerrysxie, line 132: Why does these members need to be `cell`？

Please add more description to the commit message on how do these changes improve the quality of life.

battery-service/src/lib.rs
```diff
@@ -49,7 +49,7 @@ impl comms::MailboxDelegate for Service {
     }
 }
 
-static SERVICE: OnceLock<Service> = OnceLock::new();
+pub static SERVICE: OnceLock<Service> = OnceLock::new();
```
jerrysxie, line 52: Why is this public? Why do we want to expose this to all clients?

battery-service/src/device.rs
tullom: These need to be cell so that the battery caches can be changed in functions with only an immutable reference to self (interior mutability).

I added changelog to the first commit's message, describing each change.

battery-service/src/lib.rs
tullom: This needs to be exposed so other subsystems can receive response messages from the battery service. Alternatively, it provides a way of sending commands not through the comms system.  This is due to the limitation in the comms system that our `receive` fn is not async.

felipebalbi (approved):

battery-service/src/lib.rs
jerrysxie: Does `receive` fn need to be async or just have back pressure? If we allow back pressure in receive and allow the sender to retry, do we still need this? Make this `pub` expose quite a bit of API surface to all clients.

battery-service/src/context.rs
```diff
@@ -118,32 +118,30 @@ impl Context {
...
+            },
+            Err(_) => {
+                error!("Battery state machine timeout!");
+                // Should be infalliable
+                let _ = self
+                    .do_state_machine(BatteryEvent {
+                        event: BatteryEventInner::Timeout,
+                        device_id: event.device_id,
+                    })
+                    .await;
```
asasine, line 141: If that's the case, `.expect("Error type is Infallible")` is appropriate instead of swallowing the result.

battery-service/src/device.rs
```diff
@@ -117,7 +123,19 @@ pub struct Device {
...
```
asasine, line 132: `core::cell::Cell` implements `Default` if `T: Default` so this can be `Cell::default()` or `Default::default()`.

You could also use `..Default::default()` at the end of the `Self` construction to not have to specify these by hand.

examples/rt633/src/bin/espi_battery.rs
```diff
@@ -3,42 +3,195 @@
...
+        todo!()
+    }
+
+    async fn ping(&mut self) -> Result<(), Self::ControllerError> {
+        info!("Ping!");
+        Ok(())
+    }
+
+    fn get_timeout(&self) -> Duration {
+        todo!()
```
asasine, line 188: Is this a todo for this PR or really more of an `unimplemented!()`?

asasine (approved):
Just minor things

battery-service/src/context.rs
tullom: Done.

examples/rt633/src/bin/espi_battery.rs
tullom: It should be unimplemented for now. Replaced `todo!()`s with `unimplemented!()`

battery-service/src/device.rs
tullom: Oh thats neat. Changed to `Cell::default()`

battery-service/src/lib.rs
tullom: Technically we need async because we need to await a response from the battery-service. However, I looked into this and exposed a limited set of functions to interface with the battery service as module level functions. So i have made `SERVICE` private again. Good catch with this comment.

jerrysxie (approved):

asasine (approved):

PR #273, opened 2025-05-01
type-c example: Pass in interrupts by slice rather than array

tullom:
Fixes broken example and broken CI


felipebalbi (approved):
heh, a mutable slice of mutable references. Just rolls off the tongue, doesn't it? LOL

RobertZ2011 (approved):

PR #274, opened 2025-05-02
Fix rt685-evk type-c example with new embassy-imxrt i2c config

tullom:
https://github.com/OpenDevicePartnership/embassy-imxrt/pull/398 broke how we instantiate I2C buses, this PR fixes the type-c example.

### DIFF
--- a/examples/rt685s-evk/src/bin/type_c.rs
+++ b/examples/rt685s-evk/src/bin/type_c.rs
@@ -6,7 +6,7 @@ use defmt::info;
 use embassy_embedded_hal::shared_bus::asynch::i2c::I2cDevice;
 use embassy_executor::Spawner;
 use embassy_imxrt::gpio::{Input, Inverter, Pull};
-use embassy_imxrt::i2c::master::{I2cMaster, Speed};
+use embassy_imxrt::i2c::master::{Config, I2cMaster};
 use embassy_imxrt::i2c::Async;
 use embassy_imxrt::{bind_interrupts, peripherals};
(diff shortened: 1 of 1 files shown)


jerrysxie (dismissed):

examples/rt685s-evk/src/bin/type_c.rs
```diff
@@ -139,8 +139,12 @@ async fn main(spawner: Spawner) {
 
     let int_in = Input::new(p.PIO1_7, Pull::Up, Inverter::Disabled);
     static BUS: OnceLock<Mutex<NoopRawMutex, BusMaster<'static>>> = OnceLock::new();
+    let config = embassy_imxrt::i2c::master::Config {
+        speed: embassy_imxrt::i2c::master::Speed::Standard,
+        duty_cycle: embassy_imxrt::i2c::master::DutyCycle::new(50).unwrap(),
+    };
```
felipebalbi, line 145: isn't this, basically, the default implementation? (apart from duty cycle being 40 by default, probably no difference). Do you mind trying passing `Default::default()` as the config argument?

felipebalbi (changes requested):

asasine (dismissed):

examples/rt685s-evk/src/bin/type_c.rs
tullom: Done.

jerrysxie (approved):

jerrysxie:
Bypass policy and merge since this is a trivial change to fix the workflows.

PR #278, opened 2025-05-05
Fix cargo clippy warnings

### LINKED ISSUE #247: Update empty log macro implementations so they don't result in empty branch clippy warnings
The log macro wrappers provide an empty implementation when neither `log` or `defmt` is enabled. Several checks end up running with no features enabled so these empty implementations can get used. This can result in false clippy warnings with code like
```
if a {
    info!("Message A");
} else {
    info!("Message B");
}
```

Clippy will warning that both blocks here are empty/the same. See if we can change the empty macro implementation to either naturally suppress this warning or explicitly suppress it with a linter hint.

kurtjd:
This PR addresses all `cargo clippy` warnings.

Resolves #247

### DIFF
--- a/embedded-service/src/fmt.rs
+++ b/embedded-service/src/fmt.rs
@@ -11,6 +11,7 @@ mod defmt {
     macro_rules! trace {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
+                let _ = ($s, $( &$x ),*);
                 ::defmt::trace!($s $(, $x)*);
             }
         };
@@ -22,6 +23,7 @@ mod defmt {
     macro_rules! debug {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
+                let _ = ($s, $( &$x ),*);
                 ::defmt::debug!($s $(, $x)*);
             }
         };
@@ -33,6 +35,7 @@ mod defmt {
     macro_rules! info {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
+                let _ = ($s, $( &$x ),*);
                 ::defmt::info!($s $(, $x)*);
             }
         };
@@ -44,6 +47,7 @@ mod defmt {
     macro_rules! warn {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
+                let _ = ($s, $( &$x ),*);
                 ::defmt::warn!($s $(, $x)*);
             }
         };
@@ -55,6 +59,7 @@ mod defmt {
     macro_rules! error {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
+                let _ = ($s, $( &$x ),*);
                 ::defmt::error!($s $(, $x)*);
             }
         };
@@ -128,7 +133,7 @@ mod none {
     macro_rules! trace {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
-                let _ = ($( & $x ),*);
+                let _ = ($s, $( &$x ),*);
             }
         };
     }
@@ -139,7 +144,7 @@ mod none {
     macro_rules! debug {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
-                let _ = ($( & $x ),*);
+                let _ = ($s, $( &$x ),*);
             }
         };
     }
@@ -150,7 +155,7 @@ mod none {
     macro_rules! info {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
-                let _ = ($( & $x ),*);
+                let _ = ($s, $( &$x ),*);
             }
         };
     }
@@ -161,7 +166,7 @@ mod none {
     macro_rules! warn {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
-                let _ = ($( & $x ),*);
+                let _ = ($s, $( &$x ),*);
             }
         };
     }
@@ -172,7 +177,7 @@ mod none {
     macro_rules! error {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
-                let _ = ($( & $x ),*);
+                let _ = ($s, $( &$x ),*);
             }
         };
     }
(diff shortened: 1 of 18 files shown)


RobertZ2011 (approved):

RobertZ2011:
> EXCEPT for await_holding_refcell_ref warnings as more clarification is needed if this is intentional and it is confirmed the risk of runtime panics is mitigated

These are intentional. At least in what I've written there shouldn't be any panics because the borrow happens at a single location and then the mutable reference is passed as an argument to other functions. But I'm open to switching over to synchronization types. I mostly avoided them because this code only runs in its own task and there may have been some situations where an async mutex lock wasn't compatible due to needing to be called in a non-async function.

kurtjd:
> > EXCEPT for await_holding_refcell_ref warnings as more clarification is needed if this is intentional and it is confirmed the risk of runtime panics is mitigated
> 
> These are intentional. At least in what I've written there shouldn't be any panics because the borrow happens at a single location and then the mutable reference is passed as an argument to other functions. But I'm open to switching over to synchronization types. I mostly avoided them because this code only runs in its own task and there may have been some situations where an async mutex lock wasn't compatible due to needing to be called in a non-async function.

Sounds good, makes sense to me! I can just add an explicit clippy `allow` for this to suppress the warnings if everyone is okay with that.

type-c-service/src/task.rs
```diff
@@ -83,6 +83,7 @@ impl Service {
                 connected: status.is_connected(),
             };
 
+            #[allow(clippy::if_same_then_else)]
```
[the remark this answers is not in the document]
RobertZ2011: See this bug for some background: https://github.com/OpenDevicePartnership/embedded-services/issues/247. @kurtjd you might also see if moving this into the empty log macro definitions works to suppress the warning. Then we would be able to close out that bug as well.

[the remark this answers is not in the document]
madeleyneVaca: gotcha. Thanks for the context!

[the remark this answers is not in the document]
kurtjd: Thanks Robert, didn't see that issue. I will see if I can move the `allow` tag further into the macro itself.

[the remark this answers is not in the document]
kurtjd: I was able to resolve this warning in the empty log functions by replacing:

```rust
let _ = ($( & $x ),*);
```

with:

```rust
let _ = ($s, $( &$x ),*);
```

However, when clippy is ran with the ```defmt``` feature, the warning is still generated, but strangely not the case when ran with the `log` feature. I couldn't really find a satisfying reason why, so I simply added the noop:

```rust
let _ = ($s, $( &$x ),*);
```

to the defmt log macros which worked.

kurtjd:
> > EXCEPT for await_holding_refcell_ref warnings as more clarification is needed if this is intentional and it is confirmed the risk of runtime panics is mitigated
> 
> These are intentional. At least in what I've written there shouldn't be any panics because the borrow happens at a single location and then the mutable reference is passed as an argument to other functions. But I'm open to switching over to synchronization types. I mostly avoided them because this code only runs in its own task and there may have been some situations where an async mutex lock wasn't compatible due to needing to be called in a non-async function.

I decided to add the lint suppression to each offending method (as opposed to the crate as a whole) just to keep the scope narrow, but can do the opposite if it feels too cluttered.

kurtjd:
The cargo vet error appears to be a problem on the server side? No dependencies were changed from main nor between force pushes.

jerrysxie (approved):

embedded-service/src/fmt.rs
```diff
@@ -139,7 +144,7 @@ mod none {
     macro_rules! debug {
         ($s:literal $(, $x:expr)* $(,)?) => {
             {
-                let _ = ($( & $x ),*);
+                let _ = ($s, $( &$x ),*);
```
madeleyneVaca, line 147: making sure I understand. This is so that in the "empty" cases, the parameters of the macros are technically "used" by this tuple that's created so they aren't flagged by clippy?

separately and maybe more of a @RobertZ2011 question. I understand the * usage in the macro definition so that we can have strings formatted with multiple args. but I'm not sure I understand it's usage in the tuple on this line

kurtjd: That's my understanding, yeah. And I believe it shouldn't result in any actual generated code when compiled since it's effectively a noop (though if someone passes a function with side-effects as a parameter then hm, maybe code would be generated, so I'll double check).

For the * usage, my understanding is that this basically expands the parameters into comma-separated values within the tuple, thus making it seem like each parameter is being "used" as well, but Robert probably knows more.

madeleyneVaca: I think what's throwing me off is that this is written as `( &$x ),*` where the comma is outside the parentheses for the `$x` reference instead of inside the parentheses like first usage of `*`

madeleyneVaca: [macro_rules! - The Little Book of Rust Macros](https://fprijate.github.io/tlborm/mbe-macro-rules.html#repetitions)
the * repeats both the `(blah)` and the `,` wherein it's treating `,` as the separator. I was thinking in regex

RobertZ2011: I mostly based that on how Embassy handles a similar situation, but my understand is the same as what Kurtis commented: it expands to a comma separated list. Macro syntax is weird in general unfortunately.

kurtjd: To give an example from the output of `cargo expand`:

```rust
debug!("hello world: {}, {}, {}", 1, 2, 3);
```

expands to:

```rust
let _ = ("hello world: {}, {}, {}", &1, &2, &3);
```

when the macro is as follows:

```rust
    /// Logs a debug message using the underlying logger
    #[macro_export]
    #[collapse_debuginfo(yes)]
    macro_rules! debug {
        ($s:literal $(, $x:expr)* $(,)?) => {
            {
                let _ = ($s, $( &$x ),*);
            }
        };
    }
```

PR #279, opened 2025-05-05
Add Platform Reset Service

JamesHuard:
Add an equivalent to NVIC_SystemReset(), that allows registration of pre-emptive blockers that can temporarily hold off the MCU platform reset.

### DIFF
--- a/examples/rt685s-evk/src/bin/reset.rs
+++ b/examples/rt685s-evk/src/bin/reset.rs
@@ -1,0 +1,59 @@
+#![no_std]
+#![no_main]
+
+extern crate rt685s_evk_example;
+
+use platform_service::reset;
+use {defmt_rtt as _, panic_probe as _};
+
+async fn print_watcher_number() {
+    static CONTRIVED: embassy_sync::mutex::Mutex<embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex, usize> =
+        embassy_sync::mutex::Mutex::new(0);
+
+    // yes, this could be accomplished with atomics. But using a mutex here demonstrates async functionality
+    let watcher_num = {
+        let mut current_number = CONTRIVED.lock().await;
+        *current_number += 1;
+        *current_number
+    };
+
+    defmt::info!("Reset Watcher #{}", watcher_num);
+}
+
+#[embassy_executor::task(pool_size = 10)]
+async fn reset_watcher(blocker: &'static reset::Blocker) {
+    defmt::info!("reset::Blocker watch thread ticking...");
+
+    loop {
+        blocker
+            .wait_for_reset(async || {
+                print_watcher_number().await;
+            })
+            .await;
+    }
+}
+
+#[embassy_executor::main]
+async fn main(spawner: embassy_executor::Spawner) {
+    let _p = embassy_imxrt::init(Default::default());
+
+    static BLOCKERS: embassy_sync::lazy_lock::LazyLock<[reset::Blocker; 10]> =
+        embassy_sync::lazy_lock::LazyLock::new(|| [const { reset::Blocker::uninit() }; 10]);
+
+    embedded_services::init().await;
+
+    let blockers = BLOCKERS.get();
+
+    // spawn blocker threads
+    for blocker in blockers {
+        // register before spawning blocker threads to avoid potential scheduling issues
+        // when immediately calling reset below
+        blocker.register().await.expect("Infallible");
+
+        spawner.must_spawn(reset_watcher(blocker));
+    }
+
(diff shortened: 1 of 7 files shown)


platform-service/src/reset.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    // 1. signal all events
+    for blocker in BLOCKERS.get().iter_only::<Blocker>() {
+        blocker.reset_pending.signal(());
+    }
+
+    // 2. wait for all events
+    for blocker in BLOCKERS.get().iter_only::<Blocker>() {
+        blocker.unblocked.wait().await;
+    }
```
asasine, line 62: Is blocking based on registration order intended in the design?

platform-service/src/reset.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    // 3. perform platform reset
+    #[cfg(feature = "cortex-m")]
+    cortex_m::peripheral::SCB::sys_reset();
+
+    // no equivalent reset option for std environment
+    #[cfg(not(feature = "cortex-m"))]
+    panic!("Cannot reset without NVIC");
```
asasine, line 70: I feel like a failure to compile would be more appropriate to avoid the runtime panic entirely.

platform-service/src/reset.rs
```diff
@@ -0,0 +1,71 @@
...
+        self.unblocked.signal(());
+    }
+}
+
+/// Signals and waits for all registered blockers to complete their async operations before performing a platform-specific reset, typically NVIC_RESET
+pub async fn system_reset() -> ! {
+    // signal and wait for completion as two separate events to allow for alternative scheduling algorithms to take effect
+
+    // 1. signal all events
+    for blocker in BLOCKERS.get().iter_only::<Blocker>() {
```
asasine, line 55: Can you `BLOCKERS.get()` once?

asasine (dismissed):

platform-service/src/reset.rs
JamesHuard: No, the order here is arbitrary. Since it iterates over all of them, it shouldn't really matter which order we clear the signals.

platform-service/src/reset.rs
JamesHuard: I wanted to preserve compilation with no feature flags

platform-service/src/reset.rs
JamesHuard: done

platform-service/src/reset.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// allocate a Blocker, such that it could be used in a static
+    pub const fn uninit() -> Self {
+        Self {
+            node: Node::uninit(),
+            reset_pending: Signal::new(),
+            unblocked: Signal::new(),
+        }
+    }
+
+    /// call once on startup to be registered as a Reset handling blocker
```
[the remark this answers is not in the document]
JamesHuard: done

examples/rt685s-evk/src/bin/reset.rs
```diff
@@ -0,0 +1,59 @@
...
+    for blocker in blockers {
+        // register before spawning blocker threads to avoid potential scheduling issues
+        // when immediately calling reset below
+        blocker.register().await.expect("Infallible");
+
+        spawner.must_spawn(reset_watcher(blocker));
+    }
+
+    // perform reset
+    defmt::info!("Performing platform reset!");
```
madeleyneVaca, line 57: nit but would be cool to know how long this takes for profiling. Can you add a timestamp before and after the reset?

madeleyneVaca (dismissed):

examples/rt685s-evk/src/bin/reset.rs
JamesHuard: To do that, I'll have to remove the actual call to reset. I can make that change locally, and then give you a number. This would be the time for 10 defmt::info!()s to be queued across 10 different threads, plus the signaling overhead.

JamesHuard: image
it's beyond the precision of the defmt timestamps

platform-service/src/reset.rs
asasine: In that case, you could use a join

asasine (approved):

asasine (approved):

examples/rt685s-evk/src/bin/reset.rs
madeleyneVaca: fabulous

madeleyneVaca (approved):

PR #280, opened 2025-05-05
Add general FW update trait

RobertZ2011:
Add a general FW update trait. The goal with this PR is to provide a trait that's general enough for any driver to implement and general enough to implement any FW update protocol (like CFU) on top of. The successful flow with this trait is: `start_fw_update` -> `write_fw_contents` (can be called repeatedly to write the update in small fragments) -> `finalize_fw_update`. `abort_fw_update` would be called if there's an error at any point along the normal update path.


embedded-service/src/fw_update.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn start_fw_update(&mut self) -> impl Future<Output = Result<(), Self::Error>>;
+
+    /// Terminate the firmware update after a failure
+    fn abort_fw_update(&mut self) -> impl Future<Output = Result<(), Self::Error>>;
+
+    /// Finalize the firmware update after all contents have been sent
+    fn finalize_fw_update(&mut self) -> impl Future<Output = Result<(), Self::Error>>;
+
+    /// Supply firmware update contents
+    fn write_fw_contents(&mut self, data: &[u8]) -> impl Future<Output = Result<(), Self::Error>>;
```
JamesHuard, line 52: with this API, it's up to the impl'er to keep track of the "current" write operation. Is this desirable? Would we want to provide some kind of addressing parameter alongside data?

embedded-service/src/fw_update.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+/// General firmware update trait
+pub trait FwUpdate {
+    /// Error type
+    type Error: Error;
+
+    /// Get the current firmware version
+    fn get_active_fw_version(&self) -> impl Future<Output = Result<u32, Self::Error>>;
```
asasine, line 40: Should firmware version be an associated type or generic?

embedded-service/src/fw_update.rs
asasine: Does using an address paremeter require implementers handle non-contiguous writes?

embedded-service/src/fw_update.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// The device is not in FW update mode
+    InvalidMode,
+    /// FW update Content was invalid
+    InvaildContent,
+    /// Full FW update contents were not received
+    Incomplete,
+    /// The operation did not complete in the expected time
+    Timeout,
+    /// Error communicating with the device due to a bus error
+    Bus,
```
asasine, line 17: Do we want this to wrap the bus error?

embedded-service/src/fw_update.rs
JamesHuard: From CFU perspective we do need to be able to provide an address/offset. If we need to relax the write command such that it doesn't allow "jumps" in writing, then we might consider adding some kind of base offset/addressing schema to the prepare method

madeleyneVaca: CFU payload has content formatted in {address, len, data} blocks wherein the data for a block is a contiguous chunk. In theory, valid payloads could be generated, following this format, which have non-contiguous blocks. In practice, the payload blocks are collectively contiguous. Our Macronix driver will write an entire page at a time which is a combination of multiple blocks so if the blocks weren't contiguous, we'd have issues.

embedded-service/src/fw_update.rs
RobertZ2011: I think this is the right level to use a concrete type. This simplifies things as any firmware update protocols built on top of this trait won't need to be generic over a device-specific version type.

embedded-service/src/fw_update.rs
RobertZ2011: I added a general bus error wrapper type for convenience. The reason it's not in the `ErrorKind` enum is to help simplify things a bit. This way a concrete bus error type doesn't need to be explicitly threaded through. This also allows supporting devices that don't use a bus (MMIO device maybe?) without needing them to supply a mock bus type like `()`.

JamesHuard (approved):

embedded-service/src/fw_update.rs
```diff
@@ -0,0 +1,72 @@
...
+use core::future::Future;
+
+/// FW update error kind
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ErrorKind {
+    /// The device is not in FW update mode
+    InvalidMode,
+    /// FW update Content was invalid
+    InvaildContent,
```
felipebalbi, line 11: typo: `InvalidContent`

embedded-service/src/fw_update.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// Full FW update contents were not received
+    Incomplete,
+    /// The operation did not complete in the expected time
+    Timeout,
+    /// Error communicating with the device due to a bus error
+    Bus,
+    /// Generic failure
+    Failed,
```
felipebalbi, line 19: nit: how about calling `Other`?

embedded-service/src/fw_update.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn kind(&self) -> ErrorKind;
+}
+
+impl Error for ErrorKind {
+    fn kind(&self) -> ErrorKind {
+        *self
+    }
+}
+
+/// Convienent error type for devices that are on a bus
```
felipebalbi, line 34: typo: Convenient

embedded-service/src/fw_update.rs
```diff
@@ -0,0 +1,72 @@
...
+/// General firmware update trait
+pub trait FwUpdate {
+    /// Error type
+    type Error: Error;
+
+    /// Get the current firmware version
+    fn get_active_fw_version(&self) -> impl Future<Output = Result<u32, Self::Error>>;
+
+    /// Begin a new firmware update
+    fn start_fw_update(&mut self) -> impl Future<Output = Result<(), Self::Error>>;
```
felipebalbi, line 62: From a quick glance, this looks incomplete. If the user implements A/B updates, how can you tell, from this trait, if we're updating A or B? How can you prevent other accesses to the underlying storage medium while update is happening? At a minimum, accesses to "the other" half should be prevented. That is, the "half" being updated should be, somehow, locked for exclusive access.

Additionally, there's nothing preventing the update from overwriting the location of the (optional?) bootloader. There's no notion of partitions, no checks that we're not writing out of bounds, and so on.

Finally, I think this is a somewhat solved problem by `embassy-boot`. Note how they define a [`Partition` type ](https://github.com/embassy-rs/embassy/blob/main/embassy-embedded-hal/src/flash/partition/asynch.rs#L15-L19). Accessing the partition requires locking a mutex. They're also defining the [bootloader state as a partition](https://github.com/embassy-rs/embassy/blob/main/embassy-boot/src/firmware_updater/asynch.rs#L24-L50), so the updater knows to modify the bootloader state (that is, should I boot A or B?) at the end of the update process.

embedded-service/src/fw_update.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    /// Terminate the firmware update after a failure
+    fn abort_fw_update(&mut self) -> impl Future<Output = Result<(), Self::Error>>;
```
felipebalbi, line 65: Can you explain why this is needed? From what I can tell, this will be a situation where this function *must* be called any time `start_fw_update()` fails. If that's the case, why have this method at all? Could just require the implementer to handle the errors and place the flash back in a working state as if the update had not been started.

embedded-service/src/fw_update.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    /// Finalize the firmware update after all contents have been sent
+    fn finalize_fw_update(&mut self) -> impl Future<Output = Result<(), Self::Error>>;
```
felipebalbi, line 68: I think you may need some more states here. I guess you would need to mark that update has started for a given partition (encoded in start_fw_update(), maybe?) then you should mark that update is done (this method?), but you may also need to mark that a partition has booted successfully. This means you also have a soft-dependency on a working watchdog. That is, if you try to boot a partition post update and it resets due to watchdog, you need to rollback to the non-updated partition and mark the recently updated one as "failed to boot".

felipebalbi (changes requested):

embedded-service/src/fw_update.rs
madeleyneVaca: To go with the CFU example, multiple bank quantities (2-4) are supported and it's perfectly valid to only give an offer for a single bank. Obviously if I'm only ever supplying a CFU Host with offers for the same bank for a given component and none of the others, that would be an issue. Still, it's not a hard requirement to give an offer to ALL banks. CFU doesn't have to know how many banks exist nor does it have to know, beyond the header information included in the initial offer packet, which bank it's for since the offsets in each data packet keep track of where in memory to write.

A CFU Host takes whatever offers it's given by the user and send them all down to the EC that has implemented the traits to receive them. Even within those offers, some OEMs might have additional metadata built into their offers which doesn't necessarily go to a specific bank. It's up to the implementation to figure out what to do with the content and if it's trying to go to a valid location.

I agree that bounding checks to make sure we're not overwriting anything outside of the bank that's being updated is helpful but given how much CFU leaves up to the implementation, it'd be tricky to include it in this high level trait

embedded-service/src/fw_update.rs
asasine: But propragating the inner error is the most reliable mechanism for actually debugging and handling inner errors. With the unit variant `Bus`, there's no way to know what the underlying cause was and therefore no way to fix it without guessing.

embedded-service/src/fw_update.rs
RobertZ2011: This trait is targeting a higher abstraction layer and the primary focus is getting the update contents to a specific device. There's no notion of banks, partitions, or a bootloader here because those are device-specific details that the implementation is responsible for handling. I could rename the trait to something like `UpdateWriter` to help clarify things.

PR #285, opened 2025-05-06
Create simple RPC channel

RobertZ2011:
Our existing RPC message channels are implemented from scratch for each device type. These implementations generally look like:
```rust
struct Device {
    command: Channel<_, Message, 1>,
    response: Channel<_, Response, 1>,
}

impl Device {
    pub async fn execute_command(&self, command: Command) -> Response {
        self.command.send(command).await;
        self.response.recv().await
    }
}
```

This implementation is susceptible to a timing bug with multiple tasks. Imagine two tasks: A and B. Task A calls `execute_command(Command::A)` while task B calls `execute_command(Command:B)`. Imagine the following timing:

* Task A puts its command in the `command` channel, task B blocks as the channel is now full
* The RPC is executed and the response in put in the `response` channel
* Task B is unblocked and puts its command in the channel, but can immediately return because there's already a response ready
* Task B reads the response from A's command and task A will end up with the response to B's command

This PR provides a common implementation which includes locking to prevent the above from happening. This implementation also provides wrapper structs that more clearly define the separation between the invoker of an RPC and who actually handles the RPC execution. Once this PR goes in, I'll submit a follow-up PR for the power policy and type-C code and let the owners of other service implements know to migrate over to this struct.


embedded-service/src/ipc/rpc.rs
```diff
@@ -0,0 +1,83 @@
+//! Definitions for IPC
+use core::sync::atomic::AtomicBool;
+
+use embassy_sync::{blocking_mutex::raw::RawMutex, channel::Channel, mutex::Mutex};
+
+/// A simple RPC channel for executing commands
```
asasine, line 6: bikeshedding: "RPC" makes me think it's going off-device when this is really just IPC. Another name could be simply "message handler"

embedded-service/src/ipc/rpc.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Channel for receiving responses
+    response: Channel<M, R, 1>,
+    /// Mutex for synchronizing access to command execution
+    execution_lock: Mutex<M, ()>,
+    handler_created: AtomicBool,
+}
+
+impl<M: RawMutex, C, R> RpcChannel<M, C, R> {
+    /// Create a new RPC channel
+    pub fn new() -> Self {
```
JamesHuard, line 19: Can you mark this const?

embedded-service/src/ipc/rpc.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+impl<'a, M: RawMutex, C, R> RpcHandler<'a, M, C, R> {
+    /// Create a new handler object
+    fn new(rpc: &'a RpcChannel<M, C, R>) -> Self {
+        Self { inner: rpc }
+    }
+
+    /// Wait for a command
+    pub async fn wait_command(&self) -> C {
```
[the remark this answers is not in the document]
JamesHuard: On line 39 here: https://github.com/OpenDevicePartnership/embedded-services/blob/main/platform-service/src/reset.rs is what I'm talking about:
image

examples/std/src/bin/rpc_basic.rs
```diff
@@ -0,0 +1,109 @@
...
+        handler.send_response(response).await;
+    }
+}
+
+/// Main task that initializes the device and spawns other tasks
+#[embassy_executor::task]
+async fn task(spawner: Spawner) {
+    static DEVICE: OnceLock<Device> = OnceLock::new();
+
+    let device = DEVICE.get_or_init(Device::new);
```
JamesHuard, line 94: I'm getting a little turned around with what we're calling a "device" here versus the tasks that process the invoked requests. Could we maybe use a more directionally-explicit language, like "Broker/Operator" or "Delegator/Delegate" etc., just to be clear around intended use?

asasine (approved):

embedded-service/src/ipc/rpc.rs
RobertZ2011: Done.

embedded-service/src/ipc/rpc.rs
```diff
@@ -0,0 +1,83 @@
+//! Definitions for IPC
+use core::sync::atomic::AtomicBool;
+
+use embassy_sync::{blocking_mutex::raw::RawMutex, channel::Channel, mutex::Mutex};
+
+/// A simple RPC channel for executing commands
+pub struct RpcChannel<M: RawMutex, C, R> {
```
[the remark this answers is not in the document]
RobertZ2011: Renamed to deferred.

embedded-service/src/ipc/rpc.rs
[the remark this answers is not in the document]
RobertZ2011: Done. I added in a similar function but left the existing ones as well. Access to those will be needed for composing things together but that's something we should be able to mostly hide with something like the transformers work.

embedded-service/src/ipc/rpc.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Only one handler can be created
+    pub fn handler(&self) -> Option<RpcHandler<M, C, R>> {
+        if self.handler_created.swap(true, core::sync::atomic::Ordering::SeqCst) {
+            return None;
+        }
+        Some(RpcHandler::new(self))
+    }
+}
+
+/// Invoker object for invoking commands
```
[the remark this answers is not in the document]
RobertZ2011: I'm open to changing it, but I think that client has the same issue as RPC in that it makes me think of a networked architecture.

embedded-service/src/ipc/rpc.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    /// Create an invoker object that can be used to invoke commands
+    pub fn invoker(&self) -> RpcInvoker<M, C, R> {
+        RpcInvoker::new(self)
+    }
+
+    /// Create a handler object that can be used to execute commands generated by an invoker
+    /// Only one handler can be created
```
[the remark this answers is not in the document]
RobertZ2011: Done. The original motivation was to more closely match existing code that focuses on only having a single task that processes command. But I think it's a better idea to make this more generic and users of this can decide how they want to control that.

examples/std/src/bin/rpc_basic.rs
RobertZ2011: Done. I renamed things to follow the handler/invoke nomenclature present in the library code. Hopefully that clears things up a bit.

embedded-service/src/ipc/deferred.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub async fn send_response(&self, response: R) {
+        self.inner.response.send(response).await;
+    }
+
+    /// Process a command and send a response
+    pub async fn process_command<Fut: Future<Output = R>, F: FnOnce(C) -> Fut>(&self, process: F) {
+        let command = self.wait_command().await;
+        let response = process(command).await;
+        self.send_response(response).await;
+    }
```
asasine, line 83: This isn't used, and I think the overall design may be too opinionated. IMO best to remove it for now

asasine (dismissed):

embedded-service/src/ipc/deferred.rs
JamesHuard: Do we really not want to encourage a response for each command?

JamesHuard (approved):

embedded-service/src/ipc/deferred.rs
felipebalbi: > This isn't used, and I think the overall design may be too opinionated. IMO best to remove it for now

I second what @asasine is stating here, sorta. Maybe renaming `process_command()` would suffice, as it's not only processing, but already receiving the command and sending a response. I can see how it can become too opinionated at times.

embedded-service/src/ipc/rpc.rs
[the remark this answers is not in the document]
RobertZ2011: I'm going to hold off on implementing this for now. I created an initial implementation but ran into issues surrounding exactly how the mutex guards should be handled. I want to play around more with various situations to get a better handle on that first.

embedded-service/src/ipc/deferred.rs
RobertZ2011: I removed this. I really like the benefits to correctness, but this pattern tends to encourage writing code in a way that makes things less modular and composable. E.g. if an implementation uses this then it's hard to gain access to the actual commands to allow for extending functionality.

RobertZ2011:
Latest commits change things a bit. The previous implementation had a bug around timeouts/futures being dropped in general. If a command was submitted but the response never read then that stale response would become the response for the subsequent command. Now each invocation of a command has a unique ID that can be used to ensure that the correct response is for us. I did also consider passing in a `&Signal`, but that would have to be static so we couldn't have a self-contained invoke function like this implementation provides. I also moved over to using Embassy signals instead of channels since we already have synchronization with the mutex. I removed the `Handler` and `Invoker` wrapper because each one would only need a single function anyway.

embedded-service/src/ipc/deferred.rs
```diff
@@ -0,0 +1,98 @@
...
+};
+
+use embassy_sync::{blocking_mutex::raw::RawMutex, mutex::Mutex, signal::Signal};
+
+use crate::debug;
+
+/// A unique identifier for a particular command invocation
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct InvocationId(u32);
```
asasine, line 14: Doesn't need to be public
```suggestion
struct InvocationId(u32);
```

embedded-service/src/ipc/deferred.rs
```diff
@@ -0,0 +1,98 @@
...
+/// using a unique invocation ID.
+pub struct Channel<M: RawMutex, C, R> {
+    /// Signal for sending commands
+    command: Signal<M, (C, InvocationId)>,
+    /// Signal for receiving responses
+    response: Signal<M, (R, InvocationId)>,
+    /// Mutex for synchronizing access to command invocation
+    invocation_lock: Mutex<M, ()>,
+    /// Unique ID for the next invocation
+    next_invocation_id: AtomicU32,
```
asasine, line 28: Using u32 seems aritrary so I arbitrarily recommend `AtomicUsize`

embedded-service/src/ipc/deferred.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// The channel this invocation came from
+    channel: &'a Channel<M, C, R>,
+    /// Invocation ID
+    invocation_id: InvocationId,
+    /// Command to execute
+    pub command: C,
+}
+
+impl<M: RawMutex, C, R> Invocation<'_, M, C, R> {
+    /// Send a response to the command
```
asasine, line 94: There's a nice design decision here that I think is worth highlighting so it remains long-term and doesn't trip up users
```suggestion
    /// Send a response to the command, consuming the command in the process.
    ///
    /// Consuming the command ensures each command may only be responded to once.
```

embedded-service/src/ipc/deferred.rs
```diff
@@ -0,0 +1,98 @@
...
+use embassy_sync::{blocking_mutex::raw::RawMutex, mutex::Mutex, signal::Signal};
+
+use crate::debug;
+
+/// A unique identifier for a particular command invocation
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct InvocationId(u32);
+
+/// A simple channel for executing deferred commands
```
asasine, line 16: nit: a blank line ensures the top-level decription is brief and sufficiently informative.
```suggestion
/// A simple channel for executing deferred commands.
///
```

embedded-service/src/ipc/deferred.rs
```diff
@@ -0,0 +1,98 @@
...
+            next_invocation_id: AtomicU32::new(0),
+        }
+    }
+
+    /// Invoke a command and return the response
+    /// This locks to ensure that commands are executed atomically
+    pub async fn invoke(&self, command: C) -> R {
+        let _guard = self.invocation_lock.lock().await;
+        let request_id = self.next_invocation_id.fetch_add(1, Ordering::SeqCst);
+        let request_id = InvocationId(request_id);
```
asasine, line 50: Put behind a private method so there's less likelihood that the primitive version of the value is misused. Even better, create a (unit tested) wrapper around an incrementing atomic that never exposes the primitive and use that instead.

asasine (approved):
Looks good to me.

Bikeshedding: I don't like the invoker/invocation name; it's too wordy imo and requires too much new mental chewing to grasp since it doesn't resemble existing synchronization primitives. What about sticking closer to existing nomenclature for channels or signals, which this is behaving like?
- `invoke()` -> `send()`
- `Invocation` -> `Message` (most conventional) or `Request` or `Command`
- `wait_invocation()` -> `receive()` or `wait()` or `get()`
- `send_response()` -> `respond()` or `send()`

I also think this is completely unit testable and would like to see them added.

embedded-service/src/ipc/deferred.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Signal for sending commands
+    command: Signal<M, (C, InvocationId)>,
+    /// Signal for receiving responses
+    response: Signal<M, (R, InvocationId)>,
+    /// Mutex for synchronizing access to command invocation
+    invocation_lock: Mutex<M, ()>,
+    /// Unique ID for the next invocation
+    next_invocation_id: AtomicU32,
+    /// Phantom data for the response type
+    _phantom: PhantomData<R>,
```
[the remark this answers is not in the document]
RobertZ2011: Done.

embedded-service/src/ipc/deferred.rs
RobertZ2011: Done.

embedded-service/src/ipc/deferred.rs
RobertZ2011: Done.

embedded-service/src/ipc/deferred.rs
RobertZ2011: Done.

embedded-service/src/ipc/deferred.rs
RobertZ2011: Done.

embedded-service/src/ipc/deferred.rs
RobertZ2011: Done, went with just a private method and added a test.

RobertZ2011:
> Looks good to me.
> 
> Bikeshedding: I don't like the invoker/invocation name; it's too wordy imo and requires too much new mental chewing to grasp since it doesn't resemble existing synchronization primitives. What about sticking closer to existing nomenclature for channels or signals, which this is behaving like?
> 
> * `invoke()` -> `send()`
> * `Invocation` -> `Message` (most conventional) or `Request` or `Command`
> * `wait_invocation()` -> `receive()` or `wait()` or `get()`
> * `send_response()` -> `respond()` or `send()`
> 
> I also think this is completely unit testable and would like to see them added.

Ended up going with `execute` because I think it captures the back and forth nature better. I also added logic to timeout by default since it's safer. Changed the other invocation type names to some form of either request or receive.

I added some tests, primarily by converting the example over into a test. I'm using tokio as the test executor because the embassy executor will never stop running so it's not a good match for testing. Embassy itself seems to use `futures_executor` but that doesn't seem to provide a straightforward way to spawn arbitrary tasks. Anything embassy time related can't execute on the tokio executor though so this PR only directly tests the non-timeout versions.

asasine (changes requested):

asasine (dismissed):

RobertZ2011:
Going to land the tests as a separate PR: https://github.com/OpenDevicePartnership/embedded-services/pull/297. There's some weirdness with using embassy in tests within a workspace but I want to land this PR since it should resolve several bugs.

asasine (approved):

JamesHuard (approved):

Issue #262, opened 2025-04-23
Partition manager


Wassasin:
# Partition manager

A storage medium has to be partitioned.
This needs to be centrally managed and shared over multiple firmware projects (like application and bootloader).

A partition can be:
- Read only
- Read write

Usage of a partition might require additional properties:
- XIP-able
- NOR interface
- NAND interface

(A NOR interface can always emulate a NAND interface and vice versa if some inefficiencies are allowed)

In Rust we typically build things layer by layer, each layer wrapping the previous.
This would make sense to do for the partitions as well. It also separates the partition from what it's used for.

## Proposal

The partition manager needs to be able to keep hold of the flash an synchronize access to it.
We can also allow higher level layers to get a partition with a particular interface.
If that interface is different to the native interface of the flash we can put an emulation layer inbetween.

```rust
// Pseudo code

pub struct PartitionManager<F> {
    flash: Mutex<F>,
    xip_address: Option<u32>,
}

impl<F> PartitionManager<F> {
    pub fn new(flash: F, xip_address: Option<u32>) -> Self {
        Self { flash: Mutex::new(flash), xip_address }
    }
}

impl<F: NorFlash> PartitionManager<F> {
    pub fn get_native_partition(&self, partition_config: PartitionConfig) -> NorPartition<F> {
        // ...
    }

    pub fn get_nor_partition(&self, partition_config: PartitionConfig) -> NorPartition<F> {
        self.get_native_partition(partition_config)
    }

    pub fn get_nand_partition(&self, partition_config: PartitionConfig) -> NandPartition<NandEmulator<F>> {
        // ...
    }
}

impl<F: NandFlash> PartitionManager<F> {
    pub fn get_native_partition(&self, partition_config: PartitionConfig) -> NandPartition<F> {
        // ...
    }


    pub fn get_nor_partition(&self, partition_config: PartitionConfig) -> NorPartition<NorEmulator<F>> {
        // ...
    }

    pub fn get_nand_partition(&self, partition_config: PartitionConfig) -> NandPartition<F> {
        self.get_native_partition(partition_config)
    }
}
```

A big question is how XIP is done. In general that's difficult to abstract over.
On STM32 for example you can either have XIP active or have the ability to do writes and erases.
You can't have both at the same time.

On top of these partitions you can put another layer that abstracts over what the purpose of the partition is.

For example for the logger (but same deal with config):
```rust
let manager = PartitionManager::new(flash, None);

[...]

PR #286, opened 2025-05-06
Add partition manager

Wassasin:
Implements #262.

Goal of this PR is to have a crate that enables a storage device to be separated into partitions statically and checked. It achieves this by providing a general partition framework, implement typical traits from `embedded-storage-async` and `block-device-driver`, and have proc macros generate partition map definitions.


partition-manager/generation/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+
+                    let access = match access {
+                        Some(Access::RO) => Access::RO,
+                        Some(Access::RW) => Access::RW,
+                        None if partition.access.is_empty() => Access::RW, // Nothing specified, assume RW for all.
+                        None => return None,                               // No-match, do not emit partition.
+                    };
+
+                    Some(GeneratedPartition {
+                        name: partition_name.clone(),
```
tdittr, line 202: This `clone` might be redundant since you used `into_iter`

partition-manager/generation/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+                }
+                None => None,
+            };
+
+            let variant = variant_name.map(Variant::Other).unwrap_or(Variant::Any);
+
+            Ok(self
+                .partitions
+                .into_iter()
+                .filter_map(move |(partition_name, partition)| {
```
tdittr, line 188: Nit-Pick: You could destructure `partition` here, making copying `access`, `name`, and `size` a tiny-bit less verbose.

But feel free to leave it as is.

partition-manager/generation/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+        }
+    }
+
+    #[derive(Debug, Clone, PartialEq, Deserialize)]
+    #[doc(hidden)]
+    pub struct Manifest {
+        #[serde(default)]
+        pub variants: BTreeSet<Variant>,
+        pub disk: Disk,
+        pub partitions: BTreeMap<String, Partition>,
```
tdittr, line 114: I think this will silently drop `Partition`s when there are redundant partition names.

partition-manager/generation/tests/access.rs
```diff
@@ -0,0 +1,244 @@
...
+fn variant_no_match() {
+    let mut manifest = create_manifest();
+    manifest.partitions.get_mut("l1").unwrap().access = BTreeMap::from_iter([("bootloader".into(), Access::RW)]);
+
+    let result = manifest
+        .generate(Some("application".to_string()))
+        .map(Vec::from_iter)
+        .unwrap();
+
+    assert_eq!(result, Vec::new());
```
tdittr, line 61: I really like the use of `assert_eq!` over `assert!(result.is_empty())` to get better output on error :+1:

tdittr (commented):
Reviewed the generation crate, looks nice so far!

Only thing I was supprised by are the `#[doc(hidden)]` functions, but I guess they are needed by the macro?

Can finish the rest of the review on Friday.

partition-manager/generation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "partition-manager-generation"
+version.workspace = true
```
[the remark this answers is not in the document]
Wassasin: That is up to ODP.

partition-manager/generation/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+
+        pub fn generate(
+            self,
+            variant_name: Option<String>,
+        ) -> anyhow::Result<impl Iterator<Item = GeneratedPartition>> {
+            self.check_consistency()?;
+
+            let variant_name = match variant_name {
+                Some(variant_name) => {
+                    let variant_name = variant_name.to_string();
```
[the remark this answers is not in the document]
Wassasin: Yep! Will fix it.

partition-manager/generation/src/lib.rs
Wassasin: Applied!

partition-manager/generation/src/lib.rs
Wassasin: Hmm, I am not 100% convinced this is something we should handle, but I added a custom deserializer to handle this just in case.

Wassasin:
> Reviewed the generation crate, looks nice so far!
> 
> Only thing I was supprised by are the `#[doc(hidden)]` functions, but I guess they are needed by the macro?
> 
> Can finish the rest of the review on Friday.
Unfortunately the tests for generation are considered to be in a separate crate context.

partition-manager/generation/src/lib.rs
```diff
@@ -0,0 +1,381 @@
+#![no_std]
+
+#[cfg(not(target_os = "none"))]
+pub use internal::*;
```
asasine, line 4: Question for my learning: if this crate exports nothing unless it's targetting an OS, why is it no_std? It's only used by the proc-macro crate and those don't need to bu no_std to be used by no_std crates.

asasine (commented):
Regarding json vs. yaml vs. toml, I'm of the mind that they're effectively interchangeable and if toml is already vetted and there's no need for json and yaml, we should only ship toml.

partition-manager/partition-manager/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// Partition of a disk.
+///
+/// If the underlying disk implements [embedded_storage_async::nor_flash::NorFlash] or [block_device_driver::BlockDevice], this partition will too.
+/// Requires the features `esa` and/or `bdd` to be enabled for this crate.
+#[allow(unused)]
+pub struct Partition<'a, F, MARKER, M: RawMutex = NoopRawMutex> {
+    storage: &'a Mutex<M, F>,
+    offset: u32,
+    size: u32,
```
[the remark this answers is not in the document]
Wassasin: I do not really have an opinion on this, for now I went with what is used in `embedded-storage-async`, which is [u32](https://docs.rs/embedded-storage-async/0.4.1/embedded_storage_async/nor_flash/trait.NorFlash.html#tymethod.write).

partition-manager/generation/src/lib.rs
Wassasin: This is because the current CI setup for this repository assumes all crates to be `no_std` compatible by checking `no_std` targets. Here I had to make a choice to either change the CI such that it would leave out this library crate, or change the library crate to work for `no_std` as well.

This is a long standing discussion with the cargo team and the embedded WG to get big workspaces like `embedded-services` to support very distinct targets in a workable manner, but unfortunately we are not there yet.

partition-manager/generation/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+        Other(String),
+    }
+
+    impl<'de> serde::Deserialize<'de> for Variant {
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: serde::Deserializer<'de>,
+        {
+            let s = String::deserialize(deserializer)?;
+            Ok(if s == "any" { Variant::Any } else { Variant::Other(s) })
```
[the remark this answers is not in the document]
Wassasin: There are other serialization crates with derive that use `flatten` for this, but Serde does not because it would be a bit unclear what the exact intent is there.

partition-manager/generation/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+        pub offset: u32,
+        pub size: u32,
+    }
+
+    #[allow(clippy::from_over_into)]
+    impl Into<Range<u32>> for &Partition {
+        fn into(self) -> Range<u32> {
+            self.offset..(self.offset + self.size)
+        }
+    }
```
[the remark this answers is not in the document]
Wassasin: That was my initial assumption, but the orphan rule does not apply here: you can implement the trait for a foreign struct if you defined the trait. For traits with associated types for which you defined the types, which is the case for `From<&Partition>`, it is allowed. I fixed it, thanks!

partition-manager/generation/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+
+        let partitions_def = partitions.iter().map(|partition| {
+            let partition_name = quote::format_ident!("{}", partition.name);
+
+            let access = match partition.access {
+                Access::RO => quote! { partition_manager::RO },
+                Access::RW => quote! { partition_manager::RW },
+            };
+
+            Some(quote! { pub #partition_name: partition_manager::Partition<'a, F, #access, M>, })
```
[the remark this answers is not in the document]
Wassasin: In an earlier version this was a `filter_map`, fixed it!

partition-manager/generation/tests/access.rs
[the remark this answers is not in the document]
Wassasin: Hmm, yeah in this case that would be better, but only slightly because of the `no_std` conditional. Changed it, and now some definitions can be `pub(crate)` instead of `pub` with `#[doc(hidden)]`.

asasine (approved):
Looks good to me but I wasn't around during requirements gathering so take my approval with a grain of salt.

felipebalbi (approved):

partition-manager/partition-manager/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+        config.map(&self.storage)
+    }
+}
+
+impl<F, MARKER, M: RawMutex> Partition<'_, F, MARKER, M> {
+    /// Checks whether an address range lies within the partition.
+    #[allow(unused)]
+    const fn within_bounds(&self, offset: u32, size: u32) -> bool {
+        offset + size <= self.size
+    }
```
tdittr, line 93: Should this maybe be a `checked_add`?

partition-manager/partition-manager/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Returns the block number on the parent storage medium, given a block size.
+    ///
+    /// Will not be able to return a value of the partition is not aligned to a single block.
+    #[allow(unused)]
```
tdittr, line 98: Nit-Pick: Maybe this could be moved into an `impl` block in `mod bdd`? Since it is only ever used there?

partition-manager/partition-manager/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+        Partition {
+            storage: self.storage,
+            offset: self.offset,
+            size: self.size,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Temporarily convert a writable partition into a read-only partition.
+    pub const fn readonly_mut(&mut self) -> Partition<'_, F, RO, M> {
```
tdittr, line 46: I agree with @asasine here, since you do not return an `&mut Parition` but an owned `Partition` you do not need the extra `readonly_mut`. All device accesses are protected through the `Mutex` anyway so only need a non-`mut` reference.

tdittr (approved):
I like this design, only found another few nitpicks but no blockers.

partition-manager/partition-manager/src/lib.rs
Wassasin: A yes, I see what you mean with an owned Partition. Do note that the lifetimes are bound. Then I would prefer only having a &mut self variant to limit the existence of an usable partition to one instance. The &self variant is only used at `esa` `capacity`, but I can live with the code being duplicated there.

partition-manager/partition-manager/src/ext/esa.rs
```diff
@@ -0,0 +1,73 @@
...
+        Ok(storage.erase(from + self.offset, to + self.offset).await?)
+    }
+
+    async fn write(&mut self, offset: u32, bytes: &[u8]) -> Result<(), Self::Error> {
+        if !self.within_bounds(offset, bytes.len() as u32) {
+            return Err(Error::OutOfBounds);
+        }
+
+        let mut storage = self.storage.lock().await;
+        Ok(storage.write(offset + self.offset, bytes).await?)
```
Shaibal-Microsoft, line 71: Why are we returning OK() always. Shouldn't we return the actual status?

partition-manager/partition-manager/src/ext/esa.rs
```diff
@@ -0,0 +1,73 @@
...
+    const WRITE_SIZE: usize = F::WRITE_SIZE;
+    const ERASE_SIZE: usize = F::ERASE_SIZE;
+
+    async fn erase(&mut self, from: u32, to: u32) -> Result<(), Self::Error> {
+        if !self.within_bounds(from, to.saturating_sub(from)) {
+            return Err(Error::OutOfBounds);
+        }
+
+        let mut storage = self.storage.lock().await;
+        Ok(storage.erase(from + self.offset, to + self.offset).await?)
```
Shaibal-Microsoft, line 62: Same here. We are returning Ok() always irrespective of write result

partition-manager/partition-manager/src/ext/esa.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    fn capacity(&self) -> usize {
+        self.size as usize
+    }
+}
+
+impl<F: ReadNorFlash, M: RawMutex> ReadNorFlash for Partition<'_, F, RW, M> {
+    const READ_SIZE: usize = F::READ_SIZE;
+
+    async fn read(&mut self, offset: u32, bytes: &mut [u8]) -> Result<(), Self::Error> {
```
Shaibal-Microsoft, line 43: I was not sure where to put this comment. So putting here.

I see that we are only supporting Async mode here. Macronix/Winbond driver and ODP HAL drivers support both Blocking and Async mode.

With only Async mode support and calling blocking API internally will be unexpected behavior from user point of view. Since we are pushing the code to ODP, some user may have use for Blocking mode.

partition-manager/partition-manager/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+mod ext;
+
+#[cfg(test)]
+mod test;
+
+/// Manager for the partitions for a storage device.
+///
+/// Manages concurrent device access and ties lifetime to partitions.
+pub struct PartitionManager<F, M: RawMutex = NoopRawMutex> {
```
Shaibal-Microsoft, line 18: Naive question. I dont see Send and Sync implemented for PartitionManager.
How does it work for multithreaded Executor running on different cores and tasks in both CPUs want to access the filesystem?

de-vri-es: `Send` and `Sync` are auto-traits. They are automatically implemented by the compiler if all the fields implement them :)

From docs.rs:
> ```
> pub unsafe auto trait Send { }
> ```
>
> Types that can be transferred across thread boundaries.
>
> This trait is automatically implemented when the compiler determines it’s appropriate.

partition-manager/partition-manager/src/ext/esa.rs
```diff
@@ -0,0 +1,73 @@
...
+
+impl<F: ReadNorFlash, M: RawMutex> ReadNorFlash for Partition<'_, F, RO, M> {
+    const READ_SIZE: usize = F::READ_SIZE;
+
+    async fn read(&mut self, offset: u32, bytes: &mut [u8]) -> Result<(), Self::Error> {
+        if !self.within_bounds(offset, bytes.len() as u32) {
+            return Err(Error::OutOfBounds);
+        }
+
+        let mut storage = self.storage.lock().await;
```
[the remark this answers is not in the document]
Wassasin: Currently the partition manager is generic over any device. For example we could be dealing with internal NOR, or external SPI NOR. In the latter case exclusive access to the SPI Bus must be guaranteed. This is captured in the the `embedded-storage-async` traits in that you require a mutable reference `&mut self`.

Specifically for the internal NOR case we can consider allowing simultaneous reads using something like a ReadWriteLock. Practically there is currently no ecosystem trait we can call for the underlying device that enable non-mutable `&self` reads.

If we need to specifically improve read performance when using internal NOR or if we have a RAM backed cached NOR, we could consider adding such a trait.

partition-manager/partition-manager/src/ext/esa.rs
```diff
@@ -0,0 +1,73 @@
+//! Embedded Storage Async
+
+use crate::{Error, Partition, RO, RW};
+use core::fmt::Debug;
+use embassy_sync::blocking_mutex::raw::RawMutex;
+use embedded_storage_async::nor_flash::{ErrorType, NorFlash, NorFlashError, NorFlashErrorKind, ReadNorFlash};
```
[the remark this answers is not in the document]
Wassasin: The README discusses this; see the subsections `Embedded Storage Async` and `Block Device Driver`.

partition-manager/partition-manager/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+        if self.offset % block_size != 0 {
+            None
+        } else {
+            Some(self.offset / block_size)
+        }
+    }
+}
+
+/// Marker type for read-only partitions.
+pub struct RO;
```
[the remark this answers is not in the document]
Wassasin: It is possible to degrade a partition to readonly using the `Partition::readonly(&mut self)` function. Going from readonly to readwrite is not possible. We can add some marker token struct that allows you to upgrade if you have ownership, but in use I think first degrading, later dropping the readonly partition, and hence getting use back of your readwrite instance might be easier.

partition-manager/partition-manager/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// Marker type for read/write partitions.
+pub struct RW;
+
+/// An error that can be returned on operations for partitions.
+#[derive(Debug, PartialEq)]
+pub enum Error<E> {
```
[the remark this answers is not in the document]
Wassasin: `Inner` is used to denote errors of the underlying device, including if applicable any timeout errors. Initialization of the partition manager is infallible. Initialization of the underlying device is done outside of this crate.

partition-manager/partition-manager/src/ext/esa.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+impl<F: ReadNorFlash, MARKER, M: RawMutex> ErrorType for Partition<'_, F, MARKER, M> {
+    type Error = Error<F::Error>;
+}
+
+impl<F: ReadNorFlash, M: RawMutex> ReadNorFlash for Partition<'_, F, RO, M> {
+    const READ_SIZE: usize = F::READ_SIZE;
+
+    async fn read(&mut self, offset: u32, bytes: &mut [u8]) -> Result<(), Self::Error> {
```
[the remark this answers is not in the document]
Wassasin: Any filesystem is out of scope of the partition manager. Timeout errors are bubbled upwards from the underlying storage device directly to the end user.

partition-manager/partition-manager/src/ext/esa.rs
Wassasin: Currently the blocking API can not be called internally as we only support async traits of `embedded-storage-async` and `block-device-driver`.

If there is demand, I can also implement a blocking variant of partition manager using `embassy-sync::blocking_mutex`. That is not a lot of work, just roughly double the amount of code.

In practice the purpose of partition manager is to manage concurrent access. Given that storage operations typically take a long time, I would generally not advise using blocking operations for this without something like a tick-based preempting scheduler.

Wassasin:
I removed support for the `yaml2` format, so that we do not have to vet the associated dependencies. Support for `json` and `toml` remains.

Wassasin:
### With regards to cargo vec
1. What crates (with version) need to be audited?
a. Direct dependencies:
- block-device-driver:0.2.0
- serde_json:1.0.140
- toml:0.8.22

b. Indirect dependencies:
- aligned:0.4.2
- as-slice:0.2.1
- equivalent:1.0.2
- hashbrown:0.15.3
- indexmap:2.9.0
- itoa:1.0.15
- memchr:2.7.4
- ryu:1.0.20
- serde_spanned:0.6.8
- toml_datetime:0.6.9
- toml_edit:0.22.26
- winnow:0.7.10

3. How many of the crates are version updates vs new dependencies?
All of these crates are new dependencies.

4. To confirm none of the already included crates serve your needs, please provide a brief description of the purpose of the new crates.
- block-device-driver: we implement the provided traits in order to support NAND devices.
- serde_json: the requirements state that the JSON format must be supported.
- toml: this is an alternative format that is more widely supported within the Rust ecosystem, and supports typical features you would want when specifying partition maps, namely hexadecimal literals.

6. Any extra notes to the auditors to help with their audits.

jerrysxie:
> ### With regards to cargo vet
> 1. What crates (with version) need to be audited?
>    a. Direct dependencies:
> 
> * block-device-driver:0.2.0
> * serde_json:1.0.140
> * toml:0.8.22
> 
> b. Indirect dependencies:
> 
> * aligned:0.4.2
> * as-slice:0.2.1
> * equivalent:1.0.2
> * hashbrown:0.15.3
> * indexmap:2.9.0
> * itoa:1.0.15
> * memchr:2.7.4
> * ryu:1.0.20
> * serde_spanned:0.6.8
> * toml_datetime:0.6.9
> * toml_edit:0.22.26
> * winnow:0.7.10
> 
> 3. How many of the crates are version updates vs new dependencies?
>    All of these crates are new dependencies.
> 4. To confirm none of the already included crates serve your needs, please provide a brief description of the purpose of the new crates.
> 
> * block-device-driver: we implement the provided traits in order to support NAND devices.
> * serde_json: the requirements state that the JSON format must be supported.
> * toml: this is an alternative format that is more widely supported within the Rust ecosystem, and supports typical features you would want when specifying partition maps, namely hexadecimal literals.
> 
> 6. Any extra notes to the auditors to help with their audits.

Sorry, for the long delay on this, I am starting the audit now.

I just have a couple of questions on requirements since I was not a part of the requirements gathering phase (possibly reducing the amount of code I have to audit):

- For block device driver support, is that part of the requirement in addition to embedded-storage-async traits?
- For toml and json support, are we required to support both? Supporting just one is good enough IMO. Maybe just supporting toml?

@JamesHuard @gjpmsft

Wassasin:
The original requirements list JSON only. As writing hex numbers in JSON is not really great I also added TOML. YAML was also mentioned a few times but vetting any crate for that will be a challenge.

With regards to block-device-driver, the original requirements list 'support for NOR and NAND'. Currently block-device-driver seem to be the most mature interface for those kinds of memories. If there are no immediate plans of using NAND I can remove support for that.

jerrysxie:
> The original requirements list JSON only. As writing hex numbers in JSON is not really great I also added TOML. YAML was also mentioned a few times but vetting any crate for that will be a challenge.
> 
> With regards to block-device-driver, the original requirements list 'support for NOR and NAND'. Currently block-device-driver seem to be the most mature interface for those kinds of memories. If there are no immediate plans of using NAND I can remove support for that.

Thanks for the insight, I am done auditing the dependencies bought in by `block-device-driver:0.2.0` and `toml:0.8.22` for now. Draft PR: https://github.com/OpenDevicePartnership/embedded-services/pull/366. Working through dependencies for `serde_json:1.0.140` has been a struggle due to lots of unsafe usage + manipulation of raw buffers.

@JamesHuard @gjpmsft @asasine Any chance we can just support `toml` for now and add 'json` support later?

JamesHuard:
> > The original requirements list JSON only. As writing hex numbers in JSON is not really great I also added TOML. YAML was also mentioned a few times but vetting any crate for that will be a challenge.
> > With regards to block-device-driver, the original requirements list 'support for NOR and NAND'. Currently block-device-driver seem to be the most mature interface for those kinds of memories. If there are no immediate plans of using NAND I can remove support for that.
> 
> Thanks for the insight, I am done auditing the dependencies bought in by `block-device-driver:0.2.0` and `toml:0.8.22` for now. Draft PR: #366. Working through dependencies for `serde_json:1.0.140` has been a struggle due to lots of unsafe usage + manipulation of raw buffers.
> 
> @JamesHuard @gjpmsft @asasine Any chance we can just support `toml` for now and add `json` support later?

While we'd prefer json, the specific format language is not vital. We can live with toml for the time being

JamesHuard (approved):

Wassasin:
@jerrysxie I have rebase it on the `main` branch. However I am getting still some vetting issues with:
```
  hashbrown:0.15.4 missing ["safe-to-deploy"]
  memchr:2.7.4 missing ["safe-to-deploy"]
  serde_spanned:0.6.8 missing ["safe-to-deploy"]
```

In particular `memchr` is `safe-to-run` but not `safe-to-deploy`.

jerrysxie:
> @jerrysxie I have rebased it on the `main` branch. However I am getting still some vetting issues with:
> 
> ```
>   hashbrown:0.15.4 missing ["safe-to-deploy"]
>   memchr:2.7.4 missing ["safe-to-deploy"]
>   serde_spanned:0.6.8 missing ["safe-to-deploy"]
> ```
> 
> In particular `memchr` is `safe-to-run` but not `safe-to-deploy`.

@Wassasin This PR: https://github.com/OpenDevicePartnership/embedded-services/pull/373 should deal with the rest of the dependencies. Please rebase once the PR goes in. `memchr` and `hashbrown` are a bit time-consuming to audit, so we will exempt them for now to allow more time for audit.

JamesHuard (dismissed):

gjpmsft (dismissed):

JamesHuard (dismissed):
approval, assuming we fix the clippy stuff

partition-manager/partition-manager/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+impl<F, M: RawMutex> PartitionManager<F, M> {
+    /// Wrap a disk such that it can be concurrently accessed.
+    pub const fn new(storage: F) -> Self {
+        Self {
+            storage: Mutex::new(storage),
+        }
+    }
+
+    /// Map a disk to multiple partitions given a partition configuration definition.
+    pub fn map<C: PartitionConfig>(&mut self, config: C) -> C::Map<'_, F, M> {
```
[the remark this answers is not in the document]
ayodejiige: So I gave this a try and it is not very friendly in situations where partitions needs to have a static lifetime. I could not figure out a good method to have static mutable partition manager.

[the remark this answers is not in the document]
Wassasin: `static` properties in the embedded space that need to be initialized during runtime can be achieved using something like [static_cell](https://crates.io/crates/static_cell). In std you would not use statics but something like `Arc`.

With regards to the `&mut` requirement: it enforces that only a single instance of each partition exists at the same time. For example, if at a later date we would like to implement a read-cache, having a single instance would be a helpful property.

To help I have made a quick example on how to use `static_cell` here: https://github.com/Wassasin/pm-test-application/blob/master/src/main.rs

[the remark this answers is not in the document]
ayodejiige: `static_cell` did the trick. Thank you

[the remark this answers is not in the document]
Wassasin: Great!

PR #287, opened 2025-05-06
Add timeout before updating charger settings and after connecting new power policy consumer

tullom:
FYI @sukomath @asasine @RobertZ2011 

Add timeout as a temporary quick fix to charger data race bug.

Will be reverted once a proper fix is in later this week.


RobertZ2011:
If you haven't could you also create a bug to track the removal, so it doesn't get lost?

tullom:
> If you haven't could you also create a bug to track the removal, so it doesn't get lost?

Done, #289

jerrysxie (approved):

PR #288, opened 2025-05-06
power-policy/type-c: Fix various timing bugs

RobertZ2011:


RobertZ2011:
Check failures are fixed by https://github.com/OpenDevicePartnership/embedded-services/pull/290

asasine (approved):

RobertZ2011:
Currently rebased on top of https://github.com/OpenDevicePartnership/embedded-services/pull/285

type-c-service/src/wrapper/power.rs
```diff
@@ -186,71 +187,75 @@ impl<const N: usize, C: Controller> ControllerWrapper<'_, N, C> {
...
-        trace!("Power command: device{} {:#?}", local_id, command);
-        (command, LocalPortId(local_id as u8))
+    pub(super) async fn wait_power_command(
+        &self,
+    ) -> (
+        deferred::Request<'_, NoopRawMutex, CommandData, InternalResponseData>,
+        LocalPortId,
+    ) {
+        let futures: [_; N] = from_fn(|i| self.power[i].receive());
+        let (request, local_id) = select_array(futures).await;
```
asasine, line 211: Ordering here could cause the select future to drop requests that have been received but not responded to, guaranteeing the caller gets a timeout when it didn't need to be so.

1. `self.power[0]` gets a request, starts processing, and hits an await point.
2. `self.power[1]` same, but let's assume it has two await points and could end up leaving the system corrupted (it's not drop-safe).
4. `self.power[0]` wakens, completes, and sends the response.
5. `select_array` considers one of its futures complete, dropping the rest
6. `self.power[1]` never gets awakened
7. Requester for `self.power[1]` times out, despite request being started, potentially leaving the system in an invalid state.

Yes, this is generally a "select isn't good for futures that aren't drop-safe" sort of problem, but this is very likely with this type of system. A more resilient design would be to add received requests into a queue rather than only completing the first and dropping the rest. NB since this is the existing design and not the intent of this PR but using select here, and likely elsewhere, really needs to be fixed.

embedded-service/src/power/policy/action/policy.rs
```diff
@@ -42,23 +47,32 @@ impl<'a, S: Kind> Policy<'a, S> {
...
         Ok(())
     }
 
-    /// Connect this device as a provider
-    async fn connect_provider_internal(&self, capability: PowerCapability) -> Result<(), Error> {
+    /// Common disconnect function used by multiple states
+    async fn disconnect_internal(&self) -> Result<(), Error> {
+        match with_timeout(DEFAULT_TIMEOUT, self.disconnect_internal_no_timeout()).await {
+            Ok(r) => r,
+            Err(_) => Err(Error::Timeout),
```
asasine, line 66: nit: `with_timeout` doesn't return a `non_exhaustive` type, so the catch-all isn't necessary here. If/when a new error variant is added to the error, this will erronesously compile when it otherwise should have failed (owing to the lack of `non_exhaustive`) and we'll therefore return a timeout erroneously. Prefer an exact pattern in this case. We *want* compilation failures when breaking changes are made so that we are prevented from shipping bugs.
```suggestion
            Err(embassy_time::TimeoutError) => Err(Error::Timeout),
```

here and elsewhere

embedded-service/src/power/policy/action/policy.rs
```diff
@@ -89,6 +114,23 @@ impl<'a> Policy<'a, Idle> {
...
+        }
+    }
+
+    /// Connect this device as a provider
+    pub async fn connect_provider_no_timeout(
+        self,
+        capability: PowerCapability,
+    ) -> Result<Policy<'a, ConnectedProvider>, Error> {
+        self.connect_provider_internal_no_timeout(capability).await?;
+        Ok(Policy::new(self.device))
```
asasine, line 131: subjective style:
```suggestion
        self.connect_provider_internal_no_timeout(capability).await.map(|_| Policy::new(self.device))
```

embedded-service/src/type_c/controller.rs
```diff
@@ -493,14 +491,17 @@ impl ContextToken {
...
                 port: port_id,
                 operation: command,
             }))
             .await
         {
             Response::Lpm(response) => response,
-            _ => Err(PdError::InvalidResponse),
+            r => {
+                error!("Invalid response: expected LPM, got {:?}", r);
+                Err(PdError::InvalidResponse)
```
asasine, line 503: Is there a `ResponseKind` or similar enum that can be wrapped to indicate to the caller what the actual response was? It could really help with debugging.

asasine (commented):
all non-blocking but the use of select does create some unsoundness issues that needs addressing (elsewhere, probably not here)

asasine (dismissed):

embedded-service/src/type_c/controller.rs
```diff
@@ -398,7 +387,9 @@ pub(super) async fn lookup_controller(controller_id: ControllerId) -> Result<&'s
         .ok_or(PdError::InvalidController)
 }
 
-const DEFAULT_TIMEOUT: Duration = Duration::from_millis(250);
+/// Default command timeout
+/// set to high value since this is intended to prevent an unresponsive device from blocking the service implementation
+const DEFAULT_TIMEOUT: Duration = Duration::from_millis(5000);
```
madeleyneVaca, line 392: Would we want to make this end-user configurable at some point, maybe in the init?

type-c-service/src/wrapper/power.rs
RobertZ2011: I don't think this can happen as described because we don't execute multiple commands simultaneously. If `self.power[0]` gets a request then that request will get processed and a response sent. If `self.power[1]` gets a request then that will sit until the `self.power[0]` request finishes processing and this function gets called again to produce the next request. There shouldn't be any lost requests either because select will run the first ready future and the others will get dropped so any pending requests on the others will remain unread. 

That's the motivation between splitting out the logic into a separate function that gets the requests and one that processes the requests. The first draft implementation of either the power policy or the type-C service didn't do that and was susceptible to what you're describing. @jerrysxie let me know if I missed anything.

embedded-service/src/power/policy/action/policy.rs
RobertZ2011: Done.

embedded-service/src/power/policy/action/policy.rs
RobertZ2011: Done. I think I got all of them.

embedded-service/src/type_c/controller.rs
RobertZ2011: Created an issue to implement this: https://github.com/OpenDevicePartnership/embedded-services/issues/305

embedded-service/src/type_c/controller.rs
RobertZ2011: Created an issue to implement this: https://github.com/OpenDevicePartnership/embedded-services/issues/304

embedded-service/src/power/policy/action/policy.rs
```diff
@@ -89,35 +114,80 @@ impl<'a> Policy<'a, Idle> {
...
+        self.connect_provider_internal_no_timeout(capability)
+            .await
+            .map(|_| Policy::new(self.device))
+    }
+
     /// Connect this device as a provider
     pub async fn connect_provider(self, capability: PowerCapability) -> Result<Policy<'a, ConnectedProvider>, Error> {
-        self.connect_provider_internal(capability).await?;
-        Ok(Policy::new(self.device))
+        self.connect_provider_internal(capability)
```
tullom, line 137: Nit: Why do you need `connect_provider_internal()`? `connect_consumer()` doesn't have an analogous `connect_consumer_internal()` but the function layout looks very similar, the only difference is that `connect_provider_internal()` doesn't return a new Policy.

tullom (approved):
One small nit comment, otherwise nicely done.

embedded-service/src/power/policy/action/policy.rs
RobertZ2011: This is due to a slight difference in how the provider and consumer logic works. `connect_consumer` is only ever called from the idle state, but `connect_provider` can be called from the idle and connected provider state. It's available in the latter because a device can request a higher power level and we're mostly just giving an approval. While with the consumer state if we want to switch to a new consumer contract we call `disconnect` first so that there's no power applied while we do things like configure the charger.

type-c-service/src/wrapper/power.rs
asasine: I think you'll still get a timeout if you change my example to `self.power[1]` waiting on the mutex that `self.power[0]` has locked. Once `self.power[0]` completes, `self.power[1]` will be dropped, even though it could have executed, and that will result in a timeout to the `self.power[1]` caller.

RobertZ2011: I think I see what you're saying. The concern is the function that is waiting for a message could be doing something like:
```rust
let message = receive().await;
some_other_op().await
message
```

Then the message could get lost if one of the other futures completes first. Opened https://github.com/OpenDevicePartnership/embedded-services/issues/308 to look into it.

embedded-service/src/power/policy/action/policy.rs
```diff
@@ -1,8 +1,13 @@
 //! Policy state machine
+use embassy_time::{with_timeout, Duration, TimeoutError};
+
 use super::*;
 use crate::power::policy::{device, Error, PowerCapability};
 use crate::{error, info};
 
+/// Default timeout for device commands to prevent the policy from getting stuck
+const DEFAULT_TIMEOUT: Duration = Duration::from_millis(5000);
```
jerrysxie, line 9: Is the default timeout change part of the deferred IPC channel change? It seems like it could have been a separate commit or maybe even a separate PR to keep the code change PR smaller.

RobertZ2011: These changes are related because they're fixing several timing issues present in these service implementations. I updated the commit message to mention that this is to fix timing issues.

asasine (approved):

PR #291, opened 2025-05-06
type-c: Add function to sync controller state

RobertZ2011:
Add a function to sync the state of the controller and its ports. If a charger is plugged in before reset then there won't be any interrupts to inform the EC that anything is connected. Add a function to sync the software state of the controller with the hardware state.

### DIFF
--- a/type-c-service/src/driver/tps6699x.rs
+++ b/type-c-service/src/driver/tps6699x.rs
@@ -46,8 +46,8 @@ impl<'a, const N: usize, M: RawMutex, B: I2c> Tps6699x<'a, N, M, B> {
         tps6699x: &mut tps6699x::Tps6699x<'a, M, B>,
         port: LocalPortId,
     ) -> Result<PortEventKind, Error<B::Error>> {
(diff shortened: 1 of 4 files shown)


type-c-service/src/driver/tps6699x.rs
```diff
@@ -146,6 +132,20 @@ impl<'a, const N: usize, M: RawMutex, B: I2c> Tps6699x<'a, N, M, B> {
...
+        {
+            debug!("Port{}: new sink contract", port.0);
+            events.set_new_power_contract_as_consumer(true);
+        }
+
+        if port_status.available_source_contract.is_some()
+            && port_status.available_source_contract != previous_status.available_sink_contract
+        {
+            debug!("Port{}: New contract as consumer", port.0);
+            events.set_new_power_contract_as_consumer(true);
```
asasine, line 146: This seems wrong. Shouldn't this be a new contract as a provider?

asasine (approved):

jerrysxie (approved):

PR #293, opened 2025-05-12
power-policy: Remove automatic provider upgrade logic

RobertZ2011:
Remove logic to automatically upgrade a provider contract if power is available. This can result in a device receiving more power than needed while another device receives less. This is particularly a concern for TBT compliance testing. With this change devices will have to explicitly request a higher contract. This also removes the recovery logic which was only used in cases where we attempted to downgrade a contract. The new behavior does not attempt to downgrade contracts, so it's no longer needed.

### DIFF
--- a/docs/power-policy.md
+++ b/docs/power-policy.md
@@ -28,9 +28,6 @@ stateDiagram-v2
     Idle --> Detached : NotifyDetach
 ```
 
-#### Device recovery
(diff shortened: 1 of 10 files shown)


RobertZ2011:
Going to land this on top of https://github.com/OpenDevicePartnership/embedded-services/pull/288

asasine (approved):

RobertZ2011:
Found a couple more spots where we were messing with the implicit source settings. I'm going to submit a follow-up PR to handle explicit source contracts, but I wanted to land this since its currently causing issues with docks.

asasine (approved):

tullom (approved):

PR #294, opened 2025-05-13
Simplify charger state machine and expose policy driven PSU attach and detach handlers

### LINKED ISSUE #289: More robustly fix charger data race in power policy
We are hitting a race condition where charger can receive a new power configuration from power policy, but the charger device hasn't received (but will imminently receive) a signal that the charger is good to update power capability.

A fix to add a delay is merged #287 will temporarily fix this issue, but for robustness a more dedicated fix must be implemented.

tullom:
- Remove Idle state and instead, force charger PSU detection as part of init sequence
- Force reinitialzation if policy sends InitRequest
- Replace hardcoded charge current read and writes when detaching/attaching with trait fn's that allow OEMs to define their own attach/detach sequence handlers
- Cache power capability when a new capability is recv'd by the power policy. Fixes a data race where power policy has sent us a new capability but the charger might not have noticed PSU attached yet.
- OEM state support removed 

Closes #289

### DIFF
--- a/embedded-service/src/power/policy/charger.rs
+++ b/embedded-service/src/power/policy/charger.rs
@@ -18,6 +18,11 @@ pub trait ChargeController: embedded_batteries_async::charger::Charger {
     fn init_charger(&mut self) -> impl Future<Output = Result<(), Self::BusError>>;
     /// Returns if the charger hardware detects if a PSU is attached
     fn is_psu_attached(&mut self) -> impl Future<Output = Result<bool, Self::BusError>>;
+    /// Called after power policy attaches to a power port.
+    fn attach_handler(&mut self, capability: PowerCapability) -> impl Future<Output = Result<(), Self::BusError>>;
+    /// Called after power policy detaches from a power port, either to switch consumers,
+    /// or because PSU was disconnected.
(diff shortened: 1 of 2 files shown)


embedded-service/src/power/policy/charger.rs
```diff
@@ -18,34 +18,40 @@ pub trait ChargeController: embedded_batteries_async::charger::Charger {
     fn init_charger(&mut self) -> impl Future<Output = Result<(), Self::BusError>>;
     /// Returns if the charger hardware detects if a PSU is attached
     fn is_psu_attached(&mut self) -> impl Future<Output = Result<bool, Self::BusError>>;
+    /// Called after power policy attaches to a power port.
+    fn attach_handler(&mut self, capability: PowerCapability) -> impl Future<Output = Result<(), Self::BusError>>;
+    /// Called after power policy detaches from a power port, either to switch consumers,
+    /// or because PSU was disconnected.
+    fn detach_handler(&mut self) -> impl Future<Output = Result<(), Self::BusError>>;
```
RobertZ2011, line 25: Long term we should move over to wrapping `BusError` in our own type that can also contain a charger error. This will let implementors use our existing error values. But this can stay as is for now.

RobertZ2011 (approved):

felipebalbi (approved):

PR #296, opened 2025-05-15
Remove per commit checking from workflows

jerrysxie:


JamesHuard (approved):

asasine (changes requested):
What Copilot said

jerrysxie:
> What Copilot said

This is actually harmless for yaml files.

asasine:
> > What Copilot said
> 
> This is actually harmless for yaml files. But I removed all the spurious newlines anything because I prefer it to be more compact.

My mistake; the diff actually looked like there was a stray `+` in the text itself. Perhaps Copilot's "thinking" and my own have some similarities after all.

asasine (approved):

felipebalbi (approved):

PR #298, opened 2025-05-16
Add Retimer FW status get/set command

zhuyi2024:
Add Retimer FW status get/set command


embedded-service/src/type_c/controller.rs
```diff
@@ -114,6 +120,10 @@ pub enum PortResponseData {
     PortStatus(PortStatus),
     /// ClearEvents
     ClearEvents(PortEventKind),
+    /// Retimer Fw Update status
+    RtFwUpdateStatus(bool),
+    /// Retimer Fw Update set state success with no error
+    RtFwUpdateSetState,
```
RobertZ2011, line 126: I would just use `Complete` instead. I generally only introduce new values into this enum if there's data that needs to be returned.

embedded-service/src/type_c/external.rs
```diff
@@ -62,6 +68,12 @@ pub struct PortCommand {
 pub enum PortResponseData {
     /// Get port status
     PortStatus(PortStatus),
+    /// Get retimer fw update status
+    RetimerFwUpdateGetState(bool),
+    /// Set retimer fw update status
+    RetimerFwUpdateSetState,
```
RobertZ2011, line 74: Same thing here, though you'll have to introduce the `Complete` value.

embedded-service/src/type_c/external.rs
```diff
@@ -133,3 +145,45 @@ pub async fn controller_port_to_global_id(
 ) -> Result<GlobalPortId, PdError> {
     lookup_controller(controller_id).await?.lookup_global_port(port_id)
 }
+
+/// Get the retimer fw update status of the given port
+#[allow(unreachable_patterns)]
```
RobertZ2011, line 150: I don't think these linter hints are needed anymore. I originally added them to keep the `_ => Err(...)` branches because I knew we'd be adding more variants to the `PortResponseData` enum.

embedded-service/src/type_c/controller.rs
```diff
@@ -319,6 +329,18 @@ pub trait Controller {
     /// Returns the port status
     fn get_port_status(&mut self, port: LocalPortId)
         -> impl Future<Output = Result<PortStatus, Error<Self::BusError>>>;
+    /// Returns the retimer fw update state
```
RobertZ2011, line 332: You'll also need to update the examples to get the checks to pass.

embedded-service/src/type_c/external.rs
```diff
@@ -133,3 +145,45 @@ pub async fn controller_port_to_global_id(
...
+        _ => Err(PdError::InvalidResponse),
+    }
+}
+
+/// Clear the retimer fw update state of the given port
+#[allow(unreachable_patterns)]
+pub async fn port_clear_rt_fw_update_state(port: GlobalPortId) -> Result<(), PdError> {
+    match execute_external_port_command(Command::Port(PortCommand {
+        port,
+        data: PortCommandData::RetimerFwUpdateSetState,
```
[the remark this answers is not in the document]
zhuyi2024: Fixed

embedded-service/src/type_c/controller.rs
zhuyi2024: Fixed.

type-c-service/src/wrapper/pd.rs
```diff
@@ -28,6 +28,33 @@ impl<const N: usize, C: Controller> ControllerWrapper<'_, N, C> {
...
+                    Ok(status) => Ok(controller::PortResponseData::RtFwUpdateSetState),
+                    Err(e) => match e {
+                        Error::Bus(_) => Err(PdError::Failed),
+                        Error::Pd(e) => Err(e),
+                    },
+                }
+            }
+            controller::PortCommandData::RetimerFwUpdateClearState => {
+                match controller.clear_rt_fw_update_state(local_port).await {
+                    Ok(status) => Ok(controller::PortResponseData::RtFwUpdateSetState),
```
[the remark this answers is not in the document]
zhuyi2024: Fixed

embedded-service/src/type_c/controller.rs
```diff
@@ -589,6 +611,39 @@ impl ContextToken {
...
+        }
+    }
+
+    /// Clear the retimer fw update state
+    pub async fn clear_rt_fw_update_state(&self, port: GlobalPortId) -> Result<(), PdError> {
+        match self
+            .send_port_command(port, PortCommandData::RetimerFwUpdateClearState, DEFAULT_TIMEOUT)
+            .await?
+        {
+            PortResponseData::RtFwUpdateSetState => Ok(()),
```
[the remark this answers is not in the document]
zhuyi2024: Fixed

embedded-service/src/type_c/controller.rs
zhuyi2024: Sorry, where is the examples?

embedded-service/src/type_c/external.rs
zhuyi2024: Fixed

Cargo.toml
```diff
@@ -54,4 +54,4 @@ log = "0.4"
 postcard = "1.*"
 rand_core = "0.6.4"
 serde = { version = "1.0.*", default-features = false }
-tps6699x = { git = "https://github.com/OpenDevicePartnership/tps6699x" }
+tps6699x = { git = "https://github.com/zhuyi2024/tps6699x", branch = "personal/zhuyi/RtGetSet" }
```
RobertZ2011, line 57: Left over development dependency. It's generally better to use https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section anyway.

embedded-service/src/type_c/controller.rs
RobertZ2011: I think'll you'll only need the checks the ones here: https://github.com/OpenDevicePartnership/embedded-services/tree/main/examples/std/src/bin/type_c, but the examples CI check should also show you which ones are failing.

RobertZ2011 (changes requested):
From the CI failures it looks like you may need to update and commit the `Cargo.lock` files. You'll also need to rebase.

Cargo.toml
zhuyi2024: There is some errors in checks, I updated this temporarily, will revert back after the TPS6699x PR completes.

asasine (dismissed):
Tentatively approving. You still need to update the manifest file when OpenDevicePartnership/tps6699x#40 goes in.

RobertZ2011 (approved):

felipebalbi (approved):

PR #306, opened 2025-05-16
Add minimal button tests

felipebalbi:
This commit adds a minimal set of tests for the button logic. Notice that this is, indeed, a breaking a change and coordination with dependent crates is required.

### DIFF
--- a/power-button-service/src/debounce.rs
+++ b/power-button-service/src/debounce.rs
@@ -1,7 +1,7 @@
 //! Debounce Module
 
-use embassy_time::{Duration, Timer};
 use embedded_hal::digital::InputPin;
+use embedded_hal_async::delay::DelayNs;
 use embedded_hal_async::digital::Wait;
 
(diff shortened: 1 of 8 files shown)


felipebalbi:
Still need to integrate `cargo test` into workflows. It will require coordination with dependents. Don't merge unless you know what you're doing. You've been warned.

power-button-service/src/debounce.rs
```diff
@@ -74,9 +82,297 @@ impl Default for Debouncer {
...
+
+        assert_eq!(d.integrator, 1);
+        assert!(!d.pressed);
+
+        gpio.done();
+    }
+
+    #[test]
+    fn test_last_update_state_change() {
+        // Press, active high
```
RobertZ2011, line 134: Isn't this active low because that's the default for `Debouncer`?

RobertZ2011 (changes requested):

PR #307, opened 2025-05-16
checks: Change to running tests per crate

RobertZ2011:
Running tests from the workspace root can cause build issues with embassy. Move to running tests for each crate individually instead.


.github/workflows/check.yml
```diff
@@ -144,7 +144,17 @@ jobs:
...
+          cargo test --locked -p battery-service
+          cargo test --locked -p cfu-service
+          cargo test --locked -p embedded-services
+          cargo test --locked -p espi-service
+          cargo test --locked -p hid-service
+          cargo test --locked -p platform-service
+          cargo test --locked -p power-button-service
+          cargo test --locked -p power-policy-service
+          cargo test --locked -p storage_bus
+          cargo test --locked -p type-c-service
```
felipebalbi, line 157: too much repetition. Add a for loop?

felipebalbi (changes requested):

.github/workflows/check.yml
RobertZ2011: Done, moved this over to use a workflow matrix.

.github/workflows/check.yml
```diff
@@ -133,18 +133,32 @@ jobs:
           log-level: warn
           command: check
           arguments: --all-features --locked
-
+
```
tullom, line 136: nit: small whitespace change

tullom (approved):
looks good, just small inadvertent whitespace change.

.github/workflows/check.yml
RobertZ2011: Done.

asasine (approved):

felipebalbi (approved):

PR #325, opened 2025-05-28
type-c-service: Add integrated CFU support

RobertZ2011:
This PR adds CFU support directly integrated within the existing type-C code. This is meant to get this functionality in for the moment while work on more general FW update support continues.

### DIFF
--- a/embedded-service/src/type_c/controller.rs
+++ b/embedded-service/src/type_c/controller.rs
@@ -370,6 +370,23 @@ pub trait Controller {
     fn get_controller_status(
         &mut self,
     ) -> impl Future<Output = Result<ControllerStatus<'static>, Error<Self::BusError>>>;
+
+    // TODO: remove all these once we migrate to a generic FW update trait
+    // https://github.com/OpenDevicePartnership/embedded-services/issues/242
+    /// Get current FW version
+    fn get_active_fw_version(&self) -> impl Future<Output = Result<u32, Error<Self::BusError>>>;
+    /// Start a firmware update
+    fn start_fw_update(&mut self) -> impl Future<Output = Result<(), Error<Self::BusError>>>;
+    /// Abort a firmware update
+    fn abort_fw_update(&mut self) -> impl Future<Output = Result<(), Error<Self::BusError>>>;
+    /// Finalize a firmware update
+    fn finalize_fw_update(&mut self) -> impl Future<Output = Result<(), Error<Self::BusError>>>;
+    /// Write firmware update contents
+    fn write_fw_contents(
+        &mut self,
+        offset: usize,
+        data: &[u8],
+    ) -> impl Future<Output = Result<(), Error<Self::BusError>>>;
 }
 
 /// Internal context for managing PD controllers
--- a/examples/rt685s-evk/Cargo.toml
+++ b/examples/rt685s-evk/Cargo.toml
@@ -47,6 +47,7 @@ futures = { version = "0.3.30", default-features = false, features = [
 mimxrt600-fcb = "0.1.0"
 mimxrt685s-pac = { version = "*", features = ["rt", "critical-section"] }
 
+embedded-cfu-protocol = { git = "https://github.com/OpenDevicePartnership/embedded-cfu" }
 embedded-services = { path = "../../embedded-service", features = ["defmt"] }
 power-button-service = { path = "../../power-button-service", features = [
     "defmt",
--- a/examples/rt685s-evk/src/bin/type_c_cfu.rs
+++ b/examples/rt685s-evk/src/bin/type_c_cfu.rs
@@ -1,7 +1,7 @@
 #![no_std]
 #![no_main]
 
-use ::tps6699x::ADDR0;
+use ::tps6699x::ADDR1;
 use defmt::info;
 use embassy_embedded_hal::shared_bus::asynch::i2c::I2cDevice;
 use embassy_executor::Spawner;
@@ -12,23 +12,21 @@ use embassy_imxrt::{bind_interrupts, peripherals};
 use embassy_sync::blocking_mutex::raw::NoopRawMutex;
 use embassy_sync::mutex::Mutex;
 use embassy_sync::once_lock::OnceLock;
+use embassy_time::Timer;
 use embassy_time::{self as _, Delay};
-use embedded_cfu_protocol::protocol_definitions::{FwUpdateOffer, FwUpdateOfferResponse, FwVersion, HostToken};
-use embedded_services::comms;
+use embedded_cfu_protocol::protocol_definitions::*;
+use embedded_cfu_protocol::protocol_definitions::{FwUpdateOffer, FwUpdateOfferResponse, FwVersion};
+use embedded_services::cfu;
+use embedded_services::cfu::component::InternalResponseData;
+use embedded_services::cfu::component::RequestData;
 use embedded_services::power::policy::DeviceId as PowerId;
 use embedded_services::type_c::{self, ControllerId};
 use embedded_usb_pd::GlobalPortId;
 use static_cell::StaticCell;
 use tps6699x::asynchronous::embassy as tps6699x;
-use type_c_service::driver::tps6699x::{self as tps6699x_driver, Tps66994Wrapper};
+use type_c_service::driver::tps6699x::{self as tps6699x_drv};
 
 extern crate rt685s_evk_example;
-
-const CONTROLLER0_ID: ControllerId = ControllerId(0);
-const PORT0_ID: GlobalPortId = GlobalPortId(0);
-const PORT1_ID: GlobalPortId = GlobalPortId(1);
-const PORT0_PWR_ID: PowerId = PowerId(0);
-const PORT1_PWR_ID: PowerId = PowerId(1);
 
 bind_interrupts!(struct Irqs {
     FLEXCOMM2 => embassy_imxrt::i2c::InterruptHandler<peripherals::FLEXCOMM2>;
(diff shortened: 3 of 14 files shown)


examples/rt685s-evk/Cargo.toml
```diff
@@ -71,3 +72,6 @@ platform-service = { path = "../../platform-service", features = [
     "defmt",
     "imxrt685",
 ] }
+
+[patch.'https://github.com/OpenDevicePartnership/tps6699x']
```
ankurung, line 76: remove patch use

examples/rt685s-evk/src/bin/type_c_cfu.rs
```diff
@@ -0,0 +1,199 @@
...
+extern crate rt685s_evk_example;
+
+bind_interrupts!(struct Irqs {
+    FLEXCOMM2 => embassy_imxrt::i2c::InterruptHandler<peripherals::FLEXCOMM2>;
+});
+
+struct Validator;
+
+impl type_c_service::wrapper::FwVersionValidator for Validator {
+    fn validate(&self, _current: FwVersion, _new: FwVersion) -> bool {
```
ankurung, line 38: validator would need to provide the offer packet for decision making. Decisions are typically based on component info, version, vendor_specific, and misc_and_protocol_version contents.

type-c-service/src/driver/tps6699x.rs
```diff
@@ -335,47 +357,128 @@ impl<const N: usize, M: RawMutex, B: I2c> Controller for Tps6699x<'_, N, M, B> {
...
+        Ok(customer_use.custom_fw_version())
+    }
+
+    async fn start_fw_update(&mut self) -> Result<(), Error<Self::BusError>> {
+        let mut tps6699x = self.tps6699x.borrow_mut();
+        let mut delay = Delay;
+        let mut updater: BorrowedUpdater<tps6699x_drv::Tps6699x<'_, M, B>> = Default::default();
+
+        // Abandon any previous in-progress update
+        if let Some(update) = self.update_state.replace(None) {
```
ankurung, line 373: consider adding abort_inprogress state, while abort is in progress.  can update.updater.abort_fw_update fail?

type-c-service/src/driver/tps6699x.rs
```diff
@@ -335,47 +359,127 @@ impl<const N: usize, M: RawMutex, B: I2c> Controller for Tps6699x<'_, N, M, B> {
...
+            warn!("Abandoning in-progress update");
+            let mut delay = Delay;
```
gjpmsft, line 377: please document the side-effect that abort operations  will reset the FW. Though, we should be bubbling up errors that require reset to resolve and let the product make that decision. Resetting the contorller has implications on system power continuity so its a decision that that should not be made by the driver.

Thoughts?

type-c-service/src/driver/tps6699x.rs
RobertZ2011: I ended up moving this logic up the stack a little bit as part of some broader recovery logic. If the device ends up in an unexpected state then we'll enter a recovery mode where we'll repeatedly attempt to recover the device until we succeed..

embedded-service/src/type_c/controller.rs
```diff
@@ -340,6 +340,18 @@ pub trait Controller {
     fn get_controller_status(
         &mut self,
     ) -> impl Future<Output = Result<ControllerStatus<'static>, Error<Self::BusError>>>;
+
+    //TODO: remove all these once we migrate to a generic FW update trait
+    /// Get current FW version
```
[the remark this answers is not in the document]
RobertZ2011: Done.

type-c-service/src/driver/tps6699x.rs
RobertZ2011: Added a comment noting that some functions can reset the controller. I agree that we should allow for better control over when we perform the reset, but I think the current functionality is fine for now. This is probably going to need to be part of a broader integration with system-level power states/OEM-specific code.

type-c-service/src/wrapper/mod.rs
```diff
@@ -3,46 +3,96 @@
...
+            )),
+        }
+    }
+}
+
+/// Trait for validating firmware versions before applying an update
+// TODO: remove this once we have a better framework for OEM customization
+// See https://github.com/OpenDevicePartnership/embedded-services/issues/326
+pub trait FwVersionValidator {
+    fn validate(&self, current: FwVersion, offer: &FwUpdateOffer) -> bool;
```
asasine, line 53: Document what the return value means

type-c-service/src/wrapper/mod.rs
```diff
@@ -96,9 +146,15 @@ impl<'a, const N: usize, C: Controller> ControllerWrapper<'a, N, C> {
...
     /// None of the event processing functions return errors to allow processing to continue for other ports on a controller
-    async fn process_event(&self, controller: &mut C) {
+    async fn process_event(&self, controller: &mut C, state: &mut InternalState) {
         let mut port_events = PortEventFlags::none();
 
+        if state.fw_update_state.in_progress() {
+            // Don't process events while firmware update is in progress
+            debug!("Firmware update in progress, ignoring port events");
+            return;
+        }
```
asasine, line 156: Would this lead to dropped events if they occur during CFU? E.g., plug attach being ignored causing power policy to not tstart charging

type-c-service/src/wrapper/pd.rs
```diff
@@ -31,8 +42,14 @@ impl<const N: usize, C: Controller> ControllerWrapper<'_, N, C> {
     async fn process_controller_command(
         &self,
         controller: &mut C,
+        state: &mut InternalState,
         command: &controller::InternalCommandData,
     ) -> Response<'static> {
+        if state.fw_update_state.in_progress() {
+            debug!("FW update in progress, ignoring controller command");
+            return controller::Response::Controller(Err(PdError::Busy));
+        }
```
asasine, line 51: Could this check go in `process_pd_command()` or even the top-level `process()`?

type-c-service/src/wrapper/cfu.rs
```diff
@@ -0,0 +1,281 @@
...
+use super::*;
+
+/// Current state of the firmware update process
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FwUpdateState {
+    /// None in progress
+    Idle,
+    /// Firmware update in progress
+    /// Contains number of ticks [`super::DEFAULT_FW_UPDATE_TICK_INTERVAL_MS`] that have passed
+    InProgress(u8),
```
asasine, line 18: Is it possible that the `u8` could overflow?

type-c-service/src/wrapper/cfu.rs
```diff
@@ -0,0 +1,281 @@
...
+    async fn process_give_offer(&self, target: &mut C, offer: &FwUpdateOffer) -> Option<InternalResponseData> {
+        if offer.component_info.component_id != self.cfu_device.component_id() {
+            return None;
+        }
+
+        let version = match target.get_active_fw_version().await {
+            Ok(v) => v,
+            Err(Error::Pd(e)) => {
+                error!("Failed to get active firmware version: {:?}", e);
+                return Some(InternalResponseData::ComponentBusy);
```
asasine, line 63: This error doesn't indicate the underlying cause in any way, making debugging more difficult. Can you wrap the inner `Error::Pd(e)`?

Same suggestion for other indiscriminant uses of `ComponentBusy`

type-c-service/src/driver/tps6699x.rs
```diff
@@ -9,41 +9,63 @@ use bitfield::bitfield;
...
+    #[allow(dead_code)]
+    guards: [Option<tps6699x_drv::InterruptGuard<'a, M, B>>; 2],
+}
 
 pub struct Tps6699x<'a, const N: usize, M: RawMutex, B: I2c> {
     port_events: [Cell<PortEventKind>; N],
     port_status: [Cell<PortStatus>; N],
     sw_event: Signal<M, ()>,
-    tps6699x: RefCell<tps6699x::Tps6699x<'a, M, B>>,
+    tps6699x: RefCell<tps6699x_drv::Tps6699x<'a, M, B>>,
```
asasine, line 50: I agree with clippy here in that we're holding the borrow across await points, and since there's no locking built into this type, there's potential for attempting a `borrow_mut` twice (e.g., a `join` on multiple of this type's methods) which would cause a panic. A `Mutex` is a more appropriate synchronization primitive here.

type-c-service/src/wrapper/cfu.rs
```diff
@@ -0,0 +1,281 @@
...
+            }
+            Err(Error::Bus(_)) => {
+                error!("Failed to get active firmware version, bus error");
+                return Some(InternalResponseData::ComponentBusy);
+            }
+        };
+
+        if !self.fw_version_validator.validate(FwVersion::new(version), offer) {
+            debug!("Version is not acceptable, rejecting offer");
+            return None;
```
asasine, line 73: Should this update `self.update_state` to `FwUpdateState::Idle` or is that the caller's responsibility?

type-c-service/src/wrapper/cfu.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    /// Process a GiveContent command
+    async fn process_give_content(
+        &self,
+        controller: &mut C,
+        state: &mut InternalState,
+        content: &FwUpdateContentCommand,
+    ) -> Option<InternalResponseData> {
+        let data = &content.data[0..content.header.data_length as usize];
+        if content.header.flags & FW_UPDATE_FLAG_FIRST_BLOCK != 0 {
```
ankurung, line 90: no special processing needed after the successful write of LAST_BLOCK ?  When could the controller validate the bits when it has everything?

type-c-service/src/wrapper/cfu.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// Process a CFU command
+    pub async fn process_cfu_command(
+        &self,
+        controller: &mut C,
+        state: &mut InternalState,
+        command: &RequestData,
+    ) -> Option<InternalResponseData> {
+        if state.fw_update_state == FwUpdateState::Recovery {
+            debug!("FW update in recovery state, rejecting command");
+            return None;
```
ankurung, line 221: We need to respond with a CFU response packet if we are not able to handle any cfu command.

type-c-service/src/wrapper/mod.rs
asasine: Are we utilizing the dual bank feature? Does it track which bank to boot from or do we specify it?

type-c-service/src/wrapper/cfu.rs
```diff
@@ -0,0 +1,281 @@
...
+                debug!("Got other command: {:#?}", command);
+                None
+            }
+        }
+    }
+
+    /// Sends a CFU response to the command
+    pub async fn send_cfu_response(&self, response: Option<InternalResponseData>) {
+        self.cfu_device
+            .send_response(response.unwrap_or(InternalResponseData::ComponentPrepared))
```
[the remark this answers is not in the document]
madeleyneVaca: The placeholder is fine. The only responses that would write bits back to the host are the offer response and content response so as long as you're not using those in the wrong place, it's okay

examples/rt685s-evk/src/bin/type_c_cfu.rs
```diff
@@ -0,0 +1,223 @@
...
+        info!("Sending chunk {} of {}", i, fw.len());
+        let response = device
+            .execute_device_request(RequestData::GiveContent(request))
+            .await
+            .unwrap();
+        info!("Got response: {:?}", response);
+    }
+
+    device
+        .execute_device_request(RequestData::FinalizeUpdate)
```
madeleyneVaca, line 134: You don't need to send this FinalizeUpdate request, it's not part of the spec for the Host to ask for that. The host gives the component the info through the FW update flags so once the component receives the LAST_BLOCK flag, it should, of its own accord, finalize the update. This particular variant of the RequestData enum could probably be optimized out but I'd included it for clarity. At the moment in the EC component implementation, instead of having a helper function in the module that's called if it sees the LAST_BLOCK flag, it routes a request (to itself) via the service to finalize and then it responds to the content command.

examples/rt685s-evk/src/bin/type_c_cfu.rs
```diff
@@ -0,0 +1,223 @@
...
+    let prev_version = match response {
+        InternalResponseData::FwVersionResponse(GetFwVersionResponse { component_info, .. }) => {
+            Into::<u32>::into(component_info[0].fw_version)
+        }
+        _ => panic!("Unexpected response"),
+    };
+    info!("Got version: {:#x}", prev_version);
+
+    info!("Giving offer");
+    let offer = device
```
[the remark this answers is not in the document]
RobertZ2011: Done, added in a non-zero number.

type-c-service/src/wrapper/cfu.rs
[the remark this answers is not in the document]
RobertZ2011: Got rid of this and made everything have to provide a response.

type-c-service/src/wrapper/mod.rs
RobertZ2011: For now we should assume that the PD controller is not capable of producing normal port events during a FW update. The tps6699x specifically has a dedicated FW update mode where normal operation is suspended. It does have two flash banks, but these are used to provide a backup in case of a bad update. If we ever work with a PD controller that does support running updates then we can revisit this logic.

type-c-service/src/wrapper/cfu.rs
RobertZ2011: It shouldn't. The timeout tick count is a `u8` so at worst we'd end up with 255 and we'd enter recovery after comparing that to the timeout tick count of 255.

type-c-service/src/wrapper/cfu.rs
```diff
@@ -0,0 +1,281 @@
...
+                error!("Failed to get active firmware version: {:?}", e);
+                return Some(InternalResponseData::ComponentBusy);
+            }
+            Err(Error::Bus(_)) => {
+                error!("Failed to get active firmware version, bus error");
+                return Some(InternalResponseData::ComponentBusy);
+            }
+        };
+
+        if !self.fw_version_validator.validate(FwVersion::new(version), offer) {
```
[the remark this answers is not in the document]
RobertZ2011: This function now returns a complete offer response.

type-c-service/src/wrapper/cfu.rs
RobertZ2011: Done.

type-c-service/src/wrapper/mod.rs
```diff
@@ -3,46 +3,97 @@
...
+                DEFAULT_FW_UPDATE_TICK_INTERVAL_MS,
+            )),
+        }
+    }
+}
+
+/// Trait for validating firmware versions before applying an update
+// TODO: remove this once we have a better framework for OEM customization
+// See https://github.com/OpenDevicePartnership/embedded-services/issues/326
+pub trait FwVersionValidator {
```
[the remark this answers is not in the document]
RobertZ2011: Renamed.

type-c-service/src/wrapper/cfu.rs
RobertZ2011: This isn't needed here. We're still in `Idle` until we actually start getting the FW contents.

type-c-service/src/driver/tps6699x.rs
RobertZ2011: I created a bug to clean this up: https://github.com/OpenDevicePartnership/embedded-services/issues/328. I don't think there's currently a possibility for a panic here because the only functions that borrow from the `RefCell` take `&mut self` so it's redundant.

asasine (dismissed):

RobertZ2011:
Latest changes should be minor and related to threading through configuration for the FW update.

madeleyneVaca (approved):

PR #330, opened 2025-06-03
Update CODEOWNERS to use EC team

akshat2112:
With growing number of repositories, having consistent and manageable list of codeowners becomes important. This PR updates the CODEOWNERS file to use the EC code owners team as the default codeowners for the repository.
This hardens the repository against stale permission issues down the line by tying the access to being part of the team rather than being explicitly granted permission to a repository.

This PR also adds `--locked` to a workflow check that was missed in the previous PRs.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -202,7 +202,7 @@ jobs:
       - name: cargo check
         working-directory: ${{ matrix.example_directory }}
         run: |
-          cargo check --target thumbv8m.main-none-eabihf
+          cargo check --target thumbv8m.main-none-eabihf --locked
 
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yml
```diff
@@ -202,7 +202,7 @@ jobs:
       - name: cargo check
         working-directory: ${{ matrix.example_directory }}
         run: |
-          cargo check --target thumbv8m.main-none-eabihf
+          cargo check --target thumbv8m.main-none-eabihf --locked
```
felipebalbi, line 205: accidental change?

felipebalbi (approved):

PR #332, opened 2025-06-06
Cfu splitter

RobertZ2011:
Adds code that can broadcast a CFU update to multiple other CFU devices.

### DIFF
--- a/cfu-service/src/splitter.rs
+++ b/cfu-service/src/splitter.rs
@@ -1,0 +1,259 @@
+//! Module that can broadcast CFU messages to multiple devices
+//! This allows devices to share a single component ID
+
+use core::{future::Future, iter::zip};
+
+use embassy_futures::join::{join, join3, join4};
+use embedded_cfu_protocol::protocol_definitions::*;
+use embedded_services::{
+    cfu::{
+        self,
+        component::{CfuDevice, InternalResponseData, RequestData},
+    },
+    error, intrusive_list, trace,
+};
+
+/// Trait containing customization functionality for [`Splitter`]
+pub trait Customization {
+    /// Decides which firmware version to use based on the provided versions from all devices.
+    fn resolve_fw_versions(&self, versions: &[GetFwVersionResponse]) -> GetFwVersionResponse;
+
+    /// Decides which offer response to send based on the provided responses from all devices.
+    fn resolve_offer_response(&self, offer_responses: &[FwUpdateOfferResponse]) -> FwUpdateOfferResponse;
+
+    /// Decides which content response to send based on the provided responses from all devices.
+    fn resolve_content_response(&self, content_responses: &[FwUpdateContentResponse]) -> FwUpdateContentResponse;
+}
+
+/// Splitter struct
+pub struct Splitter<'a, C: Customization> {
+    /// CFU device
+    cfu_device: CfuDevice,
+    /// Component ID for each individual device
+    devices: &'a [ComponentId],
+    /// Customization for the Splitter
+    customization: C,
+}
+
+/// Maximum number of devices supported
+pub const MAX_SUPPORTED_DEVICES: usize = 4;
+
+impl<'a, C: Customization> Splitter<'a, C> {
+    /// Create a new Splitter, returns None if the devices slice is empty or too large
+    pub fn new(component_id: ComponentId, devices: &'a [ComponentId], customization: C) -> Option<Self> {
+        if devices.is_empty() || devices.len() > MAX_SUPPORTED_DEVICES {
+            None
+        } else {
+            Some(Self {
+                cfu_device: CfuDevice::new(component_id),
+                devices,
+                customization,
+            })
+        }
+    }
+
+    /// Create a new invalid FW version response
+    fn create_invalid_fw_version_response(&self) -> InternalResponseData {
+        let dev_inf = FwVerComponentInfo::new(FwVersion::new(0xffffffff), self.cfu_device.component_id());
+        let comp_info: [FwVerComponentInfo; MAX_CMPT_COUNT] = [dev_inf; MAX_CMPT_COUNT];
+        InternalResponseData::FwVersionResponse(GetFwVersionResponse {
+            header: GetFwVersionResponseHeader::new(1, GetFwVerRespHeaderByte3::NoSpecialFlags),
+            component_info: comp_info,
+        })
+    }
+
+    /// Create a content rejection response
+    fn create_content_rejection(sequence: u16) -> InternalResponseData {
+        InternalResponseData::ContentResponse(FwUpdateContentResponse::new(
+            sequence,
+            CfuUpdateContentResponseStatus::ErrorInvalid,
+        ))
+    }
+
+    /// Process a fw version request
+    async fn process_get_fw_version(&self) -> InternalResponseData {
+        let mut versions = [GetFwVersionResponse {
+            header: Default::default(),
+            component_info: Default::default(),
+        }; MAX_SUPPORTED_DEVICES];
+
+        let success = map_slice_join(self.devices, &mut versions, |device_id| async move {
+            if let Ok(InternalResponseData::FwVersionResponse(version_info)) =
+                cfu::route_request(*device_id, RequestData::FwVersionRequest).await
+            {
+                Some(version_info)
+            } else {
+                error!("Failed to get FW version for device {}", device_id);
+                None
+            }
+        })
+        .await;
+
+        if !success {
+            self.create_invalid_fw_version_response()
+        } else {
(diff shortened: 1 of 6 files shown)


madeleyneVaca (approved):
minor questions

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,205 @@
...
+        if self.devices.len() > MAX_SUPPORTED_DEVICES {
+            error!("More devices than supported");
+            return self.create_invalid_fw_version_response();
+        }
+
+        for (device_id, version) in zip(self.devices.iter(), versions.iter_mut()) {
+            if let Ok(InternalResponseData::FwVersionResponse(version_info)) =
+                cfu::route_request(*device_id, RequestData::FwVersionRequest).await
+            {
+                *version = version_info.component_info[0].fw_version;
```
asasine, line 90: What does `component_info[0]` refer to here and is it guaranteed to always be valid? I `component_info.len() == 0`, this will panic

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,205 @@
...
+        if self.devices.len() > MAX_SUPPORTED_DEVICES {
+            error!("More devices than supported");
+            return Self::create_offer_rejection();
+        }
+
+        for (device_id, offer_response) in zip(self.devices.iter(), offer_responses.iter_mut()) {
+            let mut offer = *offer;
+
+            // Override with the correct component ID for the device
+            offer.component_info.component_id = *device_id;
```
asasine, line 119: Struct update syntax (idk the type names so will need updating)
```suggestion
            let offer = Offer {
                component_info: ComponentInfo {
                    component_id: *device_id,
                    ..offer.component_info
                },
                ..offer
            };
```

cfu-service/src/splitter.rs
RobertZ2011: This would be the version of the over-all component, the other entries are used for subcomponents. This is an array so the first index is always valid. I ended up making this work with the actual responses instead of individual FW versions so code can also decide how to handle differing versions in sub-components.

asasine: > This is an array so the first index is always valid.

Arrays can be empty. What guarantees that the array is non-empty? A non-empty array `[T; N]` should reasonably be expressed as `(T, [T; N-1])`.

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// CFU device
+    cfu_device: CfuDevice,
+    /// Component ID for each individual device
+    devices: &'a [ComponentId],
+    /// Customization for the Splitter
+    customization: C,
+}
+
+/// Maximum number of devices supported
+pub const MAX_SUPPORTED_DEVICES: usize = 8;
```
felipebalbi, line 38: Is this arbitrary limit or spec limitation?

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,205 @@
...
+    async fn process_get_fw_version(&self) -> InternalResponseData {
+        let mut versions = [GetFwVersionResponse {
+            header: Default::default(),
+            component_info: Default::default(),
+        }; MAX_SUPPORTED_DEVICES];
+
+        if self.devices.len() > MAX_SUPPORTED_DEVICES {
+            error!("More devices than supported");
+            return self.create_invalid_fw_version_response();
+        }
```
felipebalbi, line 87: How about checking for max supported devices before allocating stack space for all devices?

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,205 @@
...
+        for (device_id, version) in zip(self.devices.iter(), versions.iter_mut()) {
+            if let Ok(InternalResponseData::FwVersionResponse(version_info)) =
+                cfu::route_request(*device_id, RequestData::FwVersionRequest).await
+            {
+                *version = version_info;
+            } else {
+                error!("Failed to get FW version for device {}", device_id);
+                return self.create_invalid_fw_version_response();
+            }
+        }
```
felipebalbi, line 98: This `zip()` is mostly repeated in several places. How about adding a helper method that takes a closure for the loop body and two iterators for the `zip` call? Something like (won't compile)a;

```rust
fn helper(a: A<T>, b: B<U>, body: FnOnce....) -> Result<...> {
    for (foo, bar) in zip(a, b) {
        body(...)
    }

    Ok(...)
}

File: cfu-service/src/splitter.rs (hunk 1 above)
Comment (RobertZ2011) [reply cfu-service/src/splitter.rs]: In this case the size of the array is fixed to 7 by the CFU spec.

File: cfu-service/src/splitter.rs (hunk 3 above)
Comment (RobertZ2011) [reply cfu-service/src/splitter.rs:38]: It's an arbitrary limitation instead of using a `const N: usize` generic argument.

File: cfu-service/src/splitter.rs [hunk 6]
```diff
@@ -0,0 +1,205 @@
...
+
+        let mut overall_version = self.customization.resolve_fw_versions(&versions[..self.devices.len()]);
+        // The overall component version comes first
+        overall_version.component_info[0].component_id = self.cfu_device.component_id();
+        InternalResponseData::FwVersionResponse(overall_version)
+    }
+
+    /// Process a give offer request
+    async fn process_give_offer(&self, offer: &FwUpdateOffer) -> InternalResponseData {
+        if self.devices.len() > MAX_SUPPORTED_DEVICES {
```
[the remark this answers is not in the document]
RobertZ2011: Good idea, done.

asasine (dismissed):
Non-blocking but could tighten up the API to make design decisions more strict.

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,251 @@
...
+    fn create_content_rejection(sequence: u16) -> InternalResponseData {
+        InternalResponseData::ContentResponse(FwUpdateContentResponse::new(
+            sequence,
+            CfuUpdateContentResponseStatus::ErrorInvalid,
+        ))
+    }
+
+    /// Map items in an input slice to an output slice using an async closure.
+    ///
+    /// This function will execute the closure concurrently in groups up to 4 items at a time.
```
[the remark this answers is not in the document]
asasine: Based on how this utility is implemented, 4 seems like an arbitrary decision to balance needing to implement 12 arms on the match and limit the stack size to 12x the size of `(&I, &mut O)`

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,251 @@
...
+    async fn map_slice_join<'i, 'o, I, O, F: Future<Output = bool>>(
```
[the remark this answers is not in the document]
RobertZ2011: Done.

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,251 @@
...
+                        .await
+                        .into_iter()
+                        .all(|x| x)
+                    {
+                        // Only exit early if requested, there could be more items to process
+                        return false;
+                    }
+                }
+                _ => {
+                    // Other combinations aren't possible because there will never be a Some after a None
```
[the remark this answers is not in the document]
RobertZ2011: Thanks, updated the comment.

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,251 @@
...
+        ))
+    }
+
+    /// Map items in an input slice to an output slice using an async closure.
+    ///
+    /// This function will execute the closure concurrently in groups up to 4 items at a time.
+    async fn map_slice_join<'i, 'o, I, O, F: Future<Output = bool>>(
+        input: &'i [I],
+        output: &'o mut [O],
+        f: impl Fn(&'i I, &'o mut O) -> F,
```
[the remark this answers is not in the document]
RobertZ2011: Done.

cfu-service/src/splitter.rs
```diff
@@ -0,0 +1,251 @@
...
+        self,
+        component::{CfuDevice, InternalResponseData, RequestData},
+    },
+    error, intrusive_list, trace,
+};
+
+/// Trait containing customization functionality for [`Splitter`]
+pub trait Customization {
+    /// Decides which firmware version to use based on the provided versions from all devices.
+    fn resolve_fw_versions(&self, versions: &[GetFwVersionResponse]) -> GetFwVersionResponse;
```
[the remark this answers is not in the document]
RobertZ2011: I don't want to restrict the customization code to just the values it's given. There are situations where I can imagine the implementation may want to combine versions or introduce vendor-specific values and flags.

asasine (approved):

felipebalbi (approved):

PR #349, opened 2025-06-12
Power Policy Charger Updates

tullom:
Changelog:
- added Powered and Unpowered states
    - **Breaking Change** this change is made because an unpowered state is a valid state the charger can be in, specifically in dead battery scenarios where the charger power rail might not be turned on. Power-policy's `connect_new_consumer()` relies on being able to communicate with the charger but the charger needs to be powered, which doesn't happen until later in the function.
- added Check Ready command to transition from Unpowered to Powered state and vice-versa
    - **Breaking Change**
- Renamed ChargeController::BusError to ChargeControllerError and added a trait bound to impl From for charger::ChargerError
    - **Breaking change** but I believe this is needed since, with the more constrained need to convert the custom error type to a ChargerError, `BusError` is too limiting of a name. Receiving an error from the `ChargeController::init_charger()` trait method could be due to being in an invalid state rather than a bus error.
- Changed connect_new_consumer() to check for the charger being unpowered, and to skip the charger's PSU detach sequence if true.

### DIFF
--- a/embedded-service/src/power/policy/charger.rs
+++ b/embedded-service/src/power/policy/charger.rs
@@ -10,19 +10,33 @@ use super::PowerCapability;
 /// Charger controller trait that device drivers may use to integrate with internal messaging system
 pub trait ChargeController: embedded_batteries_async::charger::Charger {
     /// Type of error returned by the bus
-    type BusError;
+    type ChargeControllerError;
 
     /// Returns with pending events
     fn wait_event(&mut self) -> impl Future<Output = ChargerEvent>;
     /// Initialize charger hardware, after this returns the charger should be ready to charge
-    fn init_charger(&mut self) -> impl Future<Output = Result<(), Self::BusError>>;
+    fn init_charger(&mut self) -> impl Future<Output = Result<(), Self::ChargeControllerError>>;
     /// Returns if the charger hardware detects if a PSU is attached
-    fn is_psu_attached(&mut self) -> impl Future<Output = Result<bool, Self::BusError>>;
+    fn is_psu_attached(&mut self) -> impl Future<Output = Result<bool, Self::ChargeControllerError>>;
     /// Called after power policy attaches to a power port.
-    fn attach_handler(&mut self, capability: PowerCapability) -> impl Future<Output = Result<(), Self::BusError>>;
+    fn attach_handler(
+        &mut self,
+        capability: PowerCapability,
+    ) -> impl Future<Output = Result<(), Self::ChargeControllerError>>;
     /// Called after power policy detaches from a power port, either to switch consumers,
     /// or because PSU was disconnected.
-    fn detach_handler(&mut self) -> impl Future<Output = Result<(), Self::BusError>>;
+    fn detach_handler(&mut self) -> impl Future<Output = Result<(), Self::ChargeControllerError>>;
+    /// Called when a charger CheckReady request (PolicyEvent::CheckReady) is sent to the power policy.
+    /// Upon successful return of this method, the charger is assumed to be powered and ready to communicate,
+    /// transitioning state from unpowered to powered.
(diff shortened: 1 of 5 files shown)


power-policy-service/src/consumer.rs
```diff
@@ -112,11 +123,20 @@ impl PowerPolicy {
...
+            // as we do not wish to block the rest of the connect sequence from completing.
             for node in self.context.chargers().await {
                 let device = node.data::<ChargerDevice>().ok_or(Error::InvalidDevice)?;
-                device
+                if let Err(e) = device
                     .execute_command(PolicyEvent::PolicyConfiguration(new_consumer.power_capability))
-                    .await?;
+                    .await
+                {
+                    error!("Error executing charger attach sequence. Continuing connect_new_consumer()...");
```
RobertZ2011, line 136: Can you combine the `error!` and `debug!` logs?

power-policy-service/src/consumer.rs
```diff
@@ -87,14 +90,22 @@ impl PowerPolicy {
...
-                device
+                if let Err(e) = device
                     .execute_command(PolicyEvent::PolicyConfiguration(PowerCapability {
                         voltage_mv: 0,
                         current_ma: 0,
                     }))
-                    .await?;
+                    .await
+                {
+                    error!("Error executing charger detach sequence. Continuing connect_new_consumer()...");
```
RobertZ2011, line 105: Can you combine the error! and debug! logs?

power-policy-service/src/consumer.rs
```diff
@@ -87,14 +90,22 @@ impl PowerPolicy {
                 consumer.disconnect().await?;
             }
 
+            // If no chargers are registered, they won't receive the new power capability.
+            // Also, if chargers execute_command errors out, continue execution of connect_new_consumer()
```
RobertZ2011, line 94: Does this flow match Sean's design doc?

RobertZ2011 (changes requested):

embedded-service/src/power/policy/charger.rs
```diff
@@ -109,13 +126,22 @@ pub type ChargerResponse = Result<ChargerResponseData, ChargerError>;
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum PoweredSubstate {
     /// Device is initializing
     Init,
     /// PSU is attached and device can charge if desired
     PsuAttached,
     /// PSU is detached
     PsuDetached,
-    // TODO: Dead battery revival?
 }
```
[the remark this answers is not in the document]
tullom: Updated the description. These changes were made as a direct request from our partners.

power-policy-service/src/consumer.rs
tullom: Removed them both, changed up the logic a bit to explicitly check for an unpowered charger, and to just `debug!()`.

power-policy-service/src/consumer.rs
tullom: Removed them both, changed up the logic a bit to explicitly check for an unpowered charger, and to just `debug!()`.

RobertZ2011 (approved):

embedded-service/src/power/policy/charger.rs
```diff
@@ -10,19 +10,33 @@ use super::PowerCapability;
 /// Charger controller trait that device drivers may use to integrate with internal messaging system
 pub trait ChargeController: embedded_batteries_async::charger::Charger {
     /// Type of error returned by the bus
-    type BusError;
+    type ChargeControllerError;
```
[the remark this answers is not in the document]
tullom: Makes sense, its a bit more unneeded verbosity. But i'll leave it in for now.

PR #357, opened 2025-06-17
cfu-service: Add CFU buffer

RobertZ2011:


cfu-service/src/buffer.rs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        if state.component_busy {
+            // Buffer the content if the component is busy
+            // If the buffer is full, this will block until space is available
+            trace!("Component is busy, buffering content");
+            self.buffer_sender.send(*content).await;
+        } else {
+            // Buffered component can accept new content, send it
+            if let Err(e) = cfu::send_device_request(self.buffered_id, RequestData::GiveContent(*content)).await {
```
madeleyneVaca, line 156: Why do we not need to wait for a response here? `Process_give_content` is called out further down on line 259 of this file as a place to catch the response.

RobertZ2011: There are two benefits this approach. The first is that it makes it easier to detect bugs. Ideally nothing else should be sending CFU commands to the component. If this blocks then it can result in a legitimate timeout. If the calls were combined then the timeout is expected and might hide other issues. This also makes the code slightly more responsive since we don't have to wait for the full timeout period if the component finishes a previous command. With the PD controller it takes ~10 ms to write a payload so even a small speedup can help the buffer clear out very quickly.

madeleyneVaca (approved):

PR #362, opened 2025-06-19
Fix clippy workflow

kurtjd:
Updates the GitHub workflow to ensure clippy warnings cause the workflow to fail.

The explicit `clippy` job (which was nonfunctional and relied on `giraffate/clippy-action`) has been removed since we weren't using reviewdog to post reviews and thus had limited use. Instead, `check` in `cargo hack` was replaced with `clippy` to ensure we check all feature combinations. Similarly, examples were changed to use `clippy` instead of `check` as well.

Lastly, the `test` job also includes checking for clippy warnings on tests.

The repo settings will also likely need to be changed as it seems to be stuck waiting on jobs that no longer exist.

Depends on #360 getting merged (which I'll then rebase on) in order for the new checks to pass.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,10 +1,14 @@
 # This workflow runs whenever a PR is opened or updated, or a commit is pushed to main. It runs
 # several checks:
 # - fmt: checks that the code is formatted according to rustfmt
-# - clippy: checks that the code does not contain any clippy warnings
+# - semver: checks that the crate is following proper semantic versioning
 # - doc: checks that the code can be documented without errors
-# - hack: check combinations of feature flags
+# - hack: check combinations of feature flags for errors (including clippy warnings)
+# - deny: checks licenses, advisories, sources, and bans for our dependencies
+# - test: checks that unit tests pass and do not contain clippy warnings
 # - msrv: check that the msrv specified in the crate is correct
+# - clippy-arm-examples: checks that ARM examples contain no clippy warnings
+# - clippy-std-examples: checks that std examples contain no clippy warnings
 permissions:
   contents: read
 # This configuration allows maintainers of this repo to create a branch and pull request based on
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -184,42 +194,68 @@ jobs:
...
+      - name: cargo clippy
+        uses: giraffate/clippy-action@v1
+        with:
+          workdir: ${{ matrix.example_directory }}
+          reporter: 'github-pr-check'
+          reviewdog_flags: '-fail-level=any'
+          clippy_flags: --target thumbv8m.main-none-eabihf --locked --features ${{ matrix.features }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
 
-  check-std-examples:
```
RobertZ2011, line 207: Are we still running check on the std examples with this change? Likewise on the ARM examples?

kurtjd: This changes to only run `cargo clippy`. Are there things that `cargo check` can catch that `cargo clippy` cannot? If so I can have clippy run in addition to check.

RobertZ2011 (approved):

kurtjd:
I've modified it a bit to the `hack` job has been renamed to `hack-clippy` because it can double has checking for feature unification and clippy warnings on all combinations of features. Making these separate jobs would duplicate YAML code and work. The job has been renamed to `stable / clippy` and `beta / clippy` to match repository settings that require jobs with those names to pass before merging is allowed.

@jerrysxie

.github/workflows/check.yml
```diff
@@ -36,32 +40,6 @@ jobs:
...
-          submodules: true
-      - name: Install ${{ matrix.toolchain }}
-        uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: ${{ matrix.toolchain }}
-          components: clippy
-      - name: cargo clippy
-        uses: giraffate/clippy-action@v1
-        with:
-          reporter: 'github-pr-check'
```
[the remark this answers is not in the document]
kurtjd: Right, but I don't see a way for `giraffate/clippy-action` to analyze all combinations of features like `cargo hack` does. What I meant was since we didn't seem to care about the reviewdog part, it might make sense to just use `cargo hack` to avoid duplicating work.

But if we do want automatic reviews I can try to come up with a different solution (or might just have to accept that we can only get clippy lints on one or two features).

PR #365, opened 2025-06-20
Add SyncCell type to support Sync fix in embassy upstream

### LINKED ISSUE #361: Failure to compile with latest `embassy-sync`
After https://github.com/embassy-rs/embassy/pull/4320 merged, several bugs in ODP were exposed around improper `Sync` bounds. One of the major errors is `IntrusiveList` being used everywhere in `OnceLock`'s when `IntrusiveList` is `!Sync`

```Rust
error[E0277]: `core::cell::Cell<core::option::Option<&'static intrusive_list::IntrusiveNode>>` cannot be shared between threads safely         
   --> D:\.tools\.cargo\git\checkouts\embedded-services-78faf442f1f643c2\a938ac1\embedded-service\src\activity.rs:135:21
    |
135 | static SUBSCRIBERS: OnceLock<IntrusiveList> = OnceLock::new();
    |                     ^^^^^^^^^^^^^^^^^^^^^^^ `core::cell::Cell<core::option::Option<&'static intrusive_list::IntrusiveNode>>` cannot be shared between threads safely
    |
    = help: within `intrusive_list::IntrusiveList`, the trait `Sync` is not implemented for `core::cell::Cell<core::option::Option<&'static intrusive_list::IntrusiveNode>>`
    = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::RwLock`
note: required because it appears within the type `intrusive_list::IntrusiveList`
   --> D:\.tools\.cargo\git\checkouts\embedded-services-78faf442f1f643c2\a938ac1\embedded-service\src\intrusive_list.rs:62:12
    |
62  | pub struct IntrusiveList {
    |            ^^^^^^^^^^^^^
    = note: required for `OnceLock<intrusive_list::IntrusiveList>` to implement `Sync`
    = note: shared static variables must have a type that implements `Sync`
```

JamesHuard:
Embassy upstream recently fixed accidentally adding Sync to non-Sync types: https://github.com/embassy-rs/embassy/pull/4320

In preparation to fix #361, added a drop-in replacement for Cell where moving to a proper primitive usage such as embassy_sync::mutex::Mutex is untenable. This also refactors the intrusive_list to utilize this new SyncCell type, allowing it to be used directly without any containers.

Note that in most cases SyncCell is not the best solution. If it is possible, proper synchronization primitives should be leveraged.

A critical section was used as opposed to a blocking_mutex type, as blocking_mutex is non-re-entrant. This allows the most flexibility in supporting scenarios with SyncCell, and does not require permutated get and set implementations based on mutex backing type.

In short:
- New SyncCell type introduced to unblock ingesting embassy upstream Sync fix
- We should use it sparingly
- IntrusiveList was updated to use this new type, allowing it to be allocated statically without any errors

### DIFF
--- a/embedded-service/src/sync_cell.rs
+++ b/embedded-service/src/sync_cell.rs
@@ -1,7 +1,7 @@
-//! # SyncCell: a cell-like API for static interior mutability scenarios
+//! # SyncCell: a cell-like API for static interior mutability scenarios. Backed by a critical section, implying it's usage may delay or defer interrupts. Recommended to use sparingly.
 use core::cell::Cell;
 
-/// A critical section backed Cell for sync scenarios where you want Cell behaviors, but need it to be thread safe (such as used in statics)
+/// A critical section backed Cell for sync scenarios where you want Cell behaviors, but need it to be thread safe (such as used in statics). Backed by a critical section, use sparingly.
 pub struct SyncCell<T: ?Sized> {
     inner: Cell<T>,
 }
@@ -23,12 +23,15 @@ impl<T> SyncCell<T> {
         critical_section::with(|_cs| self.inner.set(value))
     }
 
-    /// Swap contents between two SyncCell's
+    /// Swap contents between two SyncCell's    
+    /// # Panics
+    ///
+    /// This function will panic if `self` and `other` are different `Cell`s that partially overlap.
+    /// (Using just standard library methods, it is impossible to create such partially overlapping `Cell`s.
+    /// However, unsafe code is allowed to e.g. create two `&Cell<[i32; 2]>` that partially overlap.)
     pub fn swap(&self, other: &Self) {
         critical_section::with(|_cs| self.inner.swap(&other.inner));
     }
-
-    // we do not implement replace() as we cannot guarantee that it will not be called outside of compile time
 
     /// consume the Synccell and return the inner value T
     pub fn into_inner(self) -> T {
@@ -66,7 +69,7 @@ impl<T: Default> SyncCell<T> {
 unsafe impl<T> Sync for SyncCell<T> {}
 
 // SAFETY: Can implement send here due to critical section without T being explicitly Send
-unsafe impl<T> Send for SyncCell<T> where T:Send {}
+unsafe impl<T> Send for SyncCell<T> where T: Send {}
 
 impl<T: Copy> Clone for SyncCell<T> {
     #[inline]
--- a/embedded-service/src/sync_cell.rs
+++ b/embedded-service/src/sync_cell.rs
@@ -1,26 +1,17 @@
 //! # SyncCell: a cell-like API for static interior mutability scenarios
-use core::cell::UnsafeCell;
+use core::cell::Cell;
 
 /// A critical section backed Cell for sync scenarios where you want Cell behaviors, but need it to be thread safe (such as used in statics)
-pub struct SyncCell<T: Copy> {
-    inner: UnsafeCell<T>,
+pub struct SyncCell<T: ?Sized> {
+    inner: Cell<T>,
 }
 
-impl<T: Copy> SyncCell<T> {
+impl<T> SyncCell<T> {
     /// Constructs a SyncCell, initializing it with initial_value
     pub const fn new(initial_value: T) -> Self {
         Self {
-            inner: UnsafeCell::new(initial_value),
+            inner: Cell::new(initial_value),
         }
-    }
-
-    /// Reads the cell's content (in a critical section) and returns a copy
-    pub fn get(&self) -> T {
-        critical_section::with(|_cs|
-            // SAFETY: safe as accessors (get/set) are always completed in a critical section
-            unsafe {
-            *self.inner.get()
-        })
     }
 
     /// Sets the cell's content in a critical section. Note that this accounts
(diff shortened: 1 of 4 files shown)


embedded-service/src/sync_cell.rs
```diff
@@ -0,0 +1,114 @@
+//! # SyncCell: a cell-like API for static interior mutability scenarios
+use core::cell::UnsafeCell;
+
+/// A critical section backed Cell for sync scenarios where you want Cell behaviors, but need it to be thread safe (such as used in statics)
+pub struct SyncCell<T: Copy> {
+    inner: UnsafeCell<T>,
```
asasine, line 6: I think wrapping `Cell<T>` would be fine, and would eliminate all `unsafe` in the `SyncCell` implementation.

embedded-service/src/sync_cell.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl<T: Copy> SyncCell<T> {
```
asasine, line 9: I recommend implementing as much of the `Cell<T>` API as possible, including traits conditionally like `Clone`, `Default`, `Ord`, `PartialEq`, `Display`, etc.

gjpmsft (approved):

embedded-service/src/sync_cell.rs
JamesHuard: I wasn't sure if using Cell or UnsafeCell would be more "taboo" -- given I've seen SyncUnsafeCell in the unstable libs. I can definitely move to Cell if impl Sync over it is "as acceptable"

embedded-service/src/sync_cell.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// # Safety
+    /// This may be used safely if and only if the pointer is held during a critical section, or
+    /// all accessors to this Cell are blocked until the pointer is released.
+    pub unsafe fn as_ptr(&self) -> *mut T {
+        self.inner.get()
+    }
+}
+
+// SAFETY: Sync is implemented here for SyncCell as T is only accessed via nestable critical sections
+unsafe impl<T: Copy> Sync for SyncCell<T> {}
```
wmmc88, line 49: You should also be able to `impl  Send` safely here (given that `T:Send`). [Cell does this too](https://doc.rust-lang.org/src/core/cell.rs.html#316)

wmmc88, line 49: This doesn't necessarily need a `Sized` bound

embedded-service/src/sync_cell.rs
```diff
@@ -0,0 +1,114 @@
...
+            *self.inner.get()
+        })
+    }
+
+    /// Sets the cell's content in a critical section. Note that this accounts
+    /// for read/write conditions but does not automatically handle logical data
+    /// race conditions. It is still possible for a user to read a value but have
+    /// it change after they've performed the read. This just ensures data integrity:
+    /// SyncCell<T> will always contain a valid T, even if it's been read "late"
+    pub fn set(&self, value: T) {
```
wmmc88, line 31: `set` along with a lot of the other `Cell` apis(ex. `update`, `swap`) don't required `T: Copy`. Some apis don't even require `T:Sized` (although im not sure how useful that would even be here).

embedded-service/src/sync_cell.rs
wmmc88: actually maybe its fine to not relax to `T: ?Sized` since I can't think of a reason you'd use that in the context of the apis here (being wrapped with a crit section)

embedded-service/src/sync_cell.rs
JamesHuard: implemented

embedded-service/src/sync_cell.rs
JamesHuard: Moved around

embedded-service/src/sync_cell.rs
```diff
@@ -0,0 +1,199 @@
...
+
+impl<T: Copy> SyncCell<T> {
+    /// Reads the cell's content (in a critical section) and returns a copy
+    pub fn get(&self) -> T {
+        critical_section::with(|_cs| self.inner.get())
+    }
+}
+
+impl<T: ?Sized> SyncCell<T> {
+    /// Unsafe: allows reads and writes without critical section guard, violating Sync guarantees.
```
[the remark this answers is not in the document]
JamesHuard: Need it in intrusive_list

embedded-service/src/sync_cell.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub fn set(&self, value: T) {
+        critical_section::with(|_cs| self.inner.set(value))
+    }
+
+    /// Swap contents between two SyncCell's
+    pub fn swap(&self, other: &Self) {
+        critical_section::with(|_cs| self.inner.swap(&other.inner));
+    }
+
+    // we do not implement replace() as we cannot guarantee that it will not be called outside of compile time
```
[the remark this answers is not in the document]
JamesHuard: Another option might be to make it const but also unsafe?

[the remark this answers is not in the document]
JamesHuard: image

Oh, I might just drop this note all together since this is an unstable feature.

embedded-service/src/sync_cell.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Sets the cell's content in a critical section. Note that this accounts
+    /// for read/write conditions but does not automatically handle logical data
+    /// race conditions. It is still possible for a user to read a value but have
+    /// it change after they've performed the read. This just ensures data integrity:
+    /// SyncCell<T> will always contain a valid T, even if it's been read "late"
+    pub fn set(&self, value: T) {
+        critical_section::with(|_cs| self.inner.set(value))
+    }
+
+    /// Swap contents between two SyncCell's
```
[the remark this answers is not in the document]
JamesHuard: called out in two places

embedded-service/src/sync_cell.rs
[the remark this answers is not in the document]
JamesHuard: removed note entirely

embedded-service/src/sync_cell.rs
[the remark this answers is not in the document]
JamesHuard: Changed my mind after thinking on it more. If someone just wants to print the backing address, that is safe. So I will mark this safe but keep the safety section in the doc

embedded-service/src/sync_cell.rs
asasine: Is `dyn Trait` sized? You could use this to swap between different trait impls.

JamesHuard: `&dyn Trait` is sized as one of those wide/fat pointer types. I don't think `dyn Trait` is inherently Sized

gjpmsft (approved):

PR #367, opened 2025-06-23
Battery service: add context config, async state getter, and remove autonomous state machine logic

tullom:
Changelog:
- add config to battery service context, allowing OEMs to tweak the inner workings of the battery service. More config features are planned as the state machine matures and locks down
- Add the ability to asynchronously query the state of the battery service
    - as a result, state is now a `Mutex` instead of a `RefCell`
- Remove state machine autonomously progressing through the states, meaning now the state machine needs to be explicitly sent commands to drive the state machine from one state to the next.

### DIFF
--- a/battery-service/src/context.rs
+++ b/battery-service/src/context.rs
@@ -1,11 +1,12 @@
 use crate::device::Device;
 use crate::device::{self, DeviceId};
 use embassy_sync::channel::Channel;
+use embassy_sync::mutex::Mutex;
 use embassy_sync::{blocking_mutex::raw::NoopRawMutex, channel::TrySendError};
 use embassy_time::{with_timeout, Duration};
 use embedded_services::{debug, error, info, intrusive_list, trace, warn, IntrusiveList};
 
-use core::cell::RefCell;
+use core::cell::Cell;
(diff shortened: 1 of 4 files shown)


battery-service/src/context.rs
```diff
@@ -188,83 +231,114 @@ impl Context {
     /// Main battery service state machine
     #[allow(clippy::await_holding_refcell_ref)]
```
kurtjd, line 232: Can this be removed now?

tullom: good catch, i thought i had removed this!

battery-service/src/context.rs
```diff
@@ -186,85 +229,115 @@ impl Context {
...
             Err(err) => return Err(err),
         }
 
-        loop {
-            let mut continue_exec = false;
-            match *state {
-                State::NotPresent => {
-                    info!("Initializing fuel gauge with ID {:?}", event.device_id);
-                    if self
+        match *state {
```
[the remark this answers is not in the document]
tullom: Done!

jerrysxie (approved):

PR #371, opened 2025-06-24
type-c-service: Make event processing more public

RobertZ2011:
Refactor to expose events coming from the type-C service. External code can use a combination of `wait_event` and `process_event` to inject its own logic into the event loop. Port events are now produced as a stream instead of a combined blob of events.

### DIFF
--- a/type-c-service/src/task.rs
+++ b/type-c-service/src/task.rs
@@ -1,13 +1,14 @@
-use core::cell::RefCell;
+use core::{cell::RefCell, future::Future};
 use embassy_futures::select::{select, Either};
-use embassy_sync::once_lock::OnceLock;
+use embassy_sync::{blocking_mutex::raw::NoopRawMutex, once_lock::OnceLock};
 use embedded_services::{
     comms::{self, EndpointID, Internal},
-    debug, error, info,
+    debug, error, info, intrusive_list,
+    ipc::deferred,
     type_c::{
         self,
         controller::PortStatus,
-        event::PortEventFlags,
+        event::{PortEventFlagsIter, PortEventKind},
         external::{self, ControllerCommandData},
         ControllerId,
     },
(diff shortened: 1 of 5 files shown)


asasine (approved):

type-c-service/src/task.rs
```diff
@@ -220,21 +217,69 @@ impl Service {
...
-            Either::First(pending) => self.process_unhandled_events(pending).await,
-            Either::Second(request) => {
+    /// Wait for port flags
+    #[allow(clippy::await_holding_refcell_ref)]
+    async fn wait_port_flags(&self) -> PortEventFlagsIter {
+        let mut state = self.state.borrow_mut();
+        if state.event_iter.is_some() {
+            // If we have an existing iterator, return it
+            // Yield first to prevent starving other tasks
+            embassy_futures::yield_now().await;
```
jerrysxie, line 227: @asasine @RobertZ2011 If I am understanding this correctly, this yield between handling each event so OEM can insert custom logic between each port event. However, this `yield_now()` is a bit suspect. Because it will not allow the CPU to sleep. We might actually want to wait some reasonable time here instead of `yield_now()`.

```
/// Yield from the current task once, allowing other tasks to run.
///
/// This can be used to easily and quickly implement simple async primitives
/// without using wakers. The following snippet will wait for a condition to
/// hold, while still allowing other tasks to run concurrently (not monopolizing
/// the executor thread).
///
/// ```rust
/// # use embassy_futures::{block_on, yield_now};
/// # async fn test_fn() {
/// # let mut iter_count: u32 = 0;
/// # let mut some_condition = || { iter_count += 1; iter_count > 10 };
/// while !some_condition() {
///     yield_now().await;
/// }
/// # }
/// # block_on(test_fn());
/// ```
///
/// The downside is this will spin in a busy loop, using 100% of the CPU, while
/// using wakers correctly would allow the CPU to sleep while waiting.
///
/// The internal implementation is: on first poll the future wakes itself and
/// returns `Poll::Pending`. On second poll, it returns `Poll::Ready`.
pub fn yield_now() -> impl Future<Output = ()> {
    YieldNowFuture { yielded: false }
}
```

jerrysxie (changes requested):

type-c-service/src/task.rs
jerrysxie: I see so the expectation is for it to wait for a signal when there is no event. But if there is an event, then after handling each event, it would yield to unblock other task and then handle the next event until all events are handled. Then it goes back to waiting for a signal that there was an event.

jerrysxie (approved):

PR #372, opened 2025-06-24
Update dependencies and replace Cell and RefCell with SyncCell and Mutex, respectively

tullom:
- NoopRawMutex changed to CriticalSectionRawMutex
- embedded_service::Buffer fixed to use `AtomicPtr<T>` rather than `*mut [T]`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,3 +66,8 @@ toml = { version = "0.8", default-features = false }
 syn = "2.0"
 tps6699x = { git = "https://github.com/OpenDevicePartnership/tps6699x" }
 tokio = { version = "1.42.0" }
+
+# Patch used as a workaround for a circular dependency in espi-service from embassy-imxrt, which depends on
+# `storage-bus` which depends on embedded-services
+[patch."https://github.com/OpenDevicePartnership/embedded-services"]
+embedded-services = { path = "./embedded-service" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,8 +66,3 @@ toml = { version = "0.8", default-features = false }
 syn = "2.0"
 tps6699x = { git = "https://github.com/OpenDevicePartnership/tps6699x" }
 tokio = { version = "1.42.0" }
-
-# Patch used as a workaround for a circular dependency in espi-service from embassy-imxrt, which depends on
-# `storage-bus` which depends on embedded-services
-[patch."https://github.com/OpenDevicePartnership/embedded-services"]
-embedded-services = { path = "./embedded-service" }
--- a/battery-service/src/context.rs
+++ b/battery-service/src/context.rs
@@ -1,13 +1,14 @@
 use crate::device::Device;
 use crate::device::{self, DeviceId};
 use embassy_sync::channel::Channel;
+use embassy_sync::channel::TrySendError;
 use embassy_sync::mutex::Mutex;
-use embassy_sync::{blocking_mutex::raw::NoopRawMutex, channel::TrySendError};
 use embassy_time::{with_timeout, Duration};
+use embedded_services::GlobalRawMutex;
 use embedded_services::{debug, error, info, intrusive_list, trace, warn, IntrusiveList};
 
-use core::cell::Cell;
 use core::ops::DerefMut;
+use core::sync::atomic::AtomicUsize;
 
 /// Battery service states.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -104,10 +105,10 @@ pub struct BatteryEvent {
 /// Battery service context, hardware agnostic state.
 pub struct Context {
     fuel_gauges: IntrusiveList,
-    state: Mutex<NoopRawMutex, State>,
-    battery_event: Channel<NoopRawMutex, BatteryEvent, 1>,
-    battery_response: Channel<NoopRawMutex, BatteryResponse, 1>,
-    no_op_retry_count: Cell<usize>,
+    state: Mutex<GlobalRawMutex, State>,
+    battery_event: Channel<GlobalRawMutex, BatteryEvent, 1>,
+    battery_response: Channel<GlobalRawMutex, BatteryResponse, 1>,
+    no_op_retry_count: AtomicUsize,
     config: Config,
 }
 
@@ -133,7 +134,7 @@ impl Context {
             state: Mutex::new(State::NotPresent),
             battery_event: Channel::new(),
             battery_response: Channel::new(),
-            no_op_retry_count: Cell::new(0),
+            no_op_retry_count: AtomicUsize::new(0),
             config: Default::default(),
         }
     }
@@ -144,7 +145,7 @@ impl Context {
             state: Mutex::new(State::NotPresent),
             battery_event: Channel::new(),
             battery_response: Channel::new(),
-            no_op_retry_count: Cell::new(0),
+            no_op_retry_count: AtomicUsize::new(0),
             config,
         }
     }
@@ -161,12 +162,13 @@ impl Context {
 
     /// Get global state machine NotOperational retry count.
     fn get_state_machine_retry_count(&self) -> usize {
-        self.no_op_retry_count.get()
+        self.no_op_retry_count.load(core::sync::atomic::Ordering::Relaxed)
     }
 
     /// Set global state machine NotOperational retry count.
     fn set_state_machine_retry_count(&self, retry_count: usize) {
-        self.no_op_retry_count.set(retry_count)
+        self.no_op_retry_count
+            .store(retry_count, core::sync::atomic::Ordering::Relaxed)
     }
 
     /// Main processing function.
--- a/battery-service/src/context.rs
+++ b/battery-service/src/context.rs
@@ -187,7 +187,7 @@ impl Context {
             },
             Err(_) => {
                 error!("Battery state machine timeout!");
-                // Should be infalliable
+                // Should be infallible
                 self.do_state_machine(BatteryEvent {
                     event: BatteryEventInner::Timeout,
                     device_id: event.device_id,
--- a/embedded-service/src/buffer.rs
+++ b/embedded-service/src/buffer.rs
@@ -12,9 +12,11 @@
 //! This allows for producer code to own the buffer through a `OwnedRef`, and then allow access to consumers
 //! through any number of `SharedRef`.
 use core::borrow::{Borrow, BorrowMut};
-use core::cell::Cell;
 use core::marker::PhantomData;
 use core::ops::Range;
+
+use crate::sync_cell::SyncCell;
+use core::sync::atomic::AtomicPtr;
 
 #[derive(Copy, Clone, PartialEq, Eq)]
 enum Status {
@@ -25,8 +27,9 @@ enum Status {
 
 /// Underlying buffer storage struct
 pub struct Buffer<'a, T> {
-    buffer: *mut [T],
-    status: Cell<Status>,
+    buffer: AtomicPtr<T>,
+    len: usize,
+    status: SyncCell<Status>,
     _lifetime: PhantomData<&'a ()>,
 }
 
@@ -36,8 +39,9 @@ impl<'a, T> Buffer<'a, T> {
     /// No other code should have access to the buffer
     pub unsafe fn new(raw_buffer: &'a mut [T]) -> Self {
         Buffer {
-            buffer: raw_buffer,
-            status: Cell::new(Status::None),
+            buffer: AtomicPtr::new(raw_buffer.as_mut_ptr()),
+            len: raw_buffer.len(),
+            status: SyncCell::new(Status::None),
             _lifetime: PhantomData,
         }
     }
@@ -52,13 +56,13 @@ impl<'a, T> Buffer<'a, T> {
     /// Returns the length of the buffer
     // SAFETY: The buffer is always valid
     pub fn len(&self) -> usize {
-        unsafe { self.buffer.as_mut().unwrap().len() }
+        self.len
     }
 
     /// Returns `true`` if the buffer has a length of 0
     // SAFETY: The buffer is always valid
     pub fn is_empty(&self) -> bool {
-        unsafe { self.buffer.as_mut().unwrap().is_empty() }
+        self.len == 0
     }
 
     fn borrow(&self, mutable: bool) {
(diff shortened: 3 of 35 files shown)


tullom:
Looks like the CI failures are due to a circular dependency between `embassy-imxrt` and `embedded-services`

battery-service/src/context.rs
```diff
@@ -94,10 +96,10 @@ pub struct BatteryEvent {
 /// Battery service context, hardware agnostic state.
 pub struct Context {
     fuel_gauges: IntrusiveList,
-    state: Mutex<NoopRawMutex, State>,
```
jerrysxie, line 97: Change these to `CriticalSectionRawMutex` will have performance impact since it will disable interrupts. `ThreadModeRawMutex` might be a better choice.

jerrysxie (changes requested):

battery-service/src/context.rs
JamesHuard: Will also flag that `ThreadModeRawMutex` is not unit test friendly, whereas `CriticalSectionRawMutex` is. What we've done is define a `GlobalRawMutex` type that swaps between these for `#[cfg(test)]` and regular builds

embedded-service/src/buffer.rs
```diff
@@ -12,10 +12,12 @@
 //! This allows for producer code to own the buffer through a `OwnedRef`, and then allow access to consumers
```
JamesHuard, line 12: Do we have any issues filed to re-examine if this + comms::route should be revisited? The buffering here makes me nervous that we might have hidden data-race conditions, and I think it's all a side effect of how comms::route is leveraged within some services. @jerrysxie I'd like to discuss with you at some point the feasibility of removing comms entirely and instead relying on "Intrusive List backed pub-sub channels" where appropriate

embedded-service/src/cfu/mod.rs
```diff
@@ -4,7 +4,7 @@ pub mod component;
 
 use core::sync::atomic::{AtomicBool, Ordering};
 
-use embassy_sync::blocking_mutex::raw::NoopRawMutex;
+use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
```
JamesHuard, line 7: echoing @jerrysxie's comment here re: `CriticalSectionRawMutex` being heavy handed with a single-executor. `ThreadMode` would be more appropriate, with the caveat that it doesn't play nicely with tokio unit tests

JamesHuard (dismissed):
Awesome work @tullom! Did we happen to test this with embassy tip of main to see if it can build now?

battery-service/src/context.rs
tullom: Done, defined a `GlobalRawMutex` type that uses `ThreadModeRawMutex` in a microcontroller context, `CriticalSectionRawMutex` when executing in a standard test environment.

embedded-service/src/buffer.rs
tullom: Out of the scope of this PR but good to bring up.

tullom:
> Awesome work @tullom! Did we happen to test this with embassy tip of main to see if it can build now?

Yes, this PR includes changes to `Cargo.lock` that updates the embassy git sha and it builds. CI is failing due to the circular dependency issue.

JamesHuard (approved):

asasine (dismissed):

asasine (dismissed):
The synchronous and blocking lock is not ideal but I don't think it's a show-stopper.

Cargo.toml
```diff
@@ -56,3 +56,6 @@ rand_core = "0.6.4"
 serde = { version = "1.0.*", default-features = false }
 tps6699x = { git = "https://github.com/OpenDevicePartnership/tps6699x" }
 tokio = { version = "1.42.0" }
+
+[patch."https://github.com/OpenDevicePartnership/embedded-services"]
+embedded-services = { path = "./embedded-service" }
```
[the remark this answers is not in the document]
tullom: Crazy how the world works sometimes, isn't it? It's a temporary workaround to fix the circular dependency we have.

embedded-service/src/comms.rs
```diff
@@ -205,7 +205,7 @@ impl Endpoint {
         Self {
             node: Node::uninit(),
             id,
-            delegator: Cell::new(None),
```
jerrysxie, line 208: Using a `SyncCell` here implies that every single time we send a message through comms to an endpoint, we need to disable interrupt. This will have some pretty serious performance inputs, so we need to rethink this a bit.

jerrysxie (changes requested):

Cargo.toml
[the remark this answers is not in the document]
kurtjd: Should be fixed now that `storage-bus` has been moved to a new repo and doesn't depend on `embedded-services`: https://github.com/OpenDevicePartnership/embedded-mcu/tree/main/storage-bus

embedded-service/src/comms.rs
jerrysxie: `'static dyn MailboxDelegate` does not implement sync so we would run afoul of the `sync` bound for `OnceLock` inner type. But realistically `delegator` is only set once and only being read afterwards. So maybe we can put `sync` bound on `Endpoint` instead of changing delegator from `Cell` into `SyncCell`.

We can put sync bound on `Node` and `IntrusiveNode` as well, because all of their fields are all set once and only being read afterwards.

jerrysxie: My prototype with `sync` bound for `comms` internal types here: https://github.com/jerrysxie/embedded-services/commit/232b738901cf807611ab5a662a6526051f99eb5e.

kurtjd: Would it not be feasible to replace this with a `ThreadModeRawMutex` like you described Jerry, but use `try_lock()` instead of `lock().await` to work in the non-async calls? I feel like there shouldn't be a case of needing to wait on the lock during normal operations so a panic should be acceptable as that means an implementation error.

Implementing `Sync` unsafely when the pattern of "fields set once then are read-only afterwards" is not actually enforced seems like it could be problematic (though I might be missing something).

jerrysxie: @kurtjd Good point!

@tullom `try_lock()` is an another option that we can do which will provide us feedbacks in terms of panics instead of undefined behavior.

jerrysxie: @tullom @kurtjd I don't think `try_lock()` will work for `intrusive_list`; we probably have to stick with `sync` bound here

```
#[derive(Copy, Clone, Debug)]
pub struct IntrusiveNode {
    /// offset from &self to struct data. Typically := sizeof(IntrusiveNode)
    address_of_data: *const dyn Any,         <--- if we replace this with mutex, we would run afoul of the copy and clone traits which is needed for the linked list implementation

    /// unsafe iterator type
    next: Option<&'static IntrusiveNode>, <--- if we replace this with mutex, we would run afoul of the copy and clone traits which is needed for the linked list implementation

    /// valid address flag: used to ensure proper initialization sequencing over address_of_data
    valid: bool,
}
```

kurtjd: Could define a `CriticalSectionCell` and a `ThreadModeCell` backed by critical section and thread mode check respectively which are in essence `Copy` and `Clone` versions of the Mutexes.

type-c-service/src/driver/tps6699x.rs
```diff
@@ -345,16 +342,14 @@ impl<const N: usize, M: RawMutex, B: I2c> Controller for Tps6699x<'_, N, M, B> {
...
-        let mut tps6699x = self.tps6699x.borrow_mut();
+        let mut tps6699x = self.tps6699x.lock().await;
         let customer_use = CustomerUse(tps6699x.get_customer_use().await?);
         Ok(customer_use.custom_fw_version())
     }
 
-    #[allow(clippy::await_holding_refcell_ref)]
     async fn start_fw_update(&mut self) -> Result<(), Error<Self::BusError>> {
-        let mut tps6699x = self.tps6699x.borrow_mut();
+        let mut tps6699x = self.tps6699x.lock().await;
```
jerrysxie, line 352: We put mutex around each field to make them sync so it compiles. However, in some function, we are grabbing multile mutexes nested. That is dangerous if we are not locking and unlocking in the right order.

embedded-service/src/comms.rs
kurtjd: I see what you mean, but I was suggesting that such a `ThreadModeCell` wouldn't be necessarily backed by a Mutex, it would just do the similar `in_thread_mode()` assertion before setting the cell just as some safety guarantee when we make it `Sync`. So no locking in this case.

embedded-service/src/keyboard.rs
```diff
@@ -60,15 +60,15 @@ pub struct BroadcastConfig {
 
 /// Keyboard service context
 struct Context {
-    broadcast_config: Cell<BroadcastConfig>,
+    broadcast_config: SyncCell<BroadcastConfig>,
```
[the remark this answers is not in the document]
tullom: Changed to `broadcast_config: Mutex<GlobalRawMutex, BroadcastConfig>` for now. Could be changed to `ThreadModeCell` in #380 if we want to shave off some overhead with locking but it should be good for now.

espi-service/src/espi_service.rs
```diff
@@ -77,7 +57,10 @@ impl Service<'_> {
 
     async fn route_to_battery_service(&self, offset: &mut usize, length: &mut usize) -> Result<(), ec_type::Error> {
         let msg = {
-            let memory_map = self.ec_memory.borrow();
+            let memory_map = self
+                .ec_memory
+                .try_lock()
+                .expect("Messages handled one after another, should be infalliable.");
```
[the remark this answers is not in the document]
tullom: I think it is more desirable for an explosive panic, rather than a silent deadlock. at least in this testing stage.

type-c-service/src/driver/tps6699x.rs
```diff
@@ -371,25 +366,29 @@ impl<const N: usize, M: RawMutex, B: I2c> Controller for Tps6699x<'_, N, M, B> {
...
+
     async fn abort_fw_update(&mut self) -> Result<(), Error<Self::BusError>> {
-        let mut tps6699x = self.tps6699x.borrow_mut();
+        let mut tps6699x = self.tps6699x.lock().await;
         // Check if we're still in firmware update mode
         if tps6699x.get_mode().await? == tps6699x::Mode::F211 {
             let mut delay = Delay;
-            if let Some(update) = self.update_state.replace(None) {
+            let mut state = self.update_state.lock().await;
+            if let Some(update) = state.deref_mut() {
```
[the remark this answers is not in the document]
tullom: Good catch! I found that `.take()` does the same thing as `.replace(None)`

type-c-service/src/driver/tps6699x.rs
tullom: Changed to `try_lock()` to async yield.

embedded-service/src/lib.rs
```diff
@@ -22,6 +22,27 @@ pub mod keyboard;
...
+///
+/// Used because ThreadModeRawMutex is not unit test friendly
+/// but CriticalSectionRawMutex would incur a significant performance impact, since it disables interrupts.
+#[cfg(all(not(test), target_os = "none"))]
+pub type GlobalRawMutex = embassy_sync::blocking_mutex::raw::ThreadModeRawMutex;
+
+/// A cell type that is Sync and Send. CriticalSectionCell is used in a standard context to support multiple cores and
+/// executors, whereas ThreadModeCell is leaner and used in a microcontroller context for when there's a guarantee of a
+/// single core and executor.
+// #[cfg(any(test, not(target_os = "none")))]
```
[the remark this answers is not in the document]
tullom: Oh, I kept that in so you could uncomment it in your PR after this gets merged in 😉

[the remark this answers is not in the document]
kurtjd: Oh makes sense!

kurtjd (dismissed):

felipebalbi (approved):

hid-service/src/i2c/device.rs
```diff
@@ -22,18 +23,17 @@ impl<A: AddressMode + Copy, B: I2c<A>> Device<A, B> {
             device: hid::Device::new(id, regs),
             buffer,
             address,
-            descriptor: Cell::new(None),
-            bus: RefCell::new(bus),
+            descriptor: SyncCell::new(None),
```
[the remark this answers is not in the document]
tullom: Done.

examples/rt685s-evk/src/bin/reset.rs
```diff
@@ -7,7 +7,7 @@ use platform_service::reset;
 use {defmt_rtt as _, panic_probe as _};
 
 async fn print_watcher_number() {
-    static CONTRIVED: embassy_sync::mutex::Mutex<embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex, usize> =
+    static CONTRIVED: embassy_sync::mutex::Mutex<embedded_services::GlobalRawMutex, usize> =
```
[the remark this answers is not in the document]
tullom: Technically not, reverting.

jerrysxie (approved):

kurtjd (approved):

embedded-service/src/lib.rs
```diff
@@ -22,6 +22,27 @@ pub mod keyboard;
...
+pub type SyncCell<T> = critical_section_cell::CriticalSectionCell<T>;
```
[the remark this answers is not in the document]
tullom: No, we are planning on introducing a `ThreadModeCell` in a future PR and gating it behind a feature/architecture, similar to `GlobalRawMutex`

type-c-service/src/driver/tps6699x.rs
```diff
@@ -227,28 +229,28 @@ impl<const N: usize, M: RawMutex, B: I2c> Controller for Tps6699x<'_, N, M, B> {
     type BusError = B::Error;
 
     /// Controller specific initialization
-    #[allow(clippy::await_holding_refcell_ref)]
     async fn sync_state(&mut self) -> Result<(), Error<Self::BusError>> {
         for i in 0..N {
             let port = LocalPortId(i as u8);
-            let mut tps6699x = self.tps6699x.borrow_mut();
+            let mut tps6699x = self.tps6699x.try_lock().expect("Should be infalliable");
```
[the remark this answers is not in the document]
tullom: Good catch, fixed. I think the thought was to not lock once before looping because then the lock is held across an await point (`self.update_port_status()` is called) and we don't need the locked mutex at that point, so explicitly drop the mutex with a new scope. This means if `self.update_port_status()` is changed down the line there is no gotcha and potential deadlock.

asasine (approved):

PR #375, opened 2025-06-26
Only send `SRDY` after Sink Ready interrupt

### LINKED ISSUE #338: type-c-service/tps6699x: Don't issue `SRDY` until we've gotten the sink ready interrupt.

asasine:
Previously, we sent `SRDY` when we received a new consumer contract, but this may fail if PD controller isn't ready to sink power from the provider. The controller indicates this with the Sink Ready interrupt, which is fired when an explicit or implicit contract has been established. As such, Sink Ready always implies a contract is available, and `type_c_service::driver::tps6699x::Tps6699x::update_port_status()` ensures that `PortStatus::available_sink_contract` is valid by the time that `type_c_service::wrapper::ControllerWrapper::process_event()` is called.

Now, we only notify power policy of the new sink contract after the Sink Ready interrupt. In turn, that notification becomes a request to send `SRDY` to the PD controller.

Fixes #338 

Also fixes a bug where we were incorrectly recovering a `power::policy::device::Device` found in an incorrect state.

### DIFF
--- a/type-c-service/src/wrapper/mod.rs
+++ b/type-c-service/src/wrapper/mod.rs
@@ -198,7 +198,8 @@ impl<'a, const N: usize, C: Controller, V: FwOfferValidator> ControllerWrapper<'
                 continue;
             }
 
-            if event.new_power_contract_as_consumer()
+            // Only notify power policy of a contract after Sink Ready event (always after explicit or implicit contract)
+            if event.sink_ready()
                 && self
                     .process_new_consumer_contract(controller, power, local_port_id, &status)
                     .await
--- a/type-c-service/src/wrapper/power.rs
+++ b/type-c-service/src/wrapper/power.rs
@@ -38,10 +38,13 @@ impl<const N: usize, C: Controller, V: FwOfferValidator> ControllerWrapper<'_, N
         info!("current power state: {:?}", current_state);
 
         // Recover if we're not in the correct state
-        if let action::device::AnyState::Detached(state) = power.device_action().await {
-            if let Err(e) = state.attach().await {
-                error!("Error attaching power device: {:?}", e);
-                return PdError::Failed.into();
+        if status.is_connected() {
+            if let action::device::AnyState::Detached(state) = power.device_action().await {
+                warn!("Power device is detached, attempting to attach");
+                if let Err(e) = state.attach().await {
+                    error!("Error attaching power device: {:?}", e);
+                    return PdError::Failed.into();
+                }
             }
         }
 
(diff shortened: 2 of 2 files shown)


type-c-service/src/wrapper/mod.rs
```diff
@@ -198,7 +198,8 @@ impl<'a, const N: usize, C: Controller, V: FwOfferValidator> ControllerWrapper<'
                 continue;
             }
 
-            if event.new_power_contract_as_consumer()
+            // Only notify power policy of a contract after Sink Ready event (always after explicit or implicit contract)
+            if event.sink_ready()
```
tullom, line 202: is the reasoning we don't check for `new_power_contract_as_consumer()` anymore because we want to handle the implicit contract? Will sink ready be latched high even after we `process_new_consumer_contract()`? Is the `is_connected()` check in `process_new_consumer_contract()` to differentiate between implicit and explicit contracts?

type-c-service/src/wrapper/power.rs
```diff
@@ -102,10 +105,13 @@ impl<const N: usize, C: Controller, V: FwOfferValidator> ControllerWrapper<'_, N
         }
 
         // Recover if we're not in the correct state
-        if let action::device::AnyState::Detached(state) = power.device_action().await {
-            if let Err(e) = state.attach().await {
-                error!("Error attaching power device: {:?}", e);
-                return PdError::Failed.into();
+        if status.is_connected() {
```
tullom, line 108: Why do this in process_new_provider_contract() if we never check `sink_ready()`? This fn is gated behind a `new_power_contract_as_provider()` check.

tullom (changes requested):

type-c-service/src/wrapper/mod.rs
asasine: The check for `status.is_connected()` in `process_new_consumer_contract()` fixes a bug that was always present.

We may get a Sink Ready IRQ, then a Plug Inserted or Removed IRQ (from a detach), then we process both together in `process_event()`. In that case, we would first transition the power policy `Device` to the `Detached` state in `process_plug_event()` then would "recover" from the device being in the wrong state in `process_new_consumer_contract()`. The fix now checks `status.is_connected()` before recovering the device, which will be `false` in this case since we retrieve the port status after the Plug Inserted or Removed IRQ which would show the port is in the Detached state.

type-c-service/src/wrapper/power.rs
asasine: See https://github.com/OpenDevicePartnership/embedded-services/pull/375#discussion_r2172455687

PR #382, opened 2025-07-03
Update crates to Rust 2024 edition

kurtjd:
Rust 2024 edition supports several new useful features: https://doc.rust-lang.org/edition-guide/rust-2024/index.html

This PR updates all crates to make use of the latest edition. Editions are designed such that crates of different editions must always be compatible (see: https://doc.rust-lang.org/stable/edition-guide/editions/index.html) thus this update causes no breaking changes for any code dependent on any of the `embedded-services` crates.

Other than a minor new Clippy warning, updating to 2024 only broke a code in a few places, due to stricter `unsafe` requirements. 2024 now requires the `link_section` attribute to be marked unsafe (see: https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-attributes.html) and requires all `extern` blocks to be marked unsafe as well (see: https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-extern.html)

Lastly, `cargo fmt` behavior has also changed (see: https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt-formatting-fixes.html), so many new formatting changes have been applied as well.


power-policy-service/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "power-policy-service"
 version = "0.1.0"
-edition = "2021"
```
jerrysxie, line 4: Non-blocking: if all the crates obeyed the same edition, then we might want to improve upon this to elevate `edition` to the workspace Cargo.toml to simplify updates in the future.

jerrysxie (approved):

power-policy-service/Cargo.toml
kurtjd: This does include updating the workspace `Cargo.toml` edition, however I'm not sure if that gets reflected recursively to all crates within the workspace or not (making it safe to remove `edition` from each individual crate)? Can try it though.

kurtjd: On a side note, while our CI enforces it (we use stable which is currently 1.88), we should consider also updating `rust-toolchain.toml` to enforce building with 1.88 so someone doesn't run into an issue where they are using a different version locally which lets their changes build locally but left scratching their head why CI is failing. That can be a separate PR however.

PR #386, opened 2025-07-07
Fix CFU breaking changes, update battery-service device to use Mutex, and update examples

tullom:
Thank you to @magravel for fixing the changes in embedded-cfu-protocol. This PR includes the same .rs changes in #341.

- Update cfu-service to use constructs in embedded-cfu-protocol v0.2.0
- Further testing showed that I missed some Cells in battery-service/device.rs that should've been fixed in #372
    - Fixed in this PR
- Update examples to be Sync safe and patch embedded-services to remove a circular dependency in the examples' `Cargo.toml`

### DIFF
--- a/battery-service/src/device.rs
+++ b/battery-service/src/device.rs
@@ -1,8 +1,6 @@
-use core::cell::Cell;
-
-use embassy_sync::channel::Channel;
+use embassy_sync::{channel::Channel, mutex::Mutex};
 use embassy_time::Duration;
-use embedded_services::{GlobalRawMutex, Node, NodeContainer};
+use embedded_services::{GlobalRawMutex, Node, NodeContainer, SyncCell};
 
 #[derive(Debug, Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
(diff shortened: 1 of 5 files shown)


kurtjd:
The changes themselves look good to me but the commit structure is a bit odd IMO. I'd suggest the first commit be renamed to something more descriptive than "Test", and it also fails to build (all commits should be able to build independently).

The second commit could maybe be split into several others as it seems to handle a few different things but that's more of a nit.

kurtjd (approved):

kurtjd:
> > The changes themselves look good to me but the commit structure is a bit odd IMO. I'd suggest the first commit be renamed to something more descriptive than "Test", and it also fails to build (all commits should be able to build independently).
> > The second commit could maybe be split into several others as it seems to handle a few different things but that's more of a nit.
> 
> @kurtjd Good catch, this PR was originally intended to be a test PR but it evolved into a full blown change. Renamed the commits to be more descriptive. The first commit is all cargo updates and CFU changes, second commit is battery changes. The commits will be squashed anyways so this separation of commits is just for ease of review.

Ah yeah, forgot that we squash commits, I apologize. But looks good!

PR #389, opened 2025-07-10
Add more information to PortStatus command

zhuyi2024:
1. Add plug_orientation, power_role, data_role to PortStatus command
2. Fix Power switch
3. Add argument "cached" in get_port_status() to call update_port_status() to get the latest port status

### DIFF
--- a/embedded-service/src/type_c/controller.rs
+++ b/embedded-service/src/type_c/controller.rs
@@ -7,7 +7,7 @@ use embassy_sync::signal::Signal;
 use embassy_time::{Duration, with_timeout};
 use embedded_usb_pd::ucsi::lpm;
 use embedded_usb_pd::{
-    Error, GlobalPortId, PdError, PortId as LocalPortId,
+    DataRole, Error, GlobalPortId, PdError, PlugOrientation, PortId as LocalPortId, PowerRole,
     pdinfo::{AltMode, PowerPathStatus},
     type_c::ConnectionState,
 };
@@ -40,6 +40,12 @@ pub struct PortStatus {
     pub connection_state: Option<ConnectionState>,
     /// Port partner supports dual-power roles
     pub dual_power: bool,
+    /// plug orientation
+    pub plug_orientation: PlugOrientation,
+    /// power role
+    pub power_role: PowerRole,
+    /// data role
+    pub data_role: DataRole,
     /// Active alt-modes
     pub alt_mode: AltMode,
     /// Power path status
@@ -55,6 +61,9 @@ impl PortStatus {
             available_sink_contract: None,
             connection_state: None,
             dual_power: false,
+            plug_orientation: PlugOrientation::CC1,
+            power_role: PowerRole::Sink,
+            data_role: DataRole::Dfp,
             alt_mode: AltMode::none(),
             power_path: PowerPathStatus::none(),
         }
@@ -313,8 +322,12 @@ pub trait Controller {
         port: LocalPortId,
     ) -> impl Future<Output = Result<PortEventKind, Error<Self::BusError>>>;
     /// Returns the port status
-    fn get_port_status(&mut self, port: LocalPortId)
-    -> impl Future<Output = Result<PortStatus, Error<Self::BusError>>>;
+    fn get_port_status(
+        &mut self,
+        port: LocalPortId,
+        cached: bool,
+    ) -> impl Future<Output = Result<PortStatus, Error<Self::BusError>>>;
+
     /// Returns the retimer fw update state
     fn get_rt_fw_update_status(
         &mut self,
--- a/embedded-service/src/type_c/controller.rs
+++ b/embedded-service/src/type_c/controller.rs
@@ -96,7 +96,7 @@ impl Default for PortStatus {
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum PortCommandData {
     /// Get port status
-    PortStatus,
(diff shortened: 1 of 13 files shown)


embedded-service/src/type_c/controller.rs
```diff
@@ -421,6 +462,18 @@ pub(super) async fn lookup_controller(controller_id: ControllerId) -> Result<&'s
         .ok_or(PdError::InvalidController)
 }
 
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
zhuyi2024: @kurtjd @tullom Removed

embedded-service/src/type_c/controller.rs
```diff
@@ -82,6 +82,39 @@ impl Default for PortStatus {
     }
 }
 
+/// Port Info
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct PortInfo {
```
RobertZ2011, line 88: Unless there's a specific need for this to be separate, I would just add these fields to the existing `PortStatus` struct. `conn_present` is already provided by the `is_connected` function.

embedded-service/src/type_c/controller.rs
```diff
@@ -82,6 +82,39 @@ impl Default for PortStatus {
...
+    /// connection present
+    pub conn_present: bool,
+    /// plug orientation
+    pub plug_orientation_flipped: bool,
+    /// power role
+    pub power_role_source: bool,
```
RobertZ2011, line 94: Use the embedded-usb-pd `PowerRole` enum instead.

embedded-service/src/type_c/controller.rs
```diff
@@ -82,6 +82,39 @@ impl Default for PortStatus {
...
 }
 
+/// Port Info
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct PortInfo {
+    /// connection present
+    pub conn_present: bool,
+    /// plug orientation
+    pub plug_orientation_flipped: bool,
```
RobertZ2011, line 92: I would be fine calling this just `flipped` but no real strong feelings.

type-c-service/src/driver/tps6699x.rs
```diff
@@ -294,9 +294,49 @@ impl<const N: usize, M: RawMutex, B: I2c> Controller for Tps6699x<'_, N, M, B> {
...
 
+        // sync port status
+        info!("update port status");
+
+        let mut tps6699x = self
+            .tps6699x
+            .try_lock()
+            .expect("Driver should not have been locked before this, thus infallible");
+
+        let _ = self.update_port_status(&mut tps6699x, port).await;
```
asasine, line 305: `embedded_services::type_c::controller::Controller::get_port_status()` gets called on every IRQ for every port so adding this `update_port_status()` will cause every IRQ to read several registers. I think this should be reverted.

FYI @RobertZ2011

zhuyi2024: For the power path status, there will be a delay to enable it, and no IRQ after enabled @asasine

embedded-service/src/type_c/controller.rs
```diff
@@ -82,6 +82,39 @@ impl Default for PortStatus {
...
+    /// power role
+    pub power_role_source: bool,
+    /// data role
+    pub data_role_dfp: bool,
```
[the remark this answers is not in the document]
zhuyi2024: @RobertZ2011 here is the usb-pd PR to add DataRole and PlugOrientation: https://github.com/OpenDevicePartnership/embedded-usb-pd/pull/24

type-c-service/src/driver/tps6699x.rs
RobertZ2011: Is this something that could be added to `enable_sink_path`? From talking with Nathan it sounds like the power path status should only change in response to `SRDY`/`SRYR` so we should get an interrupt when that command completes. Is there a delay after that completes? I have a draft PR that adds logic to read the power path status in `enable_sink_path`: https://github.com/OpenDevicePartnership/embedded-services/pull/348. There's also a power path status changed interrupt, but it's not enabled by default.

embedded-service/src/type_c/controller.rs
zhuyi2024: Fixed.

embedded-service/src/type_c/controller.rs
zhuyi2024: Fixed.

type-c-service/src/driver/tps6699x.rs
zhuyi2024: get_port_status() is called when user use debug tool to print the pdinfo, I think it is reasonable to get the latest status from PD registers which are updated in update_port_status()

asasine: I think a better solution for tne debug tool is to explicitly indicate they want `update_port_status` called, either through a direct call or some sort of argument.

RobertZ2011: I'd be fine with introducing a bool `cached` argument and call `update_port_status` when it's false.

zhuyi2024: Updated.

asasine (approved):

RobertZ2011 (approved):

PR #399, opened 2025-07-11
power-policy: Add unconstrained logic

RobertZ2011:
* Add consumer and provider flags to allow signaling more info to policy without needing future breaking changes
* Implement logic to prioritize unconstrained devices

### DIFF
--- a/embedded-service/src/power/policy/flags.rs
+++ b/embedded-service/src/power/policy/flags.rs
@@ -1,0 +1,53 @@
+//! Consumer and provider flags, these are used to signal additional information about a consumer/provider request
+
+use bitfield::bitfield;
+
+bitfield! {
+    /// Raw consumer flags bit field
+    #[derive(Copy, Clone, PartialEq, Eq)]
+    #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+    struct ConsumerRaw(u32);
+    impl Debug;
+    /// Unconstrained power, indicates that we are drawing power from something like an outlet and not a limited source like a battery
+    pub u8, unconstrained_power, set_unconstrained_power: 0, 0;
+}
+
+/// Type safe wrapper for consumer flags
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Consumer(ConsumerRaw);
+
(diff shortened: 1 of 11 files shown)


embedded-service/src/power/policy/flags.rs
```diff
@@ -0,0 +1,62 @@
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Provider(ProviderRaw);
+
+impl Provider {
+    /// Create a new provider with no flags set
+    pub const fn none() -> Self {
+        Self(ProviderRaw(0))
+    }
+}
```
asasine, line 62: I question whether we should have this at all since it's empty. Using a newtype `struct ProviderPowerCapability(PowerCapability)` in `embedded_services::power::policy` would be fine for now.

power-policy-service/src/consumer.rs
```diff
@@ -13,24 +14,30 @@ pub struct State {
...
+    if ord == Ordering::Equal {
+        // Prioritize unconstrained devices
+        match (a.flags.unconstrained_power(), b.flags.unconstrained_power()) {
+            (true, false) => Ordering::Greater,
+            (false, true) => Ordering::Less,
+            _ => Ordering::Equal,
+        }
+    } else {
+        ord
     }
```
asasine, line 35: This `unconstrained_power` ordering is just the default `impl Ord for bool` therefore this is the default Ord for a tuple
```suggestion
    // Intially compare power first, but prioritize unconstrained devices if equal
    (a.capability, a.flags.unconstrained_power()).cmp(&(b.capability, b.flags.unconstrained_power()))
```

asasine (approved):

power-policy-service/src/consumer.rs
RobertZ2011: Done.

asasine (approved):

jerrysxie (approved):

jerrysxie:
@RobertZ2011 Please announce breaking changes to Zulip and Discord:
- https://discord.com/channels/1345544766976299068/1362107636828606665
- https://opendevicepartnership.zulipchat.com/#narrow/channel/495405-embedded-controller

PR #405, opened 2025-07-15
Add target check to workflow

kurtjd:
Our code base now builds differently based on target (e.g. `std/x86_64-unknown-linux-gnu` vs `thumbv8m.main-none-eabihf`).
For example if targeting std, `CriticalSectionCell` is built, and if not, `ThreadModeCell` is built.

Our original workflow only checked the default (`x86_64-unknown-linux-gnu`) target, which led to a PR making it past that failed MSRV check since it included code only built when target is `thumbv8m.main-none-eabihf`, but that code wasn't stable under 1.85 (see #404).

This PR ensures the MSRV job analyzes both targets (and additionally the clippy jobs as well).

The test job remains unchanged since that should always run under `std`, and likewise the examples jobs are unchanged as those explicitly mention their targets.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -79,13 +79,13 @@ jobs:
     # which is required for feature unification
     # additionally, replacing `check` with `clippy` ensures all combinations of features generate no clippy warnings
     runs-on: ubuntu-latest
-    # The name is kept as 'stable / clippy' and `beta / clippy` to match required checks from repo settings
-    name: ${{ matrix.toolchain }} / clippy
+    name: ${{ matrix.toolchain }} / ${{ matrix.target }} / clippy
     strategy:
       fail-fast: false
       matrix:
         # Get early warning of new lints which are regularly introduced in beta channels.
         toolchain: [stable, beta]
+        target: [x86_64-unknown-linux-gnu, thumbv8m.main-none-eabihf]
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


RobertZ2011 (dismissed):

kurtjd:
> > Our code base now builds differently based on target (e.g. `std/x86_64-unknown-linux-gnu` vs `thumbv8m.main-none-eabihf`). For example if targeting std, `CriticalSectionCell` is built, and if not, `ThreadModeCell` is built.
> > Our original workflow only checked the default (`x86_64-unknown-linux-gnu`) target, which led to a PR making it past that failed MSRV check since it included code only built when target is `thumbv8m.main-none-eabihf`, but that code wasn't stable under 1.85 (see #404).
> > This PR ensures the MSRV job analyzes both targets (and additionally the clippy jobs as well).
> > The test job remains unchanged since that should always run under `std`, and likewise the examples jobs are unchanged as those explicitly mention their targets.
> > Small note: This now results in jobs with duplicate names appearing in the job list since the target matrix has two entries causing the job to run twice. The ideal fix would be to add the target to the job name, but unfortunately the repo settings require specific names for these jobs. Would be a good idea to change the repo settings to look for names with the targets appended since having duplicate names currently is a bit confusing (but doesn't cause any issues beyond that).
> 
> @kurtjd it is definitely possible to update the status check to point to different job, so we should append the target name just so it is more clear which architecture is failing.

Sounds good, updated it. New names are:

`stable / x86_64-unknown-linux-gnu / clippy`
`stable / thumbv8m.main-none-eabihf / clippy`
`beta / x86_64-unknown-linux-gnu / clippy`
`beta / thumbv8m.main-none-eabihf / clippy`
`ubuntu / 1.85 / x86_64-unknown-linux-gnu`
`ubuntu / 1.85 / thumbv8m.main-none-eabihf`

jerrysxie (approved):

PR #411, opened 2025-07-17
power-policy: Broadcast events

RobertZ2011:
This provides a more general way to receive events than the comms service.

PR #412, opened 2025-07-17
Immediate broadcaster

RobertZ2011:


RobertZ2011:
> @RobertZ2011 Where do we intent to use this immediate broadcaster?

The short term need is to support unconstrained power with https://github.com/OpenDevicePartnership/embedded-services/pull/411.

PR #413, opened 2025-07-17
type-c-service: Add PD alert support

RobertZ2011:
Currently rebased on top of https://github.com/OpenDevicePartnership/embedded-services/pull/390 until that lands.

### DIFF
--- a/type-c-service/src/wrapper/mod.rs
+++ b/type-c-service/src/wrapper/mod.rs
@@ -16,10 +16,13 @@ use embedded_services::type_c::event::{PortEvent, PortNotificationSingle, PortPe
 use embedded_services::GlobalRawMutex;
 use embedded_services::SyncCell;
 use embedded_services::{debug, error, info, trace, warn};
+use embedded_usb_pd::ado::Ado;
 use embedded_usb_pd::{Error, PdError, PortId as LocalPortId};
 
+use crate::wrapper::backing::Backing;
 use crate::{PortEventStreamer, PortEventVariant};
 
+pub mod backing;
 mod cfu;
 mod pd;
 mod power;
@@ -59,8 +62,8 @@ pub trait FwOfferValidator {
 pub enum Event<'a> {
     /// Port status changed
     PortStatusChanged(LocalPortId, PortStatusChanged),
-    /// Port notification received
-    PortNotification(LocalPortId, PortNotificationSingle),
+    /// PD alert
+    PdAlert(LocalPortId, Ado),
     /// Power policy command received
     PowerPolicyCommand(
         LocalPortId,
@@ -73,7 +76,7 @@ pub enum Event<'a> {
 
 /// Takes an implementation of the `Controller` trait and wraps it with logic to handle
 /// message passing and power-policy integration.
-pub struct ControllerWrapper<'a, const N: usize, C: Controller, V: FwOfferValidator> {
+pub struct ControllerWrapper<'a, const N: usize, C: Controller, BACK: Backing<'a>, V: FwOfferValidator> {
     /// PD controller to interface with PD service
     pd_controller: controller::Device<'a>,
     /// Power policy devices to interface with power policy service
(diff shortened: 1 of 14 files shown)


type-c-service/src/wrapper/mod.rs
```diff
@@ -88,6 +92,10 @@ pub struct ControllerWrapper<'a, const N: usize, C: Controller, V: FwOfferValida
     fw_version_validator: V,
     /// FW update ticker used to check for timeouts and recovery attempts
     fw_update_ticker: Mutex<GlobalRawMutex, embassy_time::Ticker>,
+    /// Pending PD alerts
+    // Even though we want the immediate publishing behavior of a PubSub channel,
+    // we'd have declare it externally because it doesn't have publish functions on it like a standard channel does
```
asasine, line 97: I don't understand why a `PubSubChannel<GlobalRawMutex, Ado, MAX_BUFFERED_PD_ALERTS, 1, 1>` wouldn't work. Immediate publishers don't take up a publisher slot so you'd get the immediate publishing feature you're after without the race that `if full() { receive() }` has

type-c-service/src/wrapper/mod.rs
```diff
@@ -71,6 +73,8 @@ pub enum Event<'a> {
     CfuEvent(cfu::Event),
 }
 
+const MAX_BUFFERED_PD_ALERTS: usize = 4;
```
asasine, line 76: How was this number chosen?

type-c-service/src/wrapper/pd.rs
```diff
@@ -72,6 +72,14 @@ impl<const N: usize, C: Controller, V: FwOfferValidator> ControllerWrapper<'_, N
                     Error::Pd(e) => Err(e),
                 },
             },
+            controller::PortCommandData::GetPdAlert => {
+                let channel = &self.pd_alerts[local_port.0 as usize];
+                if channel.is_empty() {
+                    Ok(controller::PortResponseData::PdAlert(None))
+                } else {
+                    Ok(controller::PortResponseData::PdAlert(Some(channel.receive().await)))
+                }
```
asasine, line 81: This is a `try_receive()`
```suggestion
                controller::PortResponseData::PdAlert(channel.try_receive().ok())
```

asasine (changes requested):

asasine (approved):

type-c-service/src/wrapper/mod.rs
RobertZ2011: Ended up creating a `backing` module to contain channels and other similar data needed by the wrapper. The issue is that things like `DynImmediatePublisher` are just references so attempting to store the channel and one of those in the same struct is self-referential. `backing` contains code to make it easy to declare everything the wrapper needs.

type-c-service/src/wrapper/mod.rs
```diff
@@ -215,6 +222,31 @@ impl<'a, const N: usize, C: Controller, V: FwOfferValidator> ControllerWrapper<'
         Ok(())
     }
 
+    /// Process a PD alert
+    async fn process_pd_alert(&self, port: LocalPortId, alert: Ado) -> Result<(), Error<<C as Controller>::BusError>> {
```
jerrysxie, line 226: The state machine has gotten complicated enough such that I cannot exactly follow the code flow just through code inspection. I will rely on @asasine's review on whether this is sound.

jerrysxie: Not blocking, just an observation.

jerrysxie (approved):

type-c-service/src/wrapper/mod.rs
asasine: I'm of the same opinion, and have previously brought it up in https://github.com/OpenDevicePartnership/embedded-services/pull/390#pullrequestreview-3027356979 and elsewhere.

Using this involves implementing our own event loop in product code for each controller that alternates between calling `ControllerWrapper::wait_next()` and `ControllerWrapper::process_event()`. Upon receiving an `Event::PdAlert`, we can customize the behavior beyond what's available in `process_event()`.

PR #419, opened 2025-07-18
Have rolling workflow create PR if update successful

kurtjd:
This PR modifies the rolling workflow to perform a `cargo update` and if no builds or tests are broken, create a PR to update `Cargo.lock` files. It accomplishes this by:

- Running `cargo update` on workspace root and example folders (which are excluded from the workspace)
- If this results in a diff, run our `check.yml` job (with updated `Cargo.lock` files) for thorough checking
- If `check.yml` fails, no PR is created and we can see which job in `check.yml` failed in the Actions tab
- If `check.yml` passes, a PR is created with updated `Cargo.lock` files

The `check.yml` workflow has been modified to download the updated `Cargo.lock` files artifact if called by `rolling.yml`.

This has been tested reasonably thoroughly on my fork but it's hard to test all possible cases, so if it doesn't always perform as expected this workflow can be modified to address issues in future PRs.

Of note is the PRs generated by this workflow won't trigger other workflows without special workarounds involving new tokens (see: [note](https://github.com/peter-evans/create-pull-request/blob/v5/docs/concepts-guidelines.md#triggering-further-workflow-runs)). This means the PR itself won't trigger the `check.yml` workflow. This might not be a problem since the PR doesn't get created if `check.yml` fails in the first place, and if we update the PR ourselves (e.g. to rebase it off new main changes) that should trigger the workflow then.

### DIFF
--- a/.github/workflows/rolling.yml
+++ b/.github/workflows/rolling.yml
@@ -1,101 +1,80 @@
-# This workflow runs every morning at midnight. It will run cargo hack
-# and a build with msrv. If any dependency breaks our crate, we will
-# know ASAP.
-#
-# - hack: cargo hack for all workspace packages
-# - msrv: check that the msrv specified in the crate is correct
-# - check-*-examples: ensures examples still compile
+# This workflow runs every morning at midnight.
+# It runs cargo update and checks that no builds or tests are broken.
+# If not, a PR is created with updated Cargo.lock files.
 permissions:
-  contents: read
+  contents: write
+  pull-requests: write
 
 on:
   schedule:
     - cron: '0 0 * * *'
+  workflow_dispatch:
 
 name: rolling
+
 jobs:
-
-  hack:
-    # cargo-hack checks combinations of feature flags to ensure that features are all additive
-    # which is required for feature unification
+  update-dependencies:
+    name: cargo-update / update-dependencies
     runs-on: ubuntu-latest
-    name: ubuntu / stable / features
-    strategy:
-      fail-fast: false
+    outputs:
+      pr-needed: ${{ steps.check-pr.outputs.changes }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/rolling.yml
```diff
@@ -1,101 +1,80 @@
...
-      - name: Install stable
-        uses: dtolnay/rust-toolchain@stable
-      - name: cargo check
-        working-directory: ${{ matrix.example_directory }}
-        run: |
-          cargo update
-          cargo check --target thumbv8m.main-none-eabihf
+          name: updated-lock-files
+          path: |
+            **/Cargo.lock
```
jerrysxie, line 46: Does this upload Cargo.lock for both the workspace and the examples?

.github/workflows/rolling.yml
```diff
@@ -1,101 +1,80 @@
...
+
+  check-build:
+    name: cargo-update / check-build
+    needs: update-dependencies
+    if: ${{ needs.update-dependencies.outputs.pr-needed == 'true' }}
+    uses: ./.github/workflows/check.yml
```
jerrysxie, line 52: This is a great idea, less duplication. Please update the template repo: https://github.com/OpenDevicePartnership/embedded-rust-template as well so new repo can have uniform setup.

.github/workflows/rolling.yml
kurtjd: Yep the ** pattern should match all Cargo.lock recursively from the root folder. Tested it out and it looked like it noticed all Cargo.lock files if changed.

.github/workflows/rolling.yml
kurtjd: Sounds good, will get that updated!

jerrysxie (approved):

felipebalbi (approved):

PR #427, opened 2025-07-23
Fix running `cargo test` in workspace root

dymk:
Allow running `cargo test` / `cargo build` in the workspace root by scoping `embassy-executor` features to builds for `arm` targets.


type-c-service/src/wrapper/mod.rs
```diff
@@ -311,7 +311,7 @@ impl<'a, const N: usize, C: Controller, V: FwOfferValidator> ControllerWrapper<'
     }
 }
 
-impl<'a, const N: usize, C: Controller, V: FwOfferValidator> Object<C> for ControllerWrapper<'a, N, C, V> {
+impl<const N: usize, C: Controller, V: FwOfferValidator> Object<C> for ControllerWrapper<'_, N, C, V> {
```
tullom, line 314: unrelated change?

tullom (commented):
Looks like this change exposed some new clippy errors in the CI, otherwise one comment and LGTM

type-c-service/src/wrapper/mod.rs
dymk: This was one of the new clippy errors exposed (somehow wasn't caught in a previous CI run?)

espi-service/Cargo.toml
```diff
@@ -7,21 +7,17 @@ repository = "https://github.com/OpenDevicePartnership/embedded-services"
...
-embassy-imxrt = { workspace = true, features = [
-    "time-driver-os-timer",
-    "time",
-    "mimxrt633s",
-] }
-embassy-executor = { workspace = true, features = [
-    "executor-thread",
-    "executor-interrupt",
-] }
+embassy-imxrt = { workspace = true, features = ["mimxrt633s"] }
```
jerrysxie, line 19: Oh, I see, now the test workflow will not freak out and we don't need to split it into different crate.

jerrysxie (approved):

jerrysxie (commented):
Pending fixing of all the CI failures.

type-c-service/src/wrapper/mod.rs
kurtjd: These two look like definitely something clippy would complain about, but yeah weird that CI didn't catch and I can't seem to reproduce getting this to generate clippy warnings locally either. Out of curiosity what version and toolchain did you use and which arguments (if any) did you pass to clippy?

kurtjd: Ah `cargo +1.85 clippy` does the trick. Our current MSRV workflow only does a `cargo +.185 check` not a fully clippy, so it seems only the 1.85 version generates clippy warnings (but not sure why, these seem relevant in later versions as well). Might update the workflow after this so the MSRV job does a `clippy`.

kurtjd (approved):

PR #444, opened 2025-08-01
Handle MCTP OOB messages in espi service and forward payload to subsystems

tullom:


espi-service/src/espi_service.rs
```diff
@@ -116,28 +121,123 @@ impl Service<'_> {
...
+        return Err(MctpError::WrongDestinationAddr);
+    }
+
+    // MCTP command code is 0x0F.
+    if mctp_msg[1] != 0x0F {
+        return Err(MctpError::InvalidCommandCode);
+    }
+
+    // Check the byte count is correctly formed and is not larger than the max in the spec.
+    if usize::from(mctp_msg[2]) != mctp_msg.len() - 3 || mctp_msg[2] > 69 {
```
kurtjd, line 186: Heads up Phil discovered that the N1x eSPI controller behaves oddly if packet sizes sent aren't multiples of 4, so he had to modify the secure-partition side to pad 0s to get to a multiple of 4. Thus mctp_msg.len() - 3 might not actually match what's in mctp_msg[2] due to the padding. Don't think he pushed those changes upstream yet but something to keep in mind.

espi-service/src/espi_service.rs
```diff
@@ -116,28 +121,123 @@ impl Service<'_> {
...
+    InvalidHeaderVersion,
+    InvalidDestinationEndpoint,
+    InvalidSourceEndpoint,
+    /// Multi message not supported
+    InvalidFlags,
+}
+
+type PayloadLen = usize;
+
+fn handle_mctp_header(mctp_msg: &[u8], data: &mut [u8]) -> Result<(EndpointID, PayloadLen), MctpError> {
```
jerrysxie, line 168: Is this MCTP message building code just temporary placeholder until we builds up the libraries? If it is, we probably don't need unit tests for them. But if we keep this going forward, then we should add unit tests.

espi-service/src/espi_service.rs
```diff
@@ -207,11 +309,32 @@ pub async fn espi_service(mut espi: espi::Espi<'static>, memory_map_buffer: &'st
...
+                                payload_len,
+                            };
+                            espi_service.endpoint.send(endpoint, &acpi_msg).await.unwrap();
+                        }
+                        Err(e) => error!("MCTP packet malformed: {:?}", e),
+                    }
+
+                    info!("MCTP packet sent!");
+
+                    let acpi_response = espi_service.comms_signal.wait().await;
```
jerrysxie, line 325: I believe MCTP allows messages and responses to be sent out of order. If we wait here, then each message has to be respond to sequentially before the next event on eSPI can be handle. Is that what we want?

kurtjd (approved):
Again looks great. Only minor nit: With a separate MCTP module it could make sense to move some of these numbers to constants for more clarity. E.g. a `const MAX_MCTP_PACKET_LEN: usize = 69` or something like that.

espi-service/src/espi_service.rs
```diff
@@ -116,28 +121,123 @@ impl Service<'_> {
...
+    let msg_tag = mctp_msg[7] & 0b0000_0111;
+
+    // Verify flags
+    if !som || !eom || seq_num != 1 || msg_tag != 3 {
+        return Err(MctpError::InvalidFlags);
+    }
+
+    info!("Test2");
+
+    data[..usize::from(mctp_msg[2]) - 5].copy_from_slice(&mctp_msg[8..]);
```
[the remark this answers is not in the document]
tullom: Done.

espi-service/src/espi_service.rs
tullom: Indeed this is just temporary until @dymk's MCTP library is ready to use. I tried to add unit tests anyways but espi-service has a dependency on Embassy and the cargo test runner requires `std`, which means adding tests aren't trivial. I moved the MCTP handling code to embedded-service but i get a link time error when running `cargo test`

espi-service/src/espi_service.rs
tullom: For the time being, (this E2E demo), yes, but we will want to change this in the future to handle the messages asynchronously. Logged issue in #448

kurtjd (dismissed):

jerrysxie (dismissed):

espi-service/src/espi_service.rs
```diff
@@ -207,29 +221,57 @@ pub async fn espi_service(mut espi: espi::Espi<'static>, memory_map_buffer: &'st
...
-                        Ok(dest_slice) => {
-                            dest_slice[..src_slice.len()].copy_from_slice(src_slice);
-                        }
-                        Err(_e) => {
-                            #[cfg(feature = "defmt")]
-                            error!("Failed to retrieve OOB write buffer: {}", _e);
-                            espi.complete_port(port_event.port).await;
-                            continue;
+                    match handle_mctp_header(src_slice, acpi_buf::get_mut().unwrap().borrow_mut().borrow_mut()) {
+                        Ok((endpoint, payload_len)) => {
```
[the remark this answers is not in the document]
tullom: Turns out I needed to do something like this anyways to explicitly drop the mutable reference to the shared buffer. So it all works out.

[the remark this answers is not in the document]
jerrysxie: I see, dropping the mutable reference here is more for future proofing. As it is, since we block the execution of this event loop until the response is sent, we effectively have exclusive access to the buffer.

jerrysxie (approved):

PR #445, opened 2025-08-01
type-c-service: UCSI PPM implementation

RobertZ2011:
Example output:
```
[2025-08-01T23:04:40Z INFO  type_c_ucsi] Resetting PPM...
[2025-08-01T23:04:40Z INFO  type_c_service::task] Starting type-c task
[2025-08-01T23:04:40Z INFO  type_c_ucsi] PPM reset successful
[2025-08-01T23:04:40Z INFO  type_c_ucsi] Set Notification enable...
[2025-08-01T23:04:40Z INFO  type_c_ucsi] Set Notification enable successful
[2025-08-01T23:04:40Z INFO  type_c_ucsi] Get PPM capabilities...
[2025-08-01T23:04:40Z INFO  type_c_ucsi] Get PPM capabilities successful: Some(PpmResponse(GetCapability(ResponseData { attributes: Attributes(AttributesRaw { .0: 0, disabled_state_support: false, battery_charging: false, usb_power_delivery: false, usb_type_c_current: false, bm_power_source: 0 }), num_connectors: 2, optional_features: OptionalFeatures(OptionalFeaturesRaw { .0: 0, set_ccom_supported: false, set_power_level_supported: false, altmode_details_supported: false, altmode_override_supported: false, pdo_details_supported: false, cable_details_supported: false, external_supply_notif_supported: false, pd_reset_notif_supported: false, get_pd_msg_supported: false }), num_alt_modes: 0, bcd_battery_charging_spec: 288, bcd_usb_pd_spec: 768, bcd_type_c_spec: 512 })))
```


RobertZ2011:
Depends on https://github.com/OpenDevicePartnership/embedded-usb-pd/pull/32

type-c-service/src/service/ucsi.rs
```diff
@@ -0,0 +1,193 @@
...
+        (Cci::new_reset_complete().into(), Some(PpmInput::Reset))
+    }
+
+    /// PPM get capabilities implementation
+    fn process_get_capabilities(&self) -> (Response, Option<PpmInput>) {
+        debug!("Get PPM capabilities: {:?}", self.config.ucsi_capabilities);
+        (
+            Response {
+                cci: Cci::new_cmd_complete(),
+                // TODO: pull num connectors from service
```
asasine, line 40: Isn't the number of connectors already stored in `self.config.ucsi_capabilities`?

type-c-service/src/service/ucsi.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub(super) async fn process_ucsi_command(&self, command: &Command) -> Result<external::UcsiResponse, PdError> {
+        let state = &mut self.state.lock().await.ucsi;
+        let (mut response, ppm_input) = match state.ppm_state_machine.state() {
+            PpmState::Idle(false) => self.process_ppm_state_idle_disabled(state, command).await,
+            PpmState::Idle(true) => self.process_ppm_state_idle_enabled(state, command).await,
+            PpmState::Busy(_) => self.process_ppm_state_busy(command).await,
+            PpmState::ProcessingCommand => self.process_ppm_state_processing_command(command).await,
+            PpmState::WaitForCommandCompleteAck => self.process_ppm_state_wait_for_command_complete_ack(command).await,
+            PpmState::WaitForAsyncEventAck => self.process_ppm_state_wait_for_async_event_ack(command).await,
+        }?;
```
asasine, line 157: This, in combination with how the `process_foo()` methods are implemented, feels like the state machine is being reimplemented. "Given `state` and `command`, transition the state according to the returned `ppm_input`." Why isn't the state machine responsible for the transitioning and creating the response and output?

type-c-service/src/service/ucsi.rs
```diff
@@ -0,0 +1,193 @@
...
+
+use super::*;
+
+/// UCSI state
+#[derive(Default)]
+pub(super) struct State {
+    /// PPM state machine
+    ppm_state_machine: StateMachine,
+    /// Currently enabled notifications
+    notifications_enabled: NotificationEnable,
```
asasine, line 17: This is derivable state and doesn't need to be tracked explicitly. It's already available based on which state the state machine is in.

type-c-service/src/service/ucsi.rs
RobertZ2011: Generating the response is implementation defined and I didn't want to put any kind of restrictions on that by making it part of the state machine. Transitioning from `ProcessingCommand` to either `CommandImmediate` or `CommandInProgress` also depends on if the implementation supports asynchronous execution of commands. But I do think that adding a `Command` output would allow a lot of this logic to be moved into the general state machine. Same as the other comment, I can do this as a follow-up USCI improvements PR, or let me know if you think I should land that now.

type-c-service/src/service/ucsi.rs
RobertZ2011: It is, but the number from the service should be more accurate and would reduce the amount of manual configuration. I can break up the struct to allow using just the portion without the port count to avoid confusion. But I might make that part of my next embedded-usb-pd PR, let me know if you think I should land that now.

asasine: No strong opinion

type-c-service/src/service/ucsi.rs
asasine: Follow-up is fine.

type-c-service/src/service/ucsi.rs
RobertZ2011: Not quite, these are flags containing which notifications are enabled. If the state machine is in `Idle(false)` that implies none of these flags set, but in `Idle(true)` various combinations of these could be set depending on what the OPM wants. This is also something that could end up as part of the general state machine, but let me know what you think.

asasine: I see. It wasn't obvious to me this is *which* notifications are enabled; it seemed like it was equivalent to `Idle(enabled: bool)`

No preference at the moment

asasine (approved):

jerrysxie (approved):

PR #449, opened 2025-08-05
add debug service

kat-perez:
Add defmt logging infrastructure with circular buffer system for debug message forwarding over eSPI, including the complete transport layer and mock service implementations


embedded-service/src/ec_type/mod.rs
```diff
@@ -104,6 +104,25 @@ pub fn update_time_alarm_section(msg: &message::TimeAlarmMessage, memory_map: &m
     }
 }
 
+/// Update debug section of memory map based on debug message
+pub fn update_debug_section(msg: &message::DebugMessage, memory_map: &mut structure::ECMemory) {
```
jerrysxie, line 108: So eSPI has 4 channels:
- peripheral channel (packetized memory mapped reads/writes)
- virtual channel (packetized interrupts)
- OOBE channel (smbus packet over eSPI)
- flash channel (shared flash access through eSPI)

Originally we were planning to use peripheral channel. But later on we decided to switch to OOBE channel instead. This memory map code is for the peripheral channel. We should opt to use the OOBE channel instead. Please look at https://github.com/OpenDevicePartnership/embedded-services/pull/444 as a reference. Debug service will be using the same OOBE channel mechanism.

jerrysxie: @philgweber We are still planning to route the debug message through OOBE channel?

kat-perez: @jerrysxie @philgweber in this latest iteration, I added a debug_receiver in espi_service and then start the OOB transmission there when a debug message is ready in the buffer

A sample of how this looks integrated into soc-embedded-controller is here: https://github.com/OpenDevicePartnership/soc-embedded-controller/pull/48/files

jerrysxie: Inserting debug event into the eSPI handling loop is probably not the ideal solution, then debug message can easily overload the loop and affects the eSPI transaction handling flow.

@philgweber Does OOBE channel allow the EC slave to send OOBE response unprompted? Does it make use of the eSPI alert line? The eSPI spec is kind of vague about that. Or is it more like u2c/smbus where the slave has to notify before the host sends an OOBE message to get an OOBE response from the EC slave?

@tullom Matteo's current implementation has not added support for notification yet:
1. When an OOBE message comes in, it will be forwarded to the subsystem service
2. Then it waits for the subsystem service to send a response, which will be sent on the OOB channel

Can we follow the same pattern for debug service for now before notification is added? And if host did not send an OOBE request to pick up the debug message in time, then the debug messages gets dropped.

debug-service/src/transport/espi.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Debug data to send via OOB
+    pub data: heapless::Vec<u8, MAX_OOB_DEBUG_SIZE>,
+    /// Port number for OOB transmission (0 for debug)
+    pub port: u8,
+}
+
+/// Global channel for sending debug messages to eSPI service
+/// This channel allows the debug service to queue messages for OOB transmission
+/// Increased capacity to handle burst traffic better
+static DEBUG_CHANNEL: Channel<CriticalSectionRawMutex, EspiDebugMessage, 32> = Channel::new();
```
jerrysxie, line 23: Do we need critical section or just threadmode mutex? Do we need to share data between interrupt and normal Embassy task?

jerrysxie (changes requested):

philgweber (approved):

kat-perez:
# Cargo Vet Audit

1. What crates (with version) need to be audited?
debug-service (0.1.0)

3. How many of the crates are version updates vs new dependencies?
New dependencies

4. To confirm none of the already included crates serve your needs, please provide a brief description of the purpose of the new crates.
Adding a new service

6. Any extra notes to the auditors to help with their audits.
N/A

kat-perez:
> This implementation here concerns me. Marking `debug_service::defmt_ring_logger::DefmtLogger` with `#[defmt::global_logger]` forces all users who depend on the `debug_service` crate to use this logger and prevents them from using their own.
> 
> I think the design here should consider composability first and foremost. Create the abstractions and mechanisms that are useful for OEMs to reuse the tasks and types defined here but let them have total control over the exact final composition. What if an OEM wants logs over both eSPI and RTT and a third or fourth protocol? What if they want to filter by log level, or introduce any other logic not present here? There's no way with the current implementation without reimplementing a lot, and then `debug_service` becomes either brittle or not useful. This is also reflected in the task being built on top of various private global statics in the `defmt_ring_logger` module; what if I want to use the buffering mechanism but use a different task implementation? I can't.
> 
> The upstream `defmt` project handles some of this by putting the logger into its own dedicated crate `defmt_rtt`.
> 
> Just a comment, not "request changes," since I'm not intending to block the PR. Just know that the implementation here isn't usable in any capacity for us right now despite having meaningful overlap with what we've written internally. It could be, though.

Thanks for the feedback. What I can do for now is list the things you mentioned as feature requests but for now our design is to keep it as minimal as possible for this first merge to main. This is currently designed to work only for eSPI at the moment but could be extended in the future with configure endpoints. Open to more discussion in the feature requests.

embedded-service/src/ec_type/mod.rs
philgweber: FYI eSPI does allow unprompted OOB responses. We are adding support on host side in the interrupt notification side to send a notification to the corresponding service that sends the OOB data. Until then Kat is adding a notification and we will send a GET_OOB request to retrieve the debug data.

debug-service/src/defmt_ring_logger.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+}
+
+// Static buffer for ACPI-style messages carrying defmt frames
+embedded_services::define_static_buffer!(defmt_acpi_buf, u8, [0u8; DEFMT_MAX_BYTES as usize]);
+static DEFMT_ACPI_BUF_OWNED: StaticCell<OwnedRef<'static, u8>> = StaticCell::new();
+
+#[embassy_executor::task]
+pub async fn defmt_to_host_task() {
+    defmt::info!("defmt to host task start");
```
jerrysxie, line 180: Sorry, I just noticed this. Why does it needs to call `info!` in 2 different places? Can it use the embedded-services logging macros to switch between defmt (embedded env) and log (std env) like other services?

* features: https://github.com/OpenDevicePartnership/embedded-services/blob/ea84b81c44b08c18c298537b3e95d4d3f84e89d9/thermal-service/Cargo.toml#L25C1-L26C1
* embedded-service fmt macros: https://github.com/OpenDevicePartnership/embedded-services/blob/main/embedded-service/src/fmt.rs
* use: https://github.com/OpenDevicePartnership/embedded-services/blob/ea84b81c44b08c18c298537b3e95d4d3f84e89d9/thermal-service/src/lib.rs#L7C1-L8C1

jerrysxie (changes requested):

jerrysxie (changes requested):

jerrysxie (approved):

debug-service/src/defmt_ring_logger.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    // Acquire the staging buffer once; we own it for the task lifetime.
+    let acpi_owned = defmt_acpi_buf::get_mut().expect("defmt staging buffer already initialized elsewhere");
+
+    loop {
+        // Wait for a complete defmt frame to be available (do not release yet)
+        embedded_services::info!("waiting for defmt frame");
+        let frame = framed_consumer.wait_read().await;
+
+        // Copy frame bytes into the static ACPI buffer.
```
kurtjd, line 201: Had to make room here for the payload header:

```rust
// Copy frame bytes into the static ACPI buffer.
        // Producer commits frames atomically with size ≤ DEFMT_MAX_BYTES (1024),
        // so the consumer never sees a partial frame. We still clamp to the
        // destination length to be robust if the staging buffer size changes.
        let copy_len = core::cmp::min(frame.len(), acpi_owned.len());
        {
            let mut access = acpi_owned.borrow_mut();
            let buf: &mut [u8] = BorrowMut::borrow_mut(&mut access);
            buf[0] = 1;
            buf[1] = 0;
            buf[2] = 0;
            buf[3] = 1;
            buf[4..copy_len + 4].copy_from_slice(&frame[..copy_len]);
        }
        embedded_services::info!("got frame: bytes={}, copy_len={}", frame.len(), copy_len);

        frame.release();
        embedded_services::info!("released defmt frame (staged {} bytes)", copy_len);

        // Notify the host that data is available
        let _ = comms::send(
            EndpointID::Internal(Internal::Debug),
            host_ep,
            &HostMsg::Notification(NotificationMsg { offset: 20 }),
        )
        .await;

        // Wait for host notification/ack via the debug service
        let _n = response_notify_signal().wait().await;
        embedded_services::info!("host ack received, sending defmt response");

        // Send the staged defmt bytes frame as an ACPI-style message.
        // Scope the message so the shared borrow is dropped before we clear the buffer.
        {
            let msg = HostMsg::Response(AcpiMsgComms {
                payload: defmt_acpi_buf::get(),
                payload_len: copy_len + 4,
            });
            let _ = comms::send(EndpointID::Internal(Internal::Debug), host_ep, &msg).await;
            embedded_services::info!("sent {} defmt bytes to host", copy_len);
        }
   ```
   But if switching to Matteo's updates soon not a huge deal.

kurtjd (dismissed):
Not sure if you were planning to do this as part of this PR, but the endpoint id of 4 for debug service will need to be added at these points:

https://github.com/OpenDevicePartnership/embedded-services/blob/ce5540cfcbe34283229ca7999ff833e2f68c2469/embedded-service/src/ec_type/protocols/mctp.rs#L81

https://github.com/OpenDevicePartnership/embedded-services/blob/ce5540cfcbe34283229ca7999ff833e2f68c2469/embedded-service/src/ec_type/protocols/mctp.rs#L140

To work with the current system. Feel free to disregard if you had other plans.

kurtjd (dismissed):

kurtjd (dismissed):

jerrysxie (approved):

kurtjd (approved):

PR #460, opened 2025-08-12
espi-service: Notification Support

tullom:


felipebalbi (approved):

espi-service/src/espi_service.rs
```diff
@@ -180,121 +243,125 @@ pub async fn espi_service(mut espi: espi::Espi<'static>, memory_map_buffer: &'st
...
-            }
-            Ok(espi::Event::WireChange(_)) => {
-                info!("eSPI WireChange");
-            }
-            Err(_) => {
-                error!("eSPI Failed");
+
+                espi_service.endpoint.send(acpi_msg.endpoint, &acpi_msg).await.unwrap();
+                info!("MCTP packet forwarded to service: {:?}", acpi_msg.endpoint);
+                espi.complete_port(port_event.port).await;
```
jerrysxie, line 326: We should be able to move this before the `send()`. At that point, everything we need is already copied out of the espi oob rx buffer.

jerrysxie: @felipebalbi looking at `espi.compelte_port()`, it is just clearing interrupts. There seems to be no reason to make it async. Should we change it to non-async?

espi-service/src/espi_service.rs
```diff
@@ -180,121 +243,125 @@ pub async fn espi_service(mut espi: espi::Espi<'static>, memory_map_buffer: &'st
...
+            info!("eSPI WireChange");
+        }
+        Err(e) => {
+            error!("eSPI Failed with error: {:?}", e);
+        }
+    }
+}
+
+async fn send_mctp_error_response(espi: &mut espi::Espi<'_>, port_id: usize) {
+    // SAFETY: Unwrap is safe here as battery will always be supported.
```
jerrysxie, line 344: This is just temporary, right? Not sure if we can claim battery service will always be instantiated on a system.

jerrysxie (changes requested):

espi-service/src/espi_service.rs
tullom: Yes, i need to follow up with @philgweber to figure out what id to put here long term. But for now this is good.

embedded-service/src/ec_type/message.rs
```diff
@@ -75,6 +75,8 @@ pub struct AcpiMsgComms<'a> {
     pub payload_len: usize,
     /// Endpoint ID
     pub endpoint: crate::comms::EndpointID,
```
jerrysxie, line 77: Look more closely at this type today with @kat-perez, do we need the `endpoint` field? We can get it from the comms message already through: https://github.com/OpenDevicePartnership/embedded-services/blob/e765cd74d404ca3abf0dad1cbf7ad88b7353053e/embedded-service/src/comms.rs#L156

And the `port_id`, that seems very specific to this IMXRT OOB implementation, can we just hardcoded it for it? I don't think we need to include it in this generic message wrapper type.

jerrysxie (changes requested):

jerrysxie (approved):

felipebalbi (approved):

thdesil (approved):

PR #465, opened 2025-08-13
Add other notifications event and add port command to get Vdms

zhuyi2024:
Tested with Dock3
13:55:30.197 : 689509.277065 INFO Port0: Got custom mode other VDM received: OtherVdm { data: [34, 67, 160, 94, 4, 1, 1, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }
13:55:30.229 : 689509.298622 INFO PD ControllerId(0): Custom mode entered on port PortId(0): OtherVdm { data: [65, 68, 161, 94, 4, 1, 1, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }  -- Get VDM data on Surface side
13:55:30.229 : 689509.306154 INFO Port0: Got custom mode entered: OtherVdm { data: [65, 68, 161, 94, 4, 1, 1, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }
13:55:30.229 : 689509.310239 INFO Port0: Got custom mode other VDM received: OtherVdm { data: [65, 68, 161, 94, 4, 1, 1, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -495,7 +495,7 @@ checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 [[package]]
 name = "embassy-embedded-hal"
 version = "0.4.0"
-source = "git+https://github.com/embassy-rs/embassy#f2be66a5f94a655696407e2e7bc81c322aab3ea1"
+source = "git+https://github.com/embassy-rs/embassy#63ade389303ab4f6fe9d34211629b02c16e9bb11"
 dependencies = [
  "embassy-futures",
  "embassy-hal-internal",
@@ -512,7 +512,7 @@ dependencies = [
 [[package]]
 name = "embassy-executor"
 version = "0.8.0"
-source = "git+https://github.com/embassy-rs/embassy#f2be66a5f94a655696407e2e7bc81c322aab3ea1"
+source = "git+https://github.com/embassy-rs/embassy#63ade389303ab4f6fe9d34211629b02c16e9bb11"
 dependencies = [
  "cortex-m",
  "critical-section",
@@ -525,7 +525,7 @@ dependencies = [
 [[package]]
 name = "embassy-executor-macros"
 version = "0.7.0"
-source = "git+https://github.com/embassy-rs/embassy#f2be66a5f94a655696407e2e7bc81c322aab3ea1"
+source = "git+https://github.com/embassy-rs/embassy#63ade389303ab4f6fe9d34211629b02c16e9bb11"
 dependencies = [
  "darling",
  "proc-macro2",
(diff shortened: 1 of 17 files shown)


type-c-service/src/wrapper/mod.rs
```diff
@@ -69,6 +69,8 @@ pub enum Event<'a> {
     PortStatusChanged(LocalPortId, PortStatusChanged),
     /// PD alert
     PdAlert(LocalPortId, Ado),
+    /// Other port notification
+    OtherPortNotification(LocalPortId, PortNotificationSingle),
```
RobertZ2011, line 73: For VDMs I would instead create a VDM specific variant. Something like `Vdm(LocalPortId, VdmData)` with `VdmData::Other` and `VdmData::Attn` variants. Then you can call into any VDM-related functionality from within the event-loop when you see `Event::Vdm`.

RobertZ2011 (changes requested):

RobertZ2011 (changes requested):

type-c-service/src/wrapper/mod.rs
asasine: I started working on this and there's a correctness problem with adding variants dedicated to VDM, either in a shared `Vdm` variant or top-level `CustomModeFoo` variants. Since `OtherPortNotification` has a `PortNotificationSingle`, which itself contains `CustomModeFoo` variants, it becomes ambiguous to the service from which variant it will receive VDM events. We can't remove the VDM variants from `PortNotificationSingle` as that's the same type used by the controller wrapper to process port notifications, which comes from real IRQs from the device driver.

The solution is to omit the VDM-specific variants, but given that the service doesn't use the same wait/process/finalize design as the controller, you lose the ability to pass the VDM data to the caller. A reasonable workaround is to expose the `context.get_other/attn_vdm` methods on the `Service` and let the caller handle those variants by retrieving the VDMs themselves. That workaround can be removed if we change the controller --> service data path to include the VDMs themselves, rather than having the service read them again after the controller already read them.

type-c-service/src/service/mod.rs
```diff
@@ -206,10 +216,54 @@ impl<'a> Service<'a> {
                                         continue;
                                     }
                                 }
+                                PortNotificationSingle::CustomModeEntered => {
+                                    match self.context.get_other_vdm(port_id).await {
```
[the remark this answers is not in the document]
asasine: See https://github.com/OpenDevicePartnership/embedded-services/pull/465#discussion_r2294649831

No longer reading VDMs (or PD Alerts) on port events so this is no longer relevant. Callers can retrieve those explicitly with `service.get_pd_alert(port)` (pulls from buffered content) or `service.get_other/attn_vdm(port)` (reads register directly) per their own needs.

williampMSFT (approved):

PR #474, opened 2025-08-21
espi-service: Support multiple OOB response packets

tullom:


battery-service/src/acpi.rs
```diff
@@ -45,14 +45,17 @@ impl<'a> Payload<'a> {
...
-
         buf[0] = self.version;
         buf[1] = self.instance;
         buf[2] = self.reserved;
         buf[3] = self.command as u8;
+
+        if buf.len() < self.data.len() + 4 {
+            // 1 in the reserved field is an error to the host
+            buf[2] = 1;
+            return Err(PayloadError::BufTooSmall);
```
williampMSFT, line 56: It looks a bit odd to me that the client is expected to use the buffer on an error and just know how many bytes to use - is this the right signature for this function, or would it make sense to have it return a different error type, perhaps with an attached number of bytes?

kurtjd (dismissed):

battery-service/src/acpi.rs
tullom: Revisited this and added a response length field to the return type. Indeed, it leads to less magic numbers littered around the code and the intent behind the number is better conveyed.

thdesil (approved):

williampMSFT (approved):

PR #480, opened 2025-08-29
Connect up more DisplayPort and USB config to tps6699x

bramsdell-ms:
These changes expose the tps6699x DisplayPort and USB enable/disable functionality up through the controller and wrapper.


type-c-service/src/driver/tps6699x.rs
```diff
@@ -189,6 +190,71 @@ bitfield! {
...
+    /// Corresponds to ExtPDAltDpConfig_t in C
+    #[derive(Clone, Copy, Debug)]
+    pub struct PdDpAltConfig(u32);
+
+    /// Select configuration (2 bits)
+    pub u8, select_config, set_select_config: 1, 0;
+    /// Signaling (4 bits)
+    pub u8, signaling, set_signaling: 5, 2;
+    /// Reserved field 0 (2 bits)
+    pub u8, reserved0, set_reserved0: 7, 6;
```
RobertZ2011, line 251: Don't need the reserved field.

type-c-service/src/driver/tps6699x.rs
```diff
@@ -189,6 +190,71 @@ bitfield! {
     pub u8, version, set_version: 31, 29;
 }
 
+bitflags! {
```
RobertZ2011, line 193: Is this something that can go in either `tps6699x` or `embedded-usb-pd`? But I'm fine if we want to land it here and move it over later.

RobertZ2011 (changes requested):

asasine (approved):

RobertZ2011 (approved):

PR #485, opened 2025-09-03
Properly scope eSPI borrow of response payload

kurtjd:
The eSPI service `process_response_to_host` method was holding the payload borrow across await points, resulting in, for example, the debug service to panic as it attempted to borrow the same buffer mutably while this method was still holding the immutable borrow.

This PR just scopes the borrow so that it is dropped when it is no longer needed before the await points are hit.

### DIFF
--- a/espi-service/src/espi_service.rs
+++ b/espi-service/src/espi_service.rs
@@ -159,62 +159,65 @@ impl Service<'_, '_> {
         acpi_response: &AcpiMsgComms<'_>,
         endpoint: EndpointID,
     ) {
-        let response_len = acpi_response.payload_len;
-        let num_pkts = ((acpi_response.payload_len - 1) / 64) + 1;
-        let access = acpi_response.payload.borrow();
-        let acpi_pkt: &[u8] = access.borrow();
+        // Scoped so payload borrow is dropped before await point
+        {
+            let response_len = acpi_response.payload_len;
+            let num_pkts = ((acpi_response.payload_len - 1) / 64) + 1;
+            let access = acpi_response.payload.borrow();
+            let acpi_pkt: &[u8] = access.borrow();
 
-        for i in 0..num_pkts {
-            let pkt_start_index = 64 * i;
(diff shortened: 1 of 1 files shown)


espi-service/src/espi_service.rs
```diff
@@ -159,62 +159,65 @@ impl Service<'_, '_> {
...
-
-                if res.is_err() {
-                    #[cfg(feature = "defmt")]
-                    error!("eSPI OOB write failed: {}", res.err().unwrap());
+                } else {
+                    // Packet malformed, throw it away and respond with ACPI packet with error in reserved field.
+                    error!("Error building MCTP response packet from service {:?}", endpoint);
+                    send_mctp_error_response(espi, OOB_PORT_ID);
                     return;
                 }
```
jerrysxie, line 220: The logic of this for loop has been altered. It is intending to send a OOB TX packet and then wait for it to complete, and then write the next OOB TX packet into the buffer. But now we are only waiting for OOB TX packet complete event at the end. The OOB TX packet in between will be overwritten.

If we want to drop the access to the payload, I think we have to borrow, call `build_mctp_header()`, then drop before calling `wait_for_event()` every for loop interation.

jerrysxie (changes requested):

espi-service/src/espi_service.rs
kurtjd: Hm now thinking about it, panicking probably should be expected behavior. If another task attempts to mutably borrow the same buffer that the eSPI service is in the process of sending out I think that should be a bug.

Looks like the debug service at some points will send a payload that can't fit in one packet (so the eSPI service breaks it into chunks) but then then debug service immediately attempts to clear the buffer (hence the mutable borrow) which I think is a bug and panicking should be expected.

kurtjd: This is hard to control though since the buffer isn't async capable. Might need to rename `borrow_mut` to `try_borrow_mut` and turn `borrow_mut` into `async borrow_mut` that waits until all immutable borrows are dropped before waking.

PR #492, opened 2025-09-08
type-c-service/tps6699x: Remove redundant mutexes and flatten async

RobertZ2011:
Only async functions that had a positive effect on code size were flattened.

### DIFF
--- a/embedded-service/src/type_c/controller.rs
+++ b/embedded-service/src/type_c/controller.rs
@@ -568,7 +568,7 @@ pub trait Controller {
     // TODO: remove all these once we migrate to a generic FW update trait
     // https://github.com/OpenDevicePartnership/embedded-services/issues/242
     /// Get current FW version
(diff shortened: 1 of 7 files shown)


embedded-service/src/type_c/controller.rs
```diff
@@ -568,7 +568,7 @@ pub trait Controller {
     // TODO: remove all these once we migrate to a generic FW update trait
     // https://github.com/OpenDevicePartnership/embedded-services/issues/242
     /// Get current FW version
-    fn get_active_fw_version(&self) -> impl Future<Output = Result<u32, Error<Self::BusError>>>;
+    fn get_active_fw_version(&mut self) -> impl Future<Output = Result<u32, Error<Self::BusError>>>;
```
jerrysxie, line 571: This is API breaking, but looking at its usage, as far as I can tell it should be fine because it is already being called on a `&mut`.

jerrysxie (approved):

PR #500, opened 2025-09-23
Host-EC demonstration code

tullom:
- AcpiMsgComms changed to HostRequest, since not all host messages are ACPI messages
- Moved serde of the MCTP message header from the subsystems to host service
- Baked command and status into HostRequest, standardizing Host message requests
- Moved MCTP message serde to mctp-rs library
- Commented out notification support in debug service while we wait for the host handling to stabilize. Host will poll for debug messages for now. (Thanks Kurtis!)
- Added no availibility debug message type to return when Host polls but no debug message is available. (Thanks Kurtis!)

### DIFF
--- a/espi-service/src/espi_service.rs
+++ b/espi-service/src/espi_service.rs
@@ -226,6 +226,10 @@ impl Service<'_> {
                 error!("serialize_packet_from_subsystem: {:?}", e);
                 Error::Serialize
             })?;
+
+            // Immediately service the packet with the ESPI HAL
+            let event = espi.wait_for_event().await;
+            process_controller_event(espi, self, event).await;
         }
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -73,3 +73,4 @@ tps6699x = { git = "https://github.com/OpenDevicePartnership/tps6699x" }
 tokio = { version = "1.42.0" }
 zerocopy = "0.8.26"
 uuid = { version = "=1.17.0", default-features = false }
+mctp-rs = { git = "https://github.com/tullom/mctp-rs", branch = "make-interfaces-public" }
```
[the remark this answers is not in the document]
tullom: Yeah that is correct. @dymk currently hosts mctp-rs in his personal account but the goal is to move that repo to ODP. In the meantime, i use a branch of the mctp library that hosts some fixes for the demo.

espi-service/src/espi_service.rs
```diff
@@ -156,64 +160,64 @@ impl Service<'_, '_> {
...
-                        #[cfg(feature = "defmt")]
-                        error!("Failed to retrieve OOB write buffer: {}", _e);
-                        // TODO: Ask if we need to send a response if the request is malformed
-                        return;
-                    }
+        let mut packet_state = mctp_ctx
+            .serialize_packet(reply_context, &acpi_pkt[..response_len])
+            .unwrap();
+        // Send each packet
+        while let Some(packet_result) = packet_state.next() {
```
felipebalbi, line 194: I take it `packet_state` is an iterator? But it seems it's an iterator of `Result` values, that smells a bit odd. Could you explain here?

espi-service/src/espi_service.rs
```diff
@@ -156,64 +160,64 @@ impl Service<'_, '_> {
...
+            let packet = packet_result.unwrap();
```
felipebalbi, line 195: If `packet_result` is a `None`, this will `panic!()`. Probably a good idea to avoid `panic!()`s

kurtjd (approved):

PR #511, opened 2025-10-15
Direct async calls proof of concept

RobertZ2011:
Proof of concept for calling async trait functions directly. This example introduces two device traits (`power::Device` and `io_expander::Device`) - which are basic traits for a power policy device and an IO expander device, two simple service implementations that use these traits, and a mock device that implements both device traits. Both service implementations directly call async trait functions without using device IDs or messaging as intermediates.

### Notes
* Events move from the device to each service through per-device channels. This is an implementation detail and everything should be generic over sender and receiver traits.
* The example does not use `StaticCell` and everything is declared with normal `let _ = ...` statements. This is only possible if everything is in the same embassy task. Using multiple embassy tasks will require `'static`.
* The use of an external object for device -> service events breaks circular references between the two. It is possible to have these references so that the device could call directly into the service (e.g. to notify of an event) but this requires that both have the same lifetime. The only lifetime that accomplishes this is `'static`.
* The service implementations have a maximum number of devices that they can support. This is because we need space to store the futures when we select among the event receivers. In theory we could work around this with a channel but this results in a situation where the event has to contain a reference to the device (so we know which device it came from). But the device also has to have a reference to the channel in order to send an event. This results a circular reference that requires `'static` to work. Additionally this requires the use of `dyn` to break a recursive type that arises between the channel and the device.
* Since we only accept a single device/receiver type there's a corresponding `EventSender` trait. This is to ensure that device implementations don't come up with their own bespoke `EventReceiver` types which would all be incompatible with each other.
* This example doesn't attempt to do any rigorous enforcement of initialization order at compile time, outside of what is provided by using variables. But it should be amenable to the various approaches we've discussed.

### DIFF
--- a/examples/std/src/bin/direct-async-calls/io_expander.rs
+++ b/examples/std/src/bin/direct-async-calls/io_expander.rs
@@ -1,0 +1,104 @@
+use std::{future::Future, pin::pin};
+
+use embassy_futures::select::select_slice;
+use embassy_sync::{
+    blocking_mutex::raw::NoopRawMutex,
+    mutex::Mutex,
+    pubsub::{DynImmediatePublisher, DynSubscriber, WaitResult},
+};
+use log::{info, warn};
+
+pub trait EventReceiver {
+    fn wait_next(&mut self) -> impl Future<Output = InterruptEvent>;
+}
+
+pub trait EventSender {
+    fn on_interrupt(&self, pin: u8, level: bool);
+}
+
+/// IO expander device trait
+pub trait Device {
+    fn name(&self) -> &str;
+
+    fn set_level(&mut self, pin: u8, value: bool) -> impl Future<Output = ()>;
+}
+
+pub struct Sender<'channel> {
+    publisher: DynImmediatePublisher<'channel, InterruptEvent>,
+}
+
+impl<'channel> Sender<'channel> {
+    pub fn new(publisher: DynImmediatePublisher<'channel, InterruptEvent>) -> Self {
+        Self { publisher }
+    }
+}
+
+pub struct Receiver<'channel> {
+    subscriber: DynSubscriber<'channel, InterruptEvent>,
+}
+
+impl<'channel> Receiver<'channel> {
+    pub fn new(subscriber: DynSubscriber<'channel, InterruptEvent>) -> Self {
+        Self { subscriber }
+    }
+}
+
+impl EventReceiver for Receiver<'_> {
+    async fn wait_next(&mut self) -> InterruptEvent {
+        loop {
+            match self.subscriber.next_message().await {
+                WaitResult::Message(msg) => return msg,
+                WaitResult::Lagged(n) => {
+                    warn!("Receiver lagged by {n} messages");
+                }
+            }
+        }
+    }
+}
+
+impl EventSender for Sender<'_> {
+    fn on_interrupt(&self, pin: u8, level: bool) {
+        self.publisher.publish_immediate(InterruptEvent { pin, level });
+    }
+}
+
+#[derive(Copy, Clone)]
+pub struct InterruptEvent {
+    pub pin: u8,
+    pub level: bool,
+}
+
+const MAX_SUPPORTED_DEVICES: usize = 4;
+
+pub struct ServiceImplementation<'storage, 'device, D: Device, R: EventReceiver> {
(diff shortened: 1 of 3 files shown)


examples/std/src/bin/direct-async-calls/io_expander.rs
```diff
@@ -0,0 +1,104 @@
...
+    pubsub::{DynImmediatePublisher, DynSubscriber, WaitResult},
+};
+use log::{info, warn};
+
+pub trait EventReceiver {
+    fn wait_next(&mut self) -> impl Future<Output = InterruptEvent>;
+}
+
+pub trait EventSender {
+    fn on_interrupt(&self, pin: u8, level: bool);
```
RobertZ2011: Should this take `&mut self`? Right now this supports mpsc but I'm not sure if we want support that since this is a dedicated channel between the service and the device.

examples/std/src/bin/direct-async-calls/io_expander.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/// IO expander device trait
+pub trait Device {
+    fn name(&self) -> &str;
+
+    fn set_level(&mut self, pin: u8, value: bool) -> impl Future<Output = ()>;
+}
+
+pub struct Sender<'channel> {
```
RobertZ2011: I think the event traits could be made more generic to make these implementations generic over the message type.

examples/std/src/bin/direct-async-calls/io_expander.rs
```diff
@@ -0,0 +1,104 @@
...
+impl<'channel> Receiver<'channel> {
+    pub fn new(subscriber: DynSubscriber<'channel, InterruptEvent>) -> Self {
+        Self { subscriber }
+    }
+}
+
+impl EventReceiver for Receiver<'_> {
+    async fn wait_next(&mut self) -> InterruptEvent {
+        loop {
+            match self.subscriber.next_message().await {
```
RobertZ2011: I think backpressure needs to be handled by the device implementation instead. I'm thinking that sender traits should expose `on_x` and `try_on_x` functions to support this.

examples/std/src/bin/direct-async-calls/power.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn new(devices: &'storage mut [(R, &'device Mutex<NoopRawMutex, D>)]) -> Self {
+        Self {
+            devices,
+            current_connection: None,
+        }
+    }
+
+    pub async fn wait_next(&mut self) -> (&'device Mutex<NoopRawMutex, D>, Event) {
+        let futures =
+            heapless::Vec::<_, MAX_SUPPORTED_DEVICES>::from_iter(self.devices.iter_mut().map(|(r, _)| r.wait_next()));
```
RobertZ2011: Futures can end up arbitrarily large so I'm worried that this could blow the stack and we wouldn't know until runtime. Thoughts on possibly introducing a macro/wrapper type that uses `size_of` to check the total size of the vec at compile time?

examples/std/src/bin/direct-async-calls/io_expander.rs
asasine: I considered that but to what end? The generic service impl would just no-op? The only purpose I could foresee would be vendor-defined messaging, but in that case, should the vendor simply add a side channel to their own types that implements the sender and receiver traits?

examples/std/src/bin/direct-async-calls/power.rs
asasine: Is this a different risk than when calling any other generic future?

asasine (commented):
Looking good so far!

What do you think of expanding this example to showcase the vendor customization process? I think we can essentially bifurcate into an ODP module and a vendor module. Provide the traits, service impls, and any "default" impls in the ODP mod, provide new types that reuse what's possible but extend behavior in the vendor mod, and a main function that uses the ODP tasks with the vendor types.

examples/std/src/bin/direct-async-calls/io_expander.rs
RobertZ2011: Yup, I mean that `EventSender` should be the one to consider backpressure instead of having to handle it here.

PR #551, opened 2025-10-30
Add DiscoverModeCompleted processor

zhuyi2024:
1. Update tps6699x for gcdm command
2. Add get_rx_disc_mode_vdos method to call gcdm to get rx discover mode vdos
3. Add process_disc_mode_completed_event and finalize_disc_mode_completed methods to handle DiscoverModeCompleted notification


type-c-service/src/wrapper/mod.rs
```diff
@@ -55,6 +54,9 @@ pub const DEFAULT_FW_UPDATE_TICK_INTERVAL_MS: u64 = 5000;
 /// 300 seconds at 5 seconds per tick
 pub const DEFAULT_FW_UPDATE_TIMEOUT_TICKS: u8 = 60;
 
+/// Microsoft SVID
```
RobertZ2011: I would introduce a `Config` struct, with an `Option<Svid>` field, to pass into the `ControllerWrapper` constructor.

PR #557, opened 2025-11-11
Add replace() and update() methods on ThreadModeCell

Ctru14:
ThreadModeCell object was missing some useful wrappers for other internal Cell methods. This added replace() and update() methods to the ThreadModeCell, leveraging the inner Cell's implementation.

Verified on RT633 platform.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -167,7 +167,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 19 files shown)


RobertZ2011 (approved):

kurtjd:
I had considered these before but reverted since they are not currently supported by our MSRV (as seen by the failing checks). I don't see any issue to bumping MSRV to 1.88 from 1.85 to support these though.

PR #589, opened 2025-11-17
power-policy-service: Remove global unconstrained power threshold

RobertZ2011:
This is something that is better as a per-device config value.

Breaking due to introduction of new config struct, migration can be done by using `Default::default` or providing a bespoke config value.

### DIFF
--- a/type-c-service/src/wrapper/config.rs
+++ b/type-c-service/src/wrapper/config.rs
@@ -7,14 +7,14 @@ pub struct Config {
 }
 
 /// Unconstrained behavior for sink role
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum UnconstrainedSink {
     /// Automatically signal unconstrained power based on unconstrained bit in PDO
     #[default]
     Auto,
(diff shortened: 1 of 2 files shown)


felipebalbi (dismissed):

tullom (dismissed):

type-c-service/src/wrapper/config.rs
```diff
@@ -0,0 +1,20 @@
...
+
+/// Unconstrained behavior for sink role
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum UnconstrainedSink {
+    /// Automatically signal unconstrained power based on unconstrained bit in PDO
+    #[default]
+    Auto,
+    /// Automatically signal unconstrained power for any sink that meets a power threshold in mW
+    PowerThresholdMw(u32),
```
asasine: Is there a power unit type (e.g., `Milliwatt` or `Watt`)?

asasine (approved):
nb

type-c-service/src/wrapper/config.rs
RobertZ2011: Not currently, I think that should be part of a general pass over units. For the moment I've written out milliwatts here because the alternatives are awkward.

asasine (approved):

jerrysxie (commented):
@RobertZ2011 Please announce on Zulip and ensure clients are ready to uptake this change.

PR #592, opened 2025-11-19
Remove unnecessary `OnceLock` and `async` usage

jamesmunns:
These changes switch from using a `OnceLock`, which requires access is async, which then allows for many functions to no longer be async at all.

This is intended to allow a reduction in code size.

### DIFF
--- a/battery-service/src/lib.rs
+++ b/battery-service/src/lib.rs
@@ -104,9 +104,7 @@ static SERVICE: Service = Service::new();
 ///
 /// Must be done before sending the battery service commands so that hardware device is visible
 /// to the battery service.
-pub fn register_fuel_gauge(
-    device: &'static device::Device,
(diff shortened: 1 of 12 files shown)


battery-service/src/lib.rs
```diff
@@ -99,68 +98,54 @@ impl comms::MailboxDelegate for Service {
...
-    service.endpoint.send(endpoint_id, data).await
+    SERVICE.endpoint.send(endpoint_id, data).await
 }
 
 /// Send the battery service state machine an event and await a response.
 ///
 /// This is an alternative method of interacting with the battery service (instead of using the comms service),
 /// and is a useful fn if you want to send an event and await a response sequentially.
 pub async fn execute_event(event: BatteryEvent) -> context::BatteryResponse {
-    let service = SERVICE.get().await;
```
tullom: Blocking behaviour might be expected

PR #627, opened 2025-12-09
power-policy: Add PSU type to capabilities flags

RobertZ2011:
Also update events to use full power capability structs.

### Breaking
Breaking due to replacing usages of `PowerCapability` with `ConsumerPowerCapability`. The former is still accessible as a public member within the latter; migration is to use that public member.

### DIFF
--- a/embedded-service/src/power/policy/flags.rs
+++ b/embedded-service/src/power/policy/flags.rs
@@ -1,9 +1,11 @@
 //! Consumer and provider flags, these are used to signal additional information about a consumer/provider request
 
 use bitfield::bitfield;
+use num_enum::{IntoPrimitive, TryFromPrimitive};
 
 /// PSU type
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, IntoPrimitive, TryFromPrimitive)]
+#[num_enum(error_type(name = InvalidPsuType, constructor = InvalidPsuType))]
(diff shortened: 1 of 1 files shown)


asasine (approved):

embedded-service/src/power/policy/flags.rs
```diff
@@ -2,14 +2,60 @@
...
+            1 => Self::TypeC,
+            2 => Self::DcJack,
+            12 => Self::Custom0,
+            13 => Self::Custom1,
+            14 => Self::Custom2,
+            15 => Self::Custom3,
+            _ => Self::Unknown,
+        }
+    }
+}
```
felipebalbi: perhaps a `TryFrom` would be more suitable?

embedded-service/src/power/policy/flags.rs
```diff
@@ -2,14 +2,68 @@
...
+            value => Err(InvalidPsuType(value)),
+        }
+    }
+}
+
+impl From<PsuType> for u8 {
+    fn from(value: PsuType) -> Self {
+        value as u8
+    }
+}
```
williampMSFT: FYI - There's a crate called num_enum that we're already using in this repo that can generate this implementation for you if you `#[derive(num_enum::IntoPrimitive, num_enum::TryFromPrimitive)]` the enum

RobertZ2011: Thanks for the tip!

kurtjd (approved):

seanbrns (approved):

PR #639, opened 2025-12-15
battery-service: minimize panic code paths

jerrysxie:
This pull request refactors error handling and buffer copying logic in the battery service code to make it more robust and explicit. The main improvements are in the way slices are copied into fixed-size buffers and how errors are handled and propagated in the state machine.

Buffer copying improvements:

* Refactored the buffer copying logic in `compute_bix` (`acpi.rs`) to use safer and more explicit slice operations with `get_mut` and `ok_or`, replacing the previous use of `try_into` and direct indexing. This reduces the chance of panics and makes error handling clearer.

Error handling and state machine robustness:

* Updated the state machine timeout handling in `Context` (`context.rs`) to properly propagate errors: now, if the state machine returns an error, it sends a `StateError` instead of assuming infallibility.
* Changed the handling of `BatteryEventInner::Oem` events to return a specific `InvalidActionInState` error instead of an unimplemented panic (`todo!()`), making the state machine more robust and explicit about unsupported actions.

### DIFF
--- a/battery-service/src/acpi.rs
+++ b/battery-service/src/acpi.rs
@@ -135,12 +135,12 @@ pub(crate) fn compute_bix<'a>(
             battery_swapping_capability: embedded_batteries_async::acpi::BatterySwapCapability::NonSwappable,
         };
 
-    let moduel_number_len = core::cmp::min(STD_BIX_MODEL_SIZE - 1, static_cache.device_name.len() - 1);
+    let model_number_len = core::cmp::min(STD_BIX_MODEL_SIZE - 1, static_cache.device_name.len() - 1);
     bix_return
         .model_number
-        .get_mut(..moduel_number_len)
+        .get_mut(..model_number_len)
         .ok_or(())?
-        .copy_from_slice(static_cache.device_name.get(..moduel_number_len).ok_or(())?);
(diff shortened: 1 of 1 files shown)


RobertZ2011 (dismissed):

kurtjd (approved):

battery-service/src/acpi.rs
```diff
@@ -134,27 +134,36 @@ pub(crate) fn compute_bix<'a>(
             oem_info: [0u8; STD_BIX_OEM_SIZE],
             battery_swapping_capability: embedded_batteries_async::acpi::BatterySwapCapability::NonSwappable,
         };
+
+    #[allow(clippy::indexing_slicing)] // Slicing is guarded by min functions
```
bramsdell-ms: I'm not sure that the suppressing of the warning is warranted in these places. That is, I understand here that you know the defined length of `model_number` and `STD_BIX_MODEL_SIZE` and `device_name` are all in sync, but in the future they have a possibility of being out of sync. The function is already returning `Result`, so my preference would be to still use `copy_from_slice` (which can panic, but you can see both slices are clearly the same length right here), and then use `get` for the slices and pop an `Err` out if you get `None`.

williampMSFT (dismissed):
:shipit:

danirobinson-odp (approved):

PR #641, opened 2025-12-16
hid-service: Depanic

RobertZ2011:
### Breaking
Introduces a struct to with named fields for indexing range errors. Interrupt passthrough processing function now returns a result. Migration is to move over to the named indexing error struct and properly handle the interrupt error if using passthrough.

### DIFF
--- a/battery-service/src/context.rs
+++ b/battery-service/src/context.rs
@@ -529,7 +529,7 @@ impl Context {
             embedded_services::power::policy::CommsData::ConsumerConnected(_device_id, power_capability) => {
                 *psu_state = PsuState {
                     psu_connected: true,
-                    power_capability: Some(*power_capability),
+                    power_capability: Some(power_capability.capability),
                 }
             }
             _rest => { /* Don't care about anything else */ }
--- a/embedded-service/Cargo.toml
+++ b/embedded-service/Cargo.toml
@@ -51,7 +51,7 @@ cortex-m.workspace = true
 critical-section = { workspace = true, features = ["std"] }
 embassy-futures.workspace = true
 embassy-sync = { workspace = true, features = ["std"] }
-embassy-time = { workspace = true, features = ["std"] }
+embassy-time = { workspace = true, features = ["std", "generic-queue-8"] }
 embassy-time-driver = { workspace = true }
 rstest.workspace = true
 static_cell.workspace = true
--- a/embedded-service/src/power/policy/charger.rs
+++ b/embedded-service/src/power/policy/charger.rs
@@ -3,7 +3,10 @@ use core::{future::Future, ops::DerefMut};
 
 use embassy_sync::{channel::Channel, mutex::Mutex};
 
-use crate::{GlobalRawMutex, intrusive_list, power};
+use crate::{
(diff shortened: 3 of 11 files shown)


kurtjd (approved):

williampMSFT (approved):

hid-service/src/i2c/device.rs
```diff
@@ -78,11 +84,23 @@ impl<A: AddressMode + Copy, B: I2c<A>> Device<A, B> {
...
         let mut borrow = self.buffer.borrow_mut();
         let buf: &mut [u8] = borrow.borrow_mut();
+        let buffer_len = buf.len();
         let reg = desc.w_report_desc_register.to_le_bytes();
         let len = desc.w_report_desc_length as usize;
 
         let mut bus = self.bus.lock().await;
-        if let Err(e) = bus.write_read(self.address, &reg, &mut buf[0..len]).await {
+        if let Err(e) = bus
+            .write_read(
```
jerrysxie: Non-blocking for this PR: I just realized that these fallible i2c transactions are not wrapped with `with_timeout()`, we probably should in case the i2c transaction gets into a bad state and hangs indefinitely.

hid-service/src/i2c/host.rs
```diff
@@ -145,18 +161,37 @@ impl<B: I2cSlaveAsync> Host<B> {
...
+                    DATA_READ_TIMEOUT_MS,
+                    buffer
+                        .get_mut(2..length as usize)
+                        .ok_or(Error::Hid(hid::Error::InvalidSize(InvalidSizeError {
+                            expected: length as usize,
+                            actual: buffer_len,
+                        })))?,
+                )
+                .await?;
                 Some(self.buffer.reference().slice(2..length as usize))
```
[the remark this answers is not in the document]
RobertZ2011: This will be addressed by a future refactoring.

hid-service/src/i2c/device.rs
RobertZ2011: Created a bug to track this: https://github.com/OpenDevicePartnership/embedded-services/issues/652

RobertZ2011:
> > What testing have we done with these changes? Do we have simulated integration test for HID?
> 
> This breaking changes should be minimal impact to our clients.

There should be no behavioral changes except for no more panics in this code. I tested these changes against a HID keyboard implementation and saw no differences.

hid-service/src/i2c/host.rs
```diff
@@ -43,56 +43,74 @@ impl<B: I2cSlaveAsync> Host<B> {
...
-            error!("Response timeout");
-            return Err(Error::Hid(hid::Error::Timeout));
-        }
-
-        if let Err(e) = result.unwrap() {
-            error!("Failed to rwrite to bus");
-            return Err(Error::Bus(e));
-        }
-
-        trace!("Response sent");
```
wmmc88: just fyi this trace is now gone

wmmc88 (approved):

RobertZ2011:
Holding off on merging for the moment to allow some downstream blockages to clear.

danirobinson-odp (approved):

PR #662, opened 2025-12-23
buffer: Audit for panics and refactor

### LINKED ISSUE #659: Audit buffer

kurtjd:
Removed panic paths from the `buffer` module. Instead made methods fallible and introduced an `Error` type.

Because certain traits are used with methods that assume infallibility (such as the drop and borrow traits), I couldn't just make these return a result. Instead, I added a `Poisoned` status so in the off chance that a `drop` or `borrow` call would have panicked originally, the buffer is marked poisoned and will return a `Poisoned` error on future accesses.

This also breaks the interfaces of a few services. For these I added a `Buffer` error and made methods fallible if necessary. I'm not a fan of this since any buffer errors represent an internal programming bug to the service and not something the user should expect/handle, and this breaks encapsulation. I do not want to silently just ignore buffer errors, so I'm not sure what other alternative there is here (open to suggestions).

*Edit*: Thinking about maybe renaming the `Buffer` error variant of these services to `Internal` or something. I'd prefer to make it more clear when an error represents an internal invariant being broken that is not something meant to be handled by the caller in any practical manner. It would act as a catchall since the specifics of what caused the error are implementation details that could change any time and not something the caller should be concerned with.

Also introduced a `Never` type alias which is useful now that some functions represent never-ending tasks which should never return unless there is an error (see related discussion: https://github.com/OpenDevicePartnership/embedded-services/pull/634#discussion_r2615419765).

Resolves #659

### DIFF
--- a/embedded-service/src/buffer.rs
+++ b/embedded-service/src/buffer.rs
@@ -18,11 +18,26 @@ use core::ops::Range;
 use crate::SyncCell;
 use core::sync::atomic::AtomicPtr;
 
+/// Buffer error.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
+    /// Buffer already borrowed immutably.
+    BorrowedImmutably,
+    /// Buffer already borrowed mutably.
+    BorrowedMutably,
+    /// Range is out-of-bounds.
+    InvalidRange,
+    /// Buffer is poisoned and should be considered no longer valid.
+    Poisoned,
+}
+
 #[derive(Copy, Clone, PartialEq, Eq)]
 enum Status {
     None,
     Mutable,
     Immutable(u32),
+    Poisoned,
 }
 
 /// Underlying buffer storage struct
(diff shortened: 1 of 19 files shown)


madeleyneVaca (dismissed):

jerrysxie (approved):

jerrysxie:
@kurtjd Don't forget to announce on Zulip once it is ready to be merged, thanks!

jerrysxie (approved):

embedded-service/src/buffer.rs
```diff
@@ -215,6 +263,10 @@ impl<T> Borrow<[T]> for Access<'_, T> {
                 self.buffer.len,
             )
         };
+
+        // Panic safety: The public API prevents a slice from being stored that would
+        // be outside the bounds of the buffer
+        #[allow(clippy::indexing_slicing)]
```
bramsdell-ms: Do you need to suppress this any longer here?

bramsdell-ms: I need better glasses. How about `ok_or` with an empty array?

kurtjd: Heh originally did something like that: https://github.com/OpenDevicePartnership/embedded-services/pull/662#discussion_r2644200692

I don't have a strong opinion either way, but I would like a few more people to chime in before I change it again to avoid the back and forth.

madeleyneVaca (approved):

jerrysxie:
Bypassed policy and merge since people are on vacation.

kurtjd:
Zulip announcement: https://opendevicepartnership.zulipchat.com/#narrow/channel/495405-embedded-controller/topic/embedded-services.20buffer.20breaking.20changes/with/565229931

PR #663, opened 2025-12-23
Add panic lints

jerrysxie:
* Add exceptions for test code and modules with limited dependencies

### DIFF
--- a/type-c-service/src/lib.rs
+++ b/type-c-service/src/lib.rs
@@ -120,6 +120,7 @@ impl PortEventStreamer {
 }
 
 #[cfg(test)]
+#[allow(clippy::unwrap_used)]
 mod tests {
     use core::sync::atomic::AtomicBool;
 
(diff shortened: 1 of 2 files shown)


jerrysxie:
> Could also consider placing this in the `lib.rs` of each crate to blanket allow panics in tests:
> 
> ```rust
> #![cfg_attr(
>     test,
>     allow(
>         clippy::unwrap_used,
>         clippy::expect_used,
>         clippy::unreachable,
>         clippy::unimplemented,
>         clippy::todo,
>         clippy::panic,
>         clippy::panic_in_result_fn,
>         clippy::indexing_slicing
>     )
> )]
> ```
> 
> Though this would also allow panics anywhere in the crate if testing is enabled, so technically different (but since a regular `cargo clippy` would catch them maybe not a big deal). But if the goal is to be as minimally selective as we can in allowing panics even in tests the current way is better.

That is a good alternative. Right now, let's be as minimal as possible in regard to panics but in the future if doing this in the test code becames too much of a pain then we can switch to this approach.

PR #671, opened 2026-01-08
Add temporary exception for bincode in deny.toml (#669)

williampMSFT:
Per Matteo:
I saw a regression in our stable/deny CI job. The `bincode` crate is unmaintained and ⁠[it has just been marked as
such](https://github.com/rustsec/advisory-db/pull/2574).

This PR adds an exception to our deny.toml to unblock our CI for now, but we should try to migrate to another serde crate like postcard soon.

Luckily only 2 of our crates use bincode so migration is straightforward:

-
[⁠embedded-usb-pd](https://github.com/OpenDevicePartnership/embedded-usb-pd)
- [tps6699x](https://github.com/OpenDevicePartnership/tps6699x)

Cherry-pick of commit 66c758d72849e41f8994bdae7a532dd4a3264416


kurtjd (approved):

tullom (changes requested):
Hold on, i think with mergebacks, we don't wish to cherry-pick commits from main back into this branch. We should merge commits so we don't have two commits with the same changes yet different commits.

PR #674, opened 2026-01-12
Add uart-service

### LINKED ISSUE #605: service: Implement uart-service
Similarly to the existing `espi-service`, implement a `uart-service` which listens for MCTP packets received from host over UART and forwards them to the appropriate service.

kurtjd:
Adds a basic uart service which can act a host service when eSPI is not available. Still uses the `SmbusEspiMedium` even though the SMBUS header isn't necessary just to keep code changes on the host side minimal when switching between UART/ESPI.

Did some testing on the IMXRT board and can send/receive MCTP packets from host over a COM port successfully. Might catch additional bugs as I work on getting the ratatui app working over UART and will fix them as they come.

Resolves #605

### DIFF
--- a/uart-service/src/lib.rs
+++ b/uart-service/src/lib.rs
@@ -1,0 +1,173 @@
+//! uart-service
+//!
+//! To keep things consistent with eSPI service, this also uses the `SmbusEspiMedium` (though not
+//! strictly necessary, this helps minimize code changes on the host side when swicthing between
+//! eSPI or UART).
+//!
+//! Revisit: Will also need to consider how to handle notifications (likely need to have user
+//! provide GPIO pin we can use).
+#![no_std]
+
+mod mctp;
+pub mod task;
+
+use crate::mctp::{HostRequest, HostResult, OdpHeader, OdpMessageType, OdpService};
+use core::borrow::BorrowMut;
+use embassy_sync::channel::Channel;
+use embedded_io_async::Read as UartRead;
+use embedded_io_async::Write as UartWrite;
+use embedded_services::GlobalRawMutex;
+use embedded_services::buffer::OwnedRef;
+use embedded_services::comms::{self, Endpoint, EndpointID, External};
+use embedded_services::trace;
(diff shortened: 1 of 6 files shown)


uart-service/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub fn new() -> Result<Self, Error> {
+        Ok(Self {
+            endpoint: Endpoint::uninit(EndpointID::External(External::Host)),
+            host_tx_queue: Channel::new(),
+            assembly_buf_owned_ref: assembly_buf::get_mut()
+                .ok_or(Error::Buffer(embedded_services::buffer::Error::InvalidRange))?,
+        })
+    }
+
+    async fn process_response<T: UartWrite>(&self, uart: &mut T, response: &HostResponseMessage) -> Result<(), Error> {
```
williampMSFT: FYI - I did change the name for some of the result types from "Response" to "Result" to mirror the way the actual value enums were named (previously Request was pretty unambiguous but Response could mean either Result<T, E> or just T depending on the context), so you might get a couple build breaks when this goes in: https://github.com/OpenDevicePartnership/embedded-services/pull/673

kurtjd: Sounds good, thanks for the heads up!

uart-service/src/task.rs
```diff
@@ -0,0 +1,32 @@
+use crate::{Error, Service};
+use embedded_io_async::Read as UartRead;
+use embedded_io_async::Write as UartWrite;
+use embedded_services::comms;
+use embedded_services::error;
+use static_cell::StaticCell;
+
+pub async fn uart_service<T: UartRead + UartWrite>(mut uart: T) -> Result<embedded_services::Never, Error> {
+    static UART_SERVICE: StaticCell<Service> = StaticCell::new();
```
tullom: One of the goals of the v0.2.0 refactor is to move all statics out of service level code, and into app level code. Instead of initializing the static here, could you modify the fn signature to take in a `&'static Service`?

tullom (approved):

PR #678, opened 2026-01-14
[RFC] Zerocopy for serializing/deserializing ODP service messages

kurtjd:
This PR demonstrates using zerocopy for serializing instead of doing it manually for thermal service messages (and can be used for all services if this approach seems okay).

Notes:
- Zerocopy is tricky when it comes to alignment and representations. Talking with Billy and we agreed it's pretty hard to get it working on the `ThermalRequest`/`ThermalResponse` enums themselves, so I derived the appropriate traits for structs which the enums wrapped, then just pattern match over the enum.
- Had to use the `zerocopy::byteorder::U<N>` types to ensure LE representation on the wire (in the very rare case this is running on BE-native hardware) and to force alignment of fields to 1 (since zerocopy does not allow padding and a `#repr(packed)` could cause issues when accessing the fields from Rust code).
- Those `U<N>` types unfortunately don't impl `defmt::Format` so had to manually do it (might try to upstream defmt support)
- The `mctp_rs` trait methods require serializing into a buffer, instead of just returning a `&[u8]`, so because of that, we lose the zero-copy benefit from zerocopy (and are really just using it for its serialization abilities)

The goal is also eventually to have these service message crates be useable from other areas than just the EC (such as host-side secure services).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2016,6 +2016,7 @@ dependencies = [
  "embedded-services",
  "num_enum",
  "uuid",
+ "zerocopy",
 ]
 
 [[package]]
--- a/examples/rt633/Cargo.lock
+++ b/examples/rt633/Cargo.lock
@@ -1531,6 +1531,7 @@ dependencies = [
  "embedded-services",
  "num_enum",
  "uuid",
+ "zerocopy",
 ]
 
 [[package]]
--- a/examples/std/Cargo.lock
+++ b/examples/std/Cargo.lock
@@ -1639,6 +1639,7 @@ dependencies = [
  "embedded-services",
  "num_enum",
  "uuid",
+ "zerocopy",
 ]
 
 [[package]]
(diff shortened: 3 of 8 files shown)


kurtjd:
Though I'm not a huge fan of this approach because since we are copying bytes, we shouldn't need to worry about the underlying memory structure, so all this fiddling with alignment, padding, and endianness is just noise (which would be important if we were going for zero-cost serialization, but the `mctp_rs` crate currently requires us to copy). Though if we modify the `mctp_rs` crate in the future to not copy, then maybe this will be worthwhile

But for the time-being though, ideally there would be some simple crate that is a proc macro or something that does automatically what we were doing manually (copying each field of the struct in series, in LE format, into a packed &mut [u8]). It looks like there are a couple but haven't played around with them much.

Based on the EC tech forum discussion Friday, something like `postcard` might not be acceptable because all though we don't necessarily care about the format on the wire as long as both sides agree, not sure we want to make the guarantee that the other side will be using `postcard` as well, and the packed byte format is the simplest to write a serializer/deserializer for.

PR #709, opened 2026-02-09
Add mock battery to battery-service and example demoing use

tullom:
- Mock battery is feature gated


battery-service/src/lib.rs
```diff
@@ -14,6 +14,8 @@ mod acpi;
 pub mod context;
 pub mod controller;
 pub mod device;
+// #[cfg(feature = "mock")]
```
kurtjd: Is this intended to be commented out? Otherwise looks good!

RobertZ2011 (approved):

PR #710, opened 2026-02-09
thermal-service: Add mocks

### LINKED ISSUE #619: Pico-de-gallo example: thermal-service

kurtjd:
Adds mock sensor and fan to thermal service and updates the thermal `std` example to make use of them (which simplifies the example quite a lot).

Example can be tested with `cargo run --release --bin thermal`.

Resolves #619.


tullom (approved):

tullom:
> Good start, for future expansion, we can consider adding data modeling to the mock so the mock can take a data model to generate different test cases.

I've been thinking of a trait that can encapsulate this type of behaviour and also allow for easy fuzz testing. Will create an issue for this. Created #712 to track

Issue #725, opened 2026-02-20
Migrate uart-service away from comms system


kurtjd:
The uart-service will need to be updated to be compatible with the latest embedded-services comms changes.

However, to avoid temporarily maintaining the legacy path, will wait until #721 #722 and #723 are complete. Then can fully update uart-service and drop the legacy path completely.

PR #732, opened 2026-02-24
comms: Update uart, thermal, and battery to use Relay trait

### LINKED ISSUE #721: Migrate battery service away from comms system
See what was done in #716 for the time-alarm service for an example. This will allow us to write integration tests against the service and allow OEMs to host their own services that relay messages over the same bus as our services.

kurtjd:
Refactors thermal, battery, and uart services to make use of the latest Relay trait for direct async calls so we don't need to go through comms service to interact with host/relay services.

## Thermal
- Removes all statics
- Removes all uses of intrusive list and instead just uses slices for maintaining lists of sensors and fans
- Removes comms dependency entirely and only implements Relay trait

## Battery
- Replaces communications with host from comms to Relay trait. But I've kept the comms system in place since the battery service seems to talk to other power services and I didn't want to try and change that in this PR. Likewise some uses of statics and intrusive lists still exist in battery since I'm not 100% sure yet on if changing any of that will break things so can be addressed in a future PR
- The changes I did make allowed for some code reduction since we no longer have to differentiate between ACPI events and battery state machine events so could do a little refactoring there

## Uart
- Completely removes legacy comms system and MCTP macro. Only relies now on updated Relay traits.

## Espi
- Removed thermal and battery from the legacy handler. In a followup PR, I will update Debug service and then can compeltely remove all legacy code from the espi service, completing the transition.

These changes were all tested on several dev platforms that can make use of the uart-service. Tested changes on an espi platform and things were mostly correct, but it appears there is a bug that existed even before these changes were made which we are investigating now.

Resolves #721 #723 #725

### DIFF
--- a/uart-service/src/task.rs
+++ b/uart-service/src/task.rs
@@ -5,7 +5,7 @@ use embedded_services::error;
 use embedded_services::relay::mctp::RelayHandler;
 
 pub async fn uart_service<R: RelayHandler, T: UartRead + UartWrite>(
-    uart_service: &'static Service<'_, R>,
+    uart_service: &Service<R>,
     mut uart: T,
 ) -> Result<embedded_services::Never, Error> {
     // Note: eSPI service uses `select!` to seemingly allow asyncrhonous `responses` from services,
--- a/uart-service/src/lib.rs
+++ b/uart-service/src/lib.rs
@@ -10,12 +10,10 @@
 
 pub mod task;
 
-use core::borrow::BorrowMut;
 use embassy_sync::channel::Channel;
 use embedded_io_async::Read as UartRead;
 use embedded_io_async::Write as UartWrite;
 use embedded_services::GlobalRawMutex;
-use embedded_services::buffer::OwnedRef;
 use embedded_services::relay::mctp::{RelayHandler, RelayHeader, RelayResponse};
 use embedded_services::trace;
 use mctp_rs::smbus_espi::SmbusEspiMedium;
@@ -26,8 +24,6 @@ const BUF_SIZE: usize = 256;
 const HOST_TX_QUEUE_SIZE: usize = 5;
 const SMBUS_HEADER_SIZE: usize = 4;
 const SMBUS_LEN_IDX: usize = 2;
-
-embedded_services::define_static_buffer!(assembly_buf, u8, [0u8; BUF_SIZE]);
 
 #[derive(Clone)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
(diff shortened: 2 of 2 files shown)


williampMSFT (approved):
:shipit:

williampMSFT:
> Hi, @kurtjd! Would it be possible to separate this into one subsystem per PR to make it easier to review?

I think splitting this one up might be more trouble than it's worth - doing all of them in one shot like this lets him directly convert the uart service to the new system without needing a bunch of boilerplate to simultaneously support both systems (like I had to do in the espi service) - but if you feel strongly about it we can chat about options :)

espi-service/src/mctp.rs
```diff
@@ -19,8 +19,6 @@ mod legacy_relay {
     use super::*;
     use embedded_services::relay::mctp::impl_odp_mctp_relay_types;
     impl_odp_mctp_relay_types!(
-        Battery,   0x08, (comms::EndpointID::Internal(comms::Internal::Battery)), battery_service_messages::AcpiBatteryRequest, battery_service_messages::AcpiBatteryResult;
-        Thermal,   0x09, (comms::EndpointID::Internal(comms::Internal::Thermal)), thermal_service_messages::ThermalRequest,     thermal_service_messages::ThermalResult;
         Debug,     0x0A, (comms::EndpointID::Internal(comms::Internal::Debug)),   debug_service_messages::DebugRequest,         debug_service_messages::DebugResult;
```
philgweber: When MCTP message is received by eSPI how does it get relayed to the Battery? This uses Comms service? Do we need to do the same for Debug or why is debug left here?

uart-service/src/task.rs
```diff
@@ -1,18 +1,13 @@
...
 use embedded_io_async::Read as UartRead;
 use embedded_io_async::Write as UartWrite;
-use embedded_services::comms;
 use embedded_services::error;
+use embedded_services::relay::mctp::RelayHandler;
 
-pub async fn uart_service<T: UartRead + UartWrite>(
-    uart_service: &'static Service<'_>,
+pub async fn uart_service<R: RelayHandler, T: UartRead + UartWrite>(
+    uart_service: &'static Service<'_, R>,
```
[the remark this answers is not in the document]
kurtjd: Ah yeah you're right, I missed that, thanks! I remember seeing eSPI service was doing something different but forgot to follow through on UART. Might be able to update this PR with that.

espi-service/src/mctp.rs
williampMSFT: To your question re: the debug service, yes, we do need to do it for the debug service too, but I chatted with Kurtis offline and I think the plan is to punt that to a later PR because the debug service is kinda broken right now anyway

kurtjd: Yes, as this PR was already pretty large in scope, I figured the debug service can be handled in a separate PR (since it's not being used with the uart-service anywhere currently so don't need to worry about debug/uart compatibility).

kurtjd:
> > embedded_services::define_static_buffer!(assembly_buf, u8, [0u8; BUF_SIZE]);
> 
> Possibly out of scope for this PR, but I think this can go away and be replaced with a member variable in the Service struct now that it doesn't need to outlive 'static
> 
> Refers to: uart-service/src/lib.rs:30 in [4a20de5](https://github.com/OpenDevicePartnership/embedded-services/commit/4a20de58f410569362fc1854634a1bd7608ab965). [](commit_id = [4a20de5](https://github.com/OpenDevicePartnership/embedded-services/commit/4a20de58f410569362fc1854634a1bd7608ab965), deletion_comment = False)

Was able to just remove the whole static buffer entirely. I think I originally had this to match the original espi implementation, but it's not really needed since we only use the buffer temporarily during packet creating. It doesn't need to live across function calls.

PR #735, opened 2026-02-25
espi-service: Remove legacy comms

### LINKED ISSUE #703: espi-service: Migrate from comms to direct-async, pass in relayable services at construction time
Currently, the eSPI service talks to all the services it needs to relay messages to over the comms system.  This is a problem for testing because the comms system requires that everything that talks over it have lifetime 'static, which means that you can't ever create more than one instance.  This locks us out of writing tests that instantiate services that use the comms system.

Therefore, to enable testing, we're going to provide a non-comm-system path for services to be relayed over the eSPI service that does not impose this 'static lifetime requirement.  As part of this, we'll need to modify the eSPI service to have storage provided by the caller rather than statically declared in the module, and we'll need to provide trait implementations and boilerplate to allow the eSPI service to relay an arbitrary set of relayable messages.

Once we have that, services should be able to independently migrate to using the new facility to communicate with the espi-service.  Once all relayable services have migrated, the eSPI service can remove its usage of the comms system and itself become testable against mocks (may also depend on breaking dependencies on embassy-imxrt)

### LINKED ISSUE #722: Migrate debug service away from comms system
See what was done in #716 for the time-alarm service for an example. This will allow us to write integration tests against the service and allow OEMs to host their own services that relay messages over the same bus as our services.

kurtjd:
Removes all legacy comms paths from the eSPI service and removes the `'static` lifetime (instead preferring a generic `'hw` lifetime).

Also removes the legacy comms path from the debug service. However, the effort here was minimal. The debug service is due for a major overhaul in the near future so more effort will be made getting debug up to parity with the other services in a future PR.

Integrated into soc-ec and builds for our reference platforms using eSPI, however my current eSPI device is again facing booting issues so working on trying to get that fixed and these changes tested...

Resolves #703 
Resolves #722


debug-service/src/task.rs
```diff
@@ -88,10 +84,7 @@ pub async fn no_avail_to_host_task() -> Result<embedded_services::Never, Error>
...
 
     embedded_services::info!("no avail to host task start");
-    use crate::debug_service::{host_endpoint_id, no_avail_notify_signal};
-    use embedded_services::comms::{self, EndpointID, Internal};
-
-    let host_ep = host_endpoint_id().await;
+    use crate::debug_service::no_avail_notify_signal;
 
     let acpi_owned = no_avail_acpi_buf::get_mut().expect("defmt staging buffer already initialized elsewhere");
     {
```
jerrysxie: This seems to be dead code at this point. Is the plan to remove this later?

PR #741, opened 2026-03-05
[RFC] Add notification support for relay services

kurtjd:
There was some discussion offline on if notifications should be handled as broadcast messages, but I think we can keep it pretty simple if we just keep notifications limited to point-to-point (as in, service -> relay).

This allows us to not need to know statically how many publishers there are (which, since a `PubSubChannel` is type stated over the number of publishers, this knowledge would need to bleed into the interface and all the services that need to publish notifications) The way around that would be to use dynamic senders/receivers from pubsub (which erases that particular type state) but I think it's best if we avoid dynamic dispatch if we can help it.

Thus, I decided on a regular `Channel`, with the intention that the relay service is the sole listener and we can create as many `Notifiers` as we want to at run time. My idea is if services also need to notify other services when they signal a notification, that can be handled by each particular service as it sees fit (similar to how power-policy broadcasts messages).

Mentioned in the doc strings, I use a `Channel` with a fixed size of 1 so that typestate is further hidden from the interface, and I think that is justified because notifications are pretty rare events, thus multiple services raising notifications simultaneously should be even rarer. The channel at least gives us backpressure so that no notification is lost if it does happen, and as mentioned the performance penalty should be negligible if backpressure is encountered.

Another reason I wanted to keep notifications specific to only the relay service (and not intended as broadcast messages) is because notifications are not intended to hold payloads, and the notification id would be meaningless to other services. The id is quite platform specific and only the relay service would know what to do with it. For example, the current espi-service would interpret it as an IRQ offset, and the future hid-service would interpret it as a specific GPIO line to make active. Thus, when a new `Notifier` is created, basically the caller has to know how the provided id will be used by whichever relay service they are choosing to use on their platform, and how that will ultimately bubble up to the host.

Finally, I decided to add a method to the `RelayHandler` trait for getting a reference to the notification listener, mainly just to make it easier for instantiating the relay service (so we only have to pass in the relay handler, and not the notification listener additionally).

This PR is currently RFC to get input, and I've quickly integrated it into espi-service and thermal-service to demonstrate how it would be used. When ready for merge, I will remove the espi and thermal service integrations and make those more fleshed out PRs. Fully integrated on a platform it would look like:

```rust
embedded_services::relay::mctp::impl_odp_mctp_relay_handler!(
    OdpRelayHandler;
    Battery, 0x08, battery_service::Service;
    Thermal, 0x09, thermal_service::Service<'static>;
    TimeAlarm, 0x0B, time_alarm_service::Service<'static>;
);

embedded_services::init().await;

// Setup notification handler
static NH: StaticCell<embedded_services::relay::notifications::NotificationHandler> = StaticCell::new();
let notification_handler = embedded_services::relay::notifications::NotificationHandler::new();
let notification_handler = NH.init(notification_handler);

// Configure sensors, fans, etc...
// Here we have to know that espi is our relay handler, and that an ID of 0x1 would correspond to an IRQ offset for espi,
// and that the host SoC understands that this means a thermal notification.
let thermal = ts::Service::init(&STORAGE, sensors, fans, notification_handler.new_notifier(0x1)).await;

// Configure and setup battery service...
// Configure and setup time-alarm service...

// We create the sole Listener and pass it to the relay handler
let relay_handler = OdpRelayHandler::new(notification_handler.new_listener().unwrap(), battery, thermal, time_alarm);

// Configure espi peripheral...
static ESPI_SERVICE: OnceLock<espi_service::Service<OdpRelayHandler<'static>>> = OnceLock::new();
let espi_service = espi_service::Service::init(&ESPI_SERVICE, espi, relay_handler).await;
```

### DIFF
--- a/embedded-service/src/relay/mod.rs
+++ b/embedded-service/src/relay/mod.rs
@@ -164,6 +164,9 @@ pub mod mctp {
         type ResultEnumType: for<'buf> mctp_rs::MctpMessageTrait<'buf, Header = Self::HeaderType>
             + RelayResponse<Self::ServiceIdType, Self::HeaderType>;
 
+        /// Returns a reference to the notification `Listener`.
+        fn notification_listener(&self) -> &crate::relay::notifications::Listener<'_>;
+
         /// Process the provided request and yield a result.
         fn process_request<'a>(
             &'a self,
@@ -449,6 +452,7 @@ pub mod mctp {
 
 
                     pub struct $relay_type_name<'hw> {
+                        listener: $crate::relay::notifications::Listener<'hw>,
                         $(
                             [<$service_name:snake _handler>]: &'hw $service_handler_type,
                         )+
@@ -456,11 +460,13 @@ pub mod mctp {
 
                     impl<'hw> $relay_type_name<'hw> {
                         pub fn new(
+                            listener: $crate::relay::notifications::Listener<'hw>,
                             $(
                                 [<$service_name:snake _handler>]: &'hw $service_handler_type,
                             )+
                         ) -> Self {
                             Self {
+                                listener,
                                 $(
                                     [<$service_name:snake _handler>],
                                 )+
@@ -473,6 +479,10 @@ pub mod mctp {
                         type HeaderType = OdpHeader;
                         type RequestEnumType = HostRequest;
                         type ResultEnumType = HostResult;
+
+                        fn notification_listener(&self) -> & $crate::relay::notifications::Listener<'_> {
+                            &self.listener
+                        }
 
                         fn process_request<'a>(
                             &'a self,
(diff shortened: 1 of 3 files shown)


jerrysxie (commented):
@kurtjd I am a little confused over the direction of this notification. How is this different from a service trying to send the host a message through the relay service?  I thought this notification is going this direction: host -> relay service -> corresponding EC service.

If it is intending for EC service -> relay service -> host, can we just use the existing mechanism to send a message?

embedded-service/src/relay/mod.rs
```diff
@@ -449,18 +452,21 @@ pub mod mctp {
...
                             [<$service_name:snake _handler>]: &'hw $service_handler_type,
                         )+
                     }
 
                     impl<'hw> $relay_type_name<'hw> {
                         pub fn new(
+                            notification_handler: $crate::relay::notifications::NotificationHandler,
                             $(
                                 [<$service_name:snake _handler>]: &'hw $service_handler_type,
                             )+
```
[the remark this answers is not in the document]
kurtjd: Done. The NotificationHandler is now meant to live as long as needed, and added a separate singleton Listener type which can be passed to the relay service.

embedded-service/src/relay/mod.rs
```diff
@@ -501,3 +511,129 @@ pub mod mctp {
...
+        /// Only one [`Listener`] may exist.
+        /// Attempting to create another will return [`Error::ListenerInstantiated`].
+        pub fn new_listener(&self) -> Result<Listener<'_>, Error> {
+            if !self.listener_instantiated.swap(true, Ordering::Relaxed) {
+                Ok(Listener {
+                    receiver: self.channel.receiver(),
+                })
+            } else {
+                Err(Error::ListenerInstantiated)
+            }
```
williampMSFT: This looks odd to me - it seems like this would prevent other services from also receiving these notifications, right? Why do we have this limitation?

embedded-service/src/relay/mod.rs
```diff
@@ -501,3 +511,129 @@ pub mod mctp {
...
+        /// Try to listen for a notification from any service.
+        ///
+        /// Returns [`None`] if no notification is currently available.
+        pub fn try_listen(&self) -> Option<u8> {
+            self.receiver.try_receive().ok()
+        }
+    }
+
+    /// Notification handler.
+    pub struct NotificationHandler {
```
RobertZ2011: Take a look at this `EventReceiver` struct: https://github.com/OpenDevicePartnership/embedded-services/blob/0ebad079501b7736119bae68fac93b8c209a5d34/power-policy-service/src/psu.rs#L18. I think it accomplishes something very similar to what you're trying to do here. Using the `Sender` and `Receiver` traits also makes the design more flexible. The user of the type can choose what channel is appropriate and makes it easy to inject custom logic for debugging/profiling/etc. It also makes the code generic over the mutex type used by the channel, and hides lifetimes which can make things a little cleaner.

RobertZ2011 (changes requested):

PR #748, opened 2026-03-13
[DRAFT] [Time/Alarm] Split interface and relay logic

williampMSFT:


kurtjd:
I really like this, cool idea to decouple the relay interface from the service interface. Should definitely help with composability. I don't really have any additional suggestions; this seems good to me and I'd approve once the checks pass.

time-alarm-service-relay/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+mod serialization;
+pub use serialization::{AcpiTimeAlarmRequest, AcpiTimeAlarmResponse, AcpiTimeAlarmResult};
+
+pub struct TimeAlarmServiceRelayHandler<T: TimeAlarmService> {
+    service: T,
+}
+
+impl<T: TimeAlarmService> TimeAlarmServiceRelayHandler<T> {
+    pub fn new(service: T) -> Self {
```
williampMSFT: Still thinking through this bit - by taking the service by value, I think we implicitly impose the requirement that either we take sole ownership of the service, or the service's control handle must be cloneable.

In the case of the time-alarm service I think it's fine for the control handle to be cloneable, but that may not generalize well. I think it carries the implication that the service's internal reference must be non-`mut`, which is true of the time-alarm service but may not be for other services that want external synchronization.

Maybe the answer here is that for services that want to be relayable and also have methods that take `&mut self`, their relay objects need to be generic over a `Lockable` or something? Would be interested in others' thoughts on this one

williampMSFT:
Per discussion offline, this change is going to require a breaking interface change for relay handlers (going from accepting references to the service to accepting ownership of a relay handler object that contains a reference to the service), and it's also going to touch a bunch of stuff.  Therefore, to cut down on the size of individual PRs, I spun a feature branch `feature/relay-split` that we're going to do separate PRs into for TAD, thermal, and battery, and then once that's all in we can RI that to `v0.2.0`.

This PR is obsoleted by https://github.com/OpenDevicePartnership/embedded-services/pull/776, which targets that feature branch.

PR #753, opened 2026-03-17
Update embassy-sync and embassy-executor

bramsdell-ms:
Updating versions for embassy-sync and embassy-executor. Dependent on OpenDevicePartnership/tps6699x#89

### DIFF
--- a/examples/rt633/Cargo.toml
+++ b/examples/rt633/Cargo.toml
@@ -32,9 +32,9 @@ embassy-imxrt = { git = "https://github.com/OpenDevicePartnership/embassy-imxrt"
     "unstable-pac",
 ] }
 
-embassy-sync = { version = "0.7.2", features = ["defmt"] }
-embassy-executor = { version = "0.9.1", features = [
-    "arch-cortex-m",
+embassy-sync = { version = "0.8", features = ["defmt"] }
+embassy-executor = { version = "0.10", features = [
+    "platform-cortex-m",
     "executor-thread",
     "executor-interrupt",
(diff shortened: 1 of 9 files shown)


examples/rt633/Cargo.toml
```diff
@@ -32,7 +32,7 @@ embassy-imxrt = { git = "https://github.com/OpenDevicePartnership/embassy-imxrt"
     "unstable-pac",
 ] }
 
-embassy-sync = { version = "0.7.2", features = ["defmt"] }
+embassy-sync = { version = "0.8", features = ["defmt"] }
 embassy-executor = { version = "0.9.1", features = [
```
jerrysxie, line 36: Do we not want to upgrade `embassy-executor` as well?

jerrysxie (approved):

examples/rt633/Cargo.toml
gjpmsft: Yeah intentionally kept at `0.9.1` for now. Executor `0.10` removed `must_spawn()` so task functions now return `Result<SpawnToken, SpawnError>`, so all ~70 `must_spawn(task())` calls in examples would need migration to `spawn(task().unwrap())`. Combined with `unwrap_used = "deny"` lint, that becomes a larger refactor.
 
 `embassy-executor` 0.9.1 is fully compatible with `embassy-sync` 0.8 (no cross-dependency), so the executor bump can be a follow-up PR.

felipebalbi (approved):

PR #759, opened 2026-03-20
Battery-service: convert to uniform init / external resources

williampMSFT:
Update battery service to implement the uniform init trait.  This is a precursor to splitting the public interface from the relay code.

### DIFF
--- a/battery-service/src/lib.rs
+++ b/battery-service/src/lib.rs
@@ -157,7 +157,7 @@ impl<'hw, const N: usize> Service<'hw, N> {
 #[derive(Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum InitError {
-    DeviceRegistrationFailed,
+    DeviceRegistrationFailed(crate::device::DeviceId),
     CommsRegistrationFailed,
 }
 
@@ -179,7 +179,7 @@ where
         for device in init_params.devices {
             if service.register_fuel_gauge(device).is_err() {
                 error!("Failed to register battery device with DeviceId {:?}", device.id());
-                return Err(InitError::DeviceRegistrationFailed);
+                return Err(InitError::DeviceRegistrationFailed(device.id()));
             }
         }
 


battery-service/src/lib.rs
```diff
@@ -15,112 +15,200 @@ pub mod controller;
...
+    }
+}
+
+/// Control handle for the battery service. Use this to interact with the battery service.
+#[derive(Clone, Copy)]
+pub struct Service<'hw, const N: usize> {
+    inner: &'hw ServiceInner<N>,
+}
+
+impl<'hw, const N: usize> Service<'hw, N> {
```
williampMSFT: @tullom per our conversation offline, please check me on whether this is the right public API surface - I mostly just kept the same public functions without reasoning about them, but there were some public member variables that this made non-public and I don't have a great understanding of if any of their state should be part of the public interface.
Longer term, I think we want to break this API surface out into a trait in a separate interface crate so we can't keep them as just public variables, but if we truly need public access to some of that state then please let me know and we can figure out how to express that in a trait-friendly way

battery-service/src/lib.rs
```diff
@@ -15,112 +15,200 @@ pub mod controller;
...
-    pub const fn new() -> Self {
-        Self::new_inner(context::Config::new())
-    }
-
-    /// Create a new battery service instance with context configuration.
-    pub const fn new_with_ctx_config(config: context::Config) -> Self {
-        Self::new_inner(config)
-    }
+/// The main service implementation.
+struct ServiceInner<const N: usize> {
```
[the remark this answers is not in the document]
williampMSFT: Right now, the reason it's unused is that this thing uses an IntrusiveList to track the set of batteries (see context.rs).  The plan is to get rid of that and replace it with the service being generic over the number of parameters (out of scope for this PR).  This will incidentally also let us have non-'static lifetime (and therefore write integration tests).

Since we have to be generic over the number of batteries for the InitParams anyway, I figured we could save ourselves a second breaking change if we just make it generic over the size now - that way, the follow-up can just relax the 'static lifetime bound on the trait impl rather than requiring clients to change how they're allocating storage again

battery-service/src/lib.rs
tullom: For now, I think this should be fine. I'll put out another PR adding in the new public API based on the ACPI methods (BST, BIX, etc).

PR #772, opened 2026-04-01
power-policy-service/tests: Introduce dedicated Test trait

RobertZ2011:
The existing traits used to resolve lifetime issues with test closures are messy and difficult to understand. Introduce a dedicated `Test` trait to make things cleaner.

This PR is smaller than it seems, but a lot of lines were touched due to the formatting changes required with moving the previously free functions into traits.

### DIFF
--- a/power-policy-service/tests/consumer.rs
+++ b/power-policy-service/tests/consumer.rs
@@ -92,7 +92,7 @@ impl Test for TestSingle {
 }
 
 /// Test swapping to a higher powered device.
-pub struct TestSwapHigher;
+struct TestSwapHigher;
 
 impl Test for TestSwapHigher {
     async fn run<'a>(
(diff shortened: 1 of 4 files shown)


power-policy-service/tests/consumer.rs
```diff
@@ -24,449 +25,473 @@ const PER_CALL_TIMEOUT: Duration = Duration::from_millis(1000);
...
-            .await
-            .simulate_consumer_connection(LOW_POWER.into())
-            .await;
-
-        assert_eq!(
-            with_timeout(PER_CALL_TIMEOUT, device0_signal.wait()).await.unwrap(),
-            (
-                1,
-                FnCall::ConnectConsumer(ConsumerPowerCapability {
+pub struct TestSwapHigher;
```
jerrysxie: Does this need to be `pub`? It seems like it is only used in this file.

jerrysxie (dismissed):

power-policy-service/tests/consumer.rs
RobertZ2011: Done.

jerrysxie (approved):

kurtjd (approved):

tullom (approved):

PR #777, opened 2026-04-07
thermal-service: Separate interface from implementation

### LINKED ISSUE #756: thermal-service: Refactor to make use of `runnable_service` traits

### LINKED ISSUE #781: thermal-service: Consider making relay types newtypes instead of typedef
Think about some of the pros vs cons of this and if it makes sense, then do it.

kurtjd:
This PR overhauls the thermal-service a bit. Essentially, the thermal interface has been split from the implementation into a trait, IPC has been removed and replaced with direct async calls, and a relay service handler wrapper has been introduced.

Instead of one monolithic "thermal service", there are really 3 services here: sensor service, fan service, and thermal service, each with their own interfaces. The thermal service now mainly exists as the relay handler, since it can keep track of the different spawned sensor and fan services by ID, therefore knowing which sensors and fans to talk to when a request comes from the host.

Also split into a thermal service relay handler type which is intended to wrap any thermal service implementation.

There are some things I will likely want to tweak in the future based off some work Robert has been doing such as better use of some of the event sender/receiver traits and use of the Lockable trait (I heavily use interior mutability with direct dep on embassy Mutex, not sure if I can somehow make this generic over the Lockable trait?).

With the removal of IPC, need to also think about how to handle heterogeneous collections of sensors and fans. Currently this design just assumes the user will need to handle that on their own, by introducing enum dispatch or something behind the scenes, but will investigate if there are ways to make that easier (such as using the `enum_dispatch` crate and marking these traits).

This is of course a large breaking change so will announce on Zulip after merge.

Resolves #756 
Resolves #781


thermal-service/src/sensor.rs
```diff
@@ -231,337 +91,272 @@ struct State {
...
+                self.en_signal.wait().await;
             }
         }
     }
+
+    async fn broadcast_event(&self, event: sensor::Event) {
+        let mut senders = self.event_senders.lock().await;
+        for sender in senders.iter_mut() {
+            sender.send(event).await;
+        }
```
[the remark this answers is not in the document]
kurtjd: I think this is fine as-is. Currently no other task uses this event senders mutex and even if they did I think I would prefer a single event get sent out on all senders before another task can then send off events.

thermal-service/src/fan.rs
```diff
@@ -1,600 +1,435 @@
...
-    /// The thermal service handle for this fan to communicate with a sensor.
-    pub thermal_service: &'hw crate::Service<'hw>,
+impl<'hw, T: fan::Driver, S: sensor::SensorService, E: Sender<fan::Event>, const SAMPLE_BUF_LEN: usize>
+    odp_service_common::runnable_service::ServiceRunner<'hw> for Runner<'hw, T, S, E, SAMPLE_BUF_LEN>
+{
+    async fn run(self) -> embedded_services::Never {
+        loop {
+            let _ =
+                embassy_futures::join::join(self.service.handle_sampling(), self.service.handle_auto_control()).await;
+        }
```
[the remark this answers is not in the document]
kurtjd: It looks like we are standardizing on the `loop { ... }` pattern. Also avoids risk of panic.

uart-service/Cargo.toml
```diff
@@ -22,6 +22,12 @@ mctp-rs = { workspace = true }
 embedded-io-async.workspace = true
 num_enum.workspace = true
 
+# TODO Service message type crates are a temporary dependency until we can parameterize
+#      the supported messages types at UART service creation time.
+battery-service-messages.workspace = true
+debug-service-messages.workspace = true
+thermal-service-interface.workspace = true
+time-alarm-service-interface.workspace = true
```
williampMSFT: I don't think we need these dependencies anymore - the uart service doesn't need any knowledge of specific message types, that knowledge gets passed in as a template parameter when it's instantiated by the application layer

kurtjd: Yep sorry realized that and removed these, looks like you might have seen them before I had the chance.

thermal-service-relay/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+mod serialization;
+
+pub use serialization::{ThermalError, ThermalRequest, ThermalResponse, ThermalResult};
+use thermal_service_interface::ThermalService;
+use thermal_service_interface::fan::{self, FanService};
+use thermal_service_interface::sensor::{self, SensorService};
+
+/// 16-bit ACPI variable length.
+pub type VarLen = u16;
```
williampMSFT: Not sure if this is in scope for this PR, so feel free to wontfix if it isn't, but would it make sense to have these be newtypes rather than typedefs?  i.e. 
`struct Milliseconds(pub u32);`

As currently written, I think you can pass a `Milliseconds` into a function that's expecting a `Decikelvin` because they're type aliases for the same type, but if you use newtypes then the compiler will check that you're not mixing your units for you

uart-service/Cargo.toml
williampMSFT: ah, yeah, sorry about that, I think codeflow might not update as quickly as the web UI when updates are pushed. will keep an eye out for that :)

thermal-service-relay/src/lib.rs
kurtjd: Thanks, that is a good idea and worth considering, but if you don't mind I'd like to make a separate PR for that (just to help keep the scope of this PR narrower as it is already pretty big).

Tracking with: https://github.com/OpenDevicePartnership/embedded-services/issues/781

thermal-service-relay/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+/// Convert degrees Celsius to deciKelvin.
+pub const fn c_to_dk(c: f32) -> DeciKelvin {
+    ((c + 273.15) * 10.0) as DeciKelvin
+}
+
+/// Convert deciKelvin to degrees Celsius.
+pub const fn dk_to_c(dk: DeciKelvin) -> f32 {
+    (dk as f32 / 10.0) - 273.15
+}
```
williampMSFT: nit: if these conversions are publicly useful, they might belong in the interface crate, and if they're not then we should consider making them private

thermal-service-interface/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+
+/// Thermal service interface trait.
+pub trait ThermalService: Copy {
+    /// Associated type for registered sensor services.
+    type Sensor: sensor::SensorService;
+    /// Associated type for registered fan services.
+    type Fan: fan::FanService;
+
+    /// Retrieve a reference to the sensor service with the specified instance ID, if it exists.
+    fn sensor(&self, id: u8) -> Option<Self::Sensor>;
```
williampMSFT, line 14: >id: u8 [](http://example.com/codeflow?start=21&length=6)

should these ID types be the ID type that's currently defined in the -relay crate?

thermal-service-interface/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+embassy-time.workspace = true
+embedded-fans-async = "0.2.0"
+embedded-sensors-hal-async = "0.3.0"
+embedded-services.workspace = true
+
+[lints]
+workspace = true
+
+[features]
+defmt = ["dep:defmt"]
```
williampMSFT: should this propagate the feature to embedded-services (and any other dependencies that have a defmt feature)? I just realized I got this wrong on the battery-service PR too, lol

thermal-service/src/lib.rs
```diff
@@ -1,101 +1,49 @@
...
 }
 
-impl<'hw> Service<'hw> {
+impl<'hw, S: SensorService, F: FanService> Service<'hw, S, F> {
+    /// Initializes the thermal service with the provided sensors and fans.
     pub async fn init(
-        service_storage: &'hw embassy_sync::once_lock::OnceLock<Service<'hw>>,
-        sensors: &'hw [&'hw sensor::Device],
-        fans: &'hw [&'hw fan::Device],
+        service_storage: &'hw embassy_sync::once_lock::OnceLock<Service<'hw, S, F>>,
```
williampMSFT: If you do the `Resources` / `ServiceInner` thing here, you can avoid prescribing that the storage be specifically a OnceLock - might make things more flexible for users

thermal-service/src/fan.rs
```diff
@@ -1,600 +1,435 @@
...
+    }
+
+    async fn set_rpm_update_period(&self, period: Duration) {
+        self.config.lock().await.update_period = period;
+    }
+
+    async fn set_state_temp(&self, state: fan::State, temp: DegreesCelsius) {
+        let mut config = self.config.lock().await;
+        match state {
+            fan::State::Off => (),
```
jerrysxie: This might be pre-existing already, fan state is off is ignored?

thermal-service/src/sensor.rs
```diff
@@ -231,337 +91,272 @@ struct State {
...
+{
+    fn new(init_params: InitParams<'hw, T, E>) -> Self {
         Self {
-            device: Device::new(id),
-            controller: Mutex::new(controller),
-            profile: Mutex::new(profile),
+            driver: Mutex::new(init_params.driver),
             state: Mutex::new(State::default()),
+            en_signal: Signal::new(),
+            event_senders: Mutex::new(init_params.event_senders),
```
jerrysxie: Do we need to put `event_senders` behind a mutex? It is only called by `broadcast_event()`.

thermal-service/src/sensor.rs
```diff
@@ -37,14 +19,14 @@ macro_rules! with_retry {
...
+        let mut retry_attempts = $self.config.lock().await.retry_attempts;
 
         loop {
             if retry_attempts == 0 {
-                break Err(Error::Hardware);
+                break Err(sensor::Error::RetryExhausted);
             }
 
-            match embassy_time::with_timeout(embassy_time::Duration::from_millis(BUS_TIMEOUT), $bus_method).await {
+            match with_timeout(BUS_TIMEOUT, $bus_method).await {
```
jerrysxie, line 30: We should consider bubbling up the actual hardware error or timeout error up.

thermal-service/Cargo.toml
```diff
@@ -18,7 +18,8 @@ embedded-hal.workspace = true
 embedded-services.workspace = true
 heapless.workspace = true
 odp-service-common.workspace = true
-thermal-service-messages.workspace = true
+thermal-service-interface.workspace = true
+thermal-service-relay.workspace = true
```
jerrysxie: Does `thermal-service` still need to depend on `mctp-rs`?

thermal-service/src/sensor.rs
```diff
@@ -231,337 +91,272 @@ struct State {
...
+impl<'hw, T: sensor::Driver, E: Sender<sensor::Event>, const SAMPLE_BUF_LEN: usize>
+    ServiceInner<'hw, T, E, SAMPLE_BUF_LEN>
+{
+    fn new(init_params: InitParams<'hw, T, E>) -> Self {
         Self {
-            device: Device::new(id),
-            controller: Mutex::new(controller),
-            profile: Mutex::new(profile),
+            driver: Mutex::new(init_params.driver),
             state: Mutex::new(State::default()),
```
jerrysxie: Does `state` need a mutex? It seems like it is only used in one place.

thermal-service/src/fan.rs
```diff
@@ -1,600 +1,435 @@
...
-            controller: Mutex::new(controller),
-            profile: Mutex::new(profile),
+            driver: Mutex::new(init_params.driver),
+            state: Mutex::new(fan::State::Off),
+            en_signal: Signal::new(),
+            config: Mutex::new(init_params.config),
             samples: Mutex::new(SampleBuf::create()),
-            state: Mutex::new(FanState::Off),
+            sensor: init_params.sensor_service,
+            event_senders: Mutex::new(init_params.event_senders),
```
jerrysxie: `event_senders` does not seem like it needs a mutex around it.

thermal-service/src/fan.rs
```diff
@@ -1,600 +1,435 @@
...
-                let _ = controller.set_speed_rpm(max_rpm).await.map_err(|_| Error::Hardware)?;
+                let _ = controller
+                    .set_speed_rpm(max_rpm)
+                    .await
+                    .map_err(|_| fan::Error::Hardware)?;
             }
         }
         drop(controller);
 
         let state = *self.state.lock().await;
```
jerrysxie: Maybe just lock it once:
```
let mut state = self.state.lock().await;
trace!("Fan transitioned to {:?} state from {:?} state", to, *state);
*state = to;
```

thermal-service-relay/src/lib.rs
kurtjd: Good point, thanks! I think it would make sense to keep them pub because then I can replace the re-implementation in the ec-test-app (see: https://github.com/OpenDevicePartnership/odp-platform-common/blob/a222a92f04b7af0480b627cf57c840648473d99d/ec-test-lib/src/common.rs#L15) with the ones here.

Though I suppose it still leaves the question if these are better left in here or thermal-service-interface. I'd argue they should remain there since they are tightly coupled to the relay implementation since internally, we only work in celsius (not decikelvin). Decikelvin is just the outside interface.

My goal is to also use the *-service-relay crates as the "shared" types between here and the ec-test-app side (since the ec-test-app only cares about relayable interfaces). We currently have the request/response types in the relay crates now which the ec-test-app will need so think it makes sense to keep all relay-related stuff consolidated here so the ec-test-app doesn't need to additionally pull in the interface crate.

Does this reasoning seem sound? If not, I can move them, don't mind.

williampMSFT: aha! I hadn't considered that external repos would depend on the relay crate, but now you mention the ec-test-app, that makes a lot of sense.  If decikelvin is only used for relaying then this is probably the right place for them.

thermal-service-relay/src/lib.rs
kurtjd: Ok I decided to think about this some more. I've removed most of these type aliases as they were mostly vestigial and now just used internally so no need to type alias or provide new types for them. I removed InstanceId as well because I don't think that's really any more clear than just using a `u8`.

I decided to turn `DeciKelvin` into a new type at your suggestion as I think it's good so that users don't mistake it as already in Celsius (e.g. caller might just think a u32 got returned representing an integer Celsius temperature). So, I made it a new type and provided conversion methods on the new type but still kept it public as per our other discussion.

thermal-service-interface/src/lib.rs
kurtjd: Decided to remove the InstanceId type alias (see response to your other feedback comment).

thermal-service/src/lib.rs
kurtjd: Sounds good, yeah I remember you mentioning this. You think the move is to use the Resources and ServiceInner pattern, but just not implement the runnable service trait on this (since the thermal-service handle isn't meant to be runnable on its own)? I will give it a go.

thermal-service/src/sensor.rs
kurtjd: Yeah, so basically ServiceInner has to have all its field use interior mutability because none of the methods can take `&mut self` because both `Service` and `Runner` need to hold references to `ServiceInner` simultaneously in the current `runnable service` design.

If I instead make ServiceInner methods act on `&mut self` and make Service and Runner hold Mutex<ServiceInner> instead, this is no good because the Runner will end up holding the lock indefinitely as it runs a never ending async method.

`broadcast_event()` uses the `send()` method on `event_senders` which expects an `&mut self` so the field needs to be behind a Mutex.

This heavy use of interior mutability is something I want to think more about in the future, but don't see an immediately obvious way to mitigate it. @williampMSFT do you have any ideas on how we can make `ServiceInner` types work with `&mut` in the current design, or is this a limitation?

thermal-service/src/sensor.rs
kurtjd: Yes, any field that needs to be mutated needs a Mutex (see response to your other comment) as far as I understand.

thermal-service/Cargo.toml
kurtjd: Nope, good catch. A few dead deps here now that I removed. Thanks!

thermal-service/src/fan.rs
kurtjd: See my response to your sensor comment (though if folks point out something I'm missing in my reasoning I can change it).

thermal-service/src/fan.rs
kurtjd: Good idea, thanks!

thermal-service/src/sensor.rs
williampMSFT: I think we have basically three cases for state mutation:

1) If you have state that needs to be modified on both your worker task and through your control handle, I think you're kinda stuck using interior mutability / mutexes.

2) If you have state that only needs to be modified on your worker task but can't be modified through your control handle, you can have that state be a member variable of the Runner struct rather than containing it in the ServiceInner struct to avoid needing to mutex it.  You could potentially use a Channel or Signal or something to have a function on the control handle signal the runner to change its state to sort of promote something from case (1) to case (2), but Signal and Channel use interior mutability themselves, so you're not actually getting away from using interior mutability, you're just spelling it differently.

3) If you have state that needs to be modified on your control handle but not on your worker task, then if you're willing to not have the control handle be cloneable, you could maybe do something where the Resources type contains two Options, one for shared state and one for control-handle-only state, and then have your control handle take a mutable borrow on the control-handle-only state.  However, in the case where you're writing a service that needs a relay service to send messages across espi/i2c/whatever, that relay service will need to either take exclusive ownership over your control handle (since it can't be both cloneable and contain mutable references), or you'll need to use an external mutex to synchronize access to the control handle between the relay handler and whatever else needs to call stuff on the control handle.  This means introducing another StaticCell to contain the control handle, so the ergonomics aren't great, and you're still stuck using a mutex, so it's not clear to me that this buys you anything.

jerrysxie: I see because the different async tasks holding references, so we have to utilize interior mutability. I wonder if there is a way around this. Let me ponder on this for a bit. Let's not the block the PR for it.

williampMSFT: In this specific case, it looks like this only gets used by your worker task, so this might actually be an instance of (2) - move this state to be on the runner?

williampMSFT: or a combination if the runner state is annoying to move, something like

```rust

struct Resources {
  shared_inner: Option<ServiceInner>,
  runner_inner: Option<State>
}

struct Runner<'hw> {
  shared_inner: &'hw ServiceInner,
  runner_inner: &'hw mut State
}

impl Service<'hw> {
  fn new(resources: &'hw mut Resources /*, ... */) -> Result<(Self, Runner)> {
    // ...
    let service = resources.shared_inner.insert(ServiceInner::new(...));
    let runner_resources = resources.runner_inner.insert(State::new());
    Ok( (Self{service}, Runner{service, runner_resources}) )
  }
}

```

williampMSFT: Actually, in this specific case I don't think you even need to put it in the resources, you can just have it be a stack variable on the worker thread?

thermal-service/src/sensor.rs
kurtjd: Currently a `RetryExhausted` error will be bubbled up and sent out as a sensor event, but I agree it would make sense to include the underlying bus error with that.

However that is out of scope of this PR and would require a bit of additional thinking how to generically capture the bus error so made an issue to track this: https://github.com/OpenDevicePartnership/embedded-services/issues/782

thermal-service/src/fan.rs
kurtjd: Yeah, since the idea is the `Off` state is just when the temperature falls below the `config.on_temp` threshold. So it doesn't make much sense to explicitly set the temperature at which we transition to the off state.

Though maybe I should make the interface more explicit that passing in Off as the state argument is invalid... I'd rather not return a Result since this isn't a failure that is expected to happen at runtime (it's just a straight programmer error) but then I'd need to split the State enum up and only accept an enum that doesn't contain the off state. Not sure the best approach.

kurtjd: Will probably take Copilot's suggestion in a related comment about documenting that Off is invalid and will do nothing.

williampMSFT: I kinda like splitting the `State` enum into something like

```rust
struct FanRunMode {
  Constant,
  Max,
  Ramping
}
struct FanState {
  Off,
  On(FanRunMode)
}
```

and then just taking a `FanRunMode` (or some other name, idk)?  That also makes it a bit clearer what the difference between "On" and "Max" is, which I found a bit confusing when I initially read this

kurtjd: Hm yeah that could be a good idea. I'll try that out, thanks!

williampMSFT: Probably less critical because this isn't part of the public interface though, totally reasonable to punt on this

thermal-service/src/sensor.rs
```diff
@@ -231,337 +91,272 @@ struct State {
...
+        self.samples.lock().await.recent()
     }
 
-    /// Retrieve a Mutex wrapping the underlying controller
-    ///
-    /// Should only be used to update OEM specific state
-    pub fn controller(&self) -> &Mutex<GlobalRawMutex, T> {
-        &self.controller
+    async fn temperature_average(&self) -> DegreesCelsius {
+        self.samples.lock().await.average()
```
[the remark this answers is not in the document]
kurtjd: Updated these functions in `utils` to return 0 if divisor is 0 and documented that.

thermal-service/src/fan.rs
```diff
@@ -1,600 +1,435 @@
...
+        let min_rpm = driver.min_start_rpm();
+        let max_rpm = driver.max_rpm();
+
+        // Provide a linear fan response between its min and max RPM relative to temperature between ramp start and max temp
+        let rpm = if temp <= config.ramp_temp {
+            min_rpm
+        } else if temp >= config.max_temp {
+            max_rpm
+        } else {
+            let ratio = (temp - config.ramp_temp) / (config.max_temp - config.ramp_temp);
```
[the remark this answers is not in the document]
kurtjd: This does make a good suggestion that we should check for the invalid case where the user attempts to set, for example, an on temperature threshold that is higher than the ramp temperature threshold. But that check doesn't belong here. Opened #783 to track this.

thermal-service/src/fan.rs
```diff
@@ -1,600 +1,435 @@
...
+    }
+
+    async fn state_temp(&self, state: fan::State) -> DegreesCelsius {
+        let config = self.config.lock().await;
+        match state {
+            fan::State::Off => DegreesCelsius::default(),
+            fan::State::On => config.on_temp,
+            fan::State::Ramping => config.ramp_temp,
+            fan::State::Max => config.max_temp,
         }
```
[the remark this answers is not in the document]
kurtjd: Refactored so there is now a separate `OnState`.

thermal-service/src/fan.rs
kurtjd: Yeah this worked out well. Thanks for the suggestion, updated the PR with it.

thermal-service/src/lib.rs
kurtjd: Alright I've updated thermal service to use this pattern and updated the example (so no longer are we forced into OnceLock). Thanks!

thermal-service/src/fan.rs
```diff
@@ -1,600 +1,433 @@
...
+    async fn set_rpm_update_period(&self, period: Duration) {
+        self.config.lock().await.update_period = period;
+    }
+
+    async fn set_state_temp(&self, state: fan::OnState, temp: DegreesCelsius) {
+        let mut config = self.config.lock().await;
+        match state {
+            fan::OnState::Min => config.min_temp = temp,
+            fan::OnState::Ramping => config.ramp_temp = temp,
+            fan::OnState::Max => config.max_temp = temp,
```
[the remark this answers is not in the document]
kurtjd: This will be addressed via https://github.com/OpenDevicePartnership/embedded-services/issues/783

thermal-service/src/sensor.rs
kurtjd: I think I'm following your other comments which gave me some ideas, but I don't think I follow what you mean making it a stack variable. The event_senders list needs to be passed in by the user but the `ServiceRunner::run()` method only takes self, so we'd need to store the list somewhere in one of these structs (like the Runner struct which I'm working on doing).

williampMSFT: Whoops, sorry, I think I was looking at the wrong file when I wrote this - I was talking about the `state` variable in `sensor.rs`, which appears to only be used in `async fn check_thresholds(&self, temp: DegreesCelsius)`, which appears to only be called by `async fn handle_sampling(&self)`.  In that case, `state` can be a stack variable in `handle_sampling` and you can get rid of the mutex around it.  Will take another look at `event_senders`, it's possible there's nothing to be done for that one

kurtjd: Got it. Right now I'm trying to split some of ServiceInner up into fields and methods that might only explicitly belong with Runner and explicitly only with Service and keep ServiceInner only for what needs to be common between them. Hopefully should figure out a way to minimize some of this.

williampMSFT: Yeah, took another look at both fan.rs and sensor.rs, and in both of them, it looks like the event_senders is used exclusively by the worker thread, so you can make it either member state of the Runner do that snippet a couple comments back where you split the Resources into two different `Option`s and have the runner take a mutable borrow against the runner-only piece.  You're right that any state that needs to be passed in by the caller of new() can't be a stack variable; it only works for `sensor.rs -> state` because that has a `Default::default()` implementation that's the only way it gets constructed

thermal-service-interface/src/fan.rs
```diff
@@ -0,0 +1,131 @@
...
+use embedded_fans_async::{Fan, RpmSense};
+use embedded_sensors_hal_async::temperature::DegreesCelsius;
+
+/// Ensures all necessary traits are implemented for the underlying fan driver.
+pub trait Driver: Fan + RpmSense {}
+
+/// Fan error.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
```
felipebalbi, line 13: You should probably add a non_exhaustive here. Otherwise, adding a new variant in the future will be a breaking change.

thermal-service-interface/src/fan.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Fan encountered a hardware failure.
+    Hardware,
+}
+
+/// Fan event.
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Event {
```
felipebalbi, line 22: Ditto

thermal-service-interface/src/fan.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+/// Fan on (running) state.
+#[derive(Debug, Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum OnState {
+    /// Fan is on and running at its minimum speed.
+    Min,
+    /// Fan is ramping up or down along a curve in response to a temperature change.
+    Ramping,
```
felipebalbi, line 34: Should this be `Ramping(upsize)` or something similar where the `usize` carries the target set point?

thermal-service-interface/src/fan.rs
```diff
@@ -0,0 +1,131 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum State {
+    /// Fan is off.
+    Off,
+    /// Fan is on in the specified [`OnState`].
+    On(OnState),
+}
+
+/// Fan service interface trait.
+pub trait FanService: Copy {
```
felipebalbi: Why `Copy`? Everything takes a reference to `self`. It's not immediately clear why that's required.

thermal-service-interface/src/sensor.rs
```diff
@@ -0,0 +1,96 @@
...
+use embassy_time::Duration;
+use embedded_sensors_hal_async::temperature::{DegreesCelsius, TemperatureSensor};
+
+/// Ensures all necessary traits are implemented for the underlying sensor driver.
+pub trait Driver: TemperatureSensor {}
+
+/// Sensor error.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
```
felipebalbi, line 12: `non_exhaustive`?

thermal-service-interface/src/sensor.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// Sensor encountered a hardware failure.
+    Hardware,
```
felipebalbi, line 14: Carry a `kind` internally, like embedded-Hal traits?

thermal-service-relay/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+use thermal_service_interface::fan::{self, FanService};
+use thermal_service_interface::sensor::{self, SensorService};
+
+/// DeciKelvin temperature representation.
+///
+/// This exists because the host to EC interface expects DeciKelvin,
+/// though internally we still use Celsius for ease of use.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct DeciKelvin(pub u32);
```
felipebalbi, line 16: Why not f32? I don't think anyone is trying to run an EC without an FPU. Could be wrong though.

thermal-service-relay/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// The temperature threshold at which a fan should start ramping up.
+    pub const FAN_RAMP_TEMP: uuid::Bytes = uuid::uuid!("3a62688c-d95b-4d2d-bacc-90d7a5816bcd").to_bytes_le();
+    /// The temperature threshold at which a fan should be at max speed.
+    pub const FAN_MAX_TEMP: uuid::Bytes = uuid::uuid!("dcb758b1-f0fd-4ec7-b2c0-ef1e2a547b76").to_bytes_le();
+    /// The minimum RPM a fan is capable of running at reliably.
+    pub const FAN_MIN_RPM: uuid::Bytes = uuid::uuid!("db261c77-934b-45e2-9742-256c62badb7a").to_bytes_le();
+    /// The maximum RPM a fan is capable of running at reliably.
+    pub const FAN_MAX_RPM: uuid::Bytes = uuid::uuid!("5cf839df-8be7-42b9-9ac5-3403ca2c8a6a").to_bytes_le();
+    /// The current RPM of a fan.
+    pub const FAN_CURRENT_RPM: uuid::Bytes = uuid::uuid!("adf95492-0776-4ffc-84f3-b6c8b5269683").to_bytes_le();
```
felipebalbi, line 47: Looks like we need an MPTF crate which defines these and provides `no_std` helpers.

felipebalbi (changes requested):

thermal-service-interface/src/fan.rs
kurtjd: Hm I might be misunderstanding what you're saying, but `Ramping` state here means the fan is setting its RPM according to a curve based on temperature readings. It doesn't necessarily represent a target RPM that we are trying to hit.

`Ramping` might be a confusing name, I really just chose this name since it was used in some of the earlier MPTF docs. A better name might be `CurveResponse` or something.

Though that is an additional feature I want to add at some point, where instead of immediately setting the RPM, we set a target RPM and slowly ramp up or down to it to prevent jerkiness. But that is a separate idea from here.

thermal-service-relay/src/lib.rs
kurtjd: My understanding is the host-to-EC interface expects integer decikelvin units. This is just how the interface is. Internally we use f32 Celsisus.

thermal-service-relay/src/lib.rs
kurtjd: That could be a good idea. Because these are used by the ec-test-app. Was planning to have ec-test-app use thermal-service-relay as a dep, but might be better to instead split these into a separate crate that can be sued here and by ec-test-app.

thermal-service-relay/src/lib.rs
kurtjd: image

thermal-service/src/sensor.rs
kurtjd: Alright I think I split up methods/fields between `Service` and `Runner` for sensor and fan as much as I could, and was able to remove a few mutexes.

Though mainly still need the driver, samples, and config shared so those must remain in `ServceInner` behind a mutex.

I've kept `state` as a field of `Runner` in sensor.rs because `check_thresholds` still needs `&mut self` for `broadcast_event` so don't see a huge benefit of making it a stack variable (and we still don't need a mutex regardless).

Note: Had to add `PhantomData` fields to the `Service` structs to satisfy the compiler since we need these generics to set up the trait implementation even though we don't actually use them in the `Service` struct.

thermal-service-relay/src/lib.rs
kurtjd: Definitely want to do this, but this is out of scope for this PR. Created https://github.com/OpenDevicePartnership/embedded-services/issues/784 to track this.

thermal-service-interface/src/fan.rs
kurtjd: So a lot of our design is dependent on Service handles being copyable, and specifically without the Copy trait bound this snippet will break:

```rust
impl<'hw, S: SensorService, F: FanService> thermal_service_interface::ThermalService for Service<'hw, S, F> {
    type Sensor = S;
    type Fan = F;

    fn sensor(&self, id: u8) -> Option<Self::Sensor> {
        self.inner.sensors.get(id as usize).copied()
    }

    fn fan(&self, id: u8) -> Option<Self::Fan> {
        self.inner.fans.get(id as usize).copied()
    }
}
```

But your arguments provided in the recent watchdog PR could apply here as well. Instead of forcing Service traits to be copyable, I can move the bound to here:

```rust
impl<'hw, S: SensorService + Copy, F: FanService + Copy> thermal_service_interface::ThermalService
    for Service<'hw, S, F>
{
    type Sensor = S;
    type Fan = F;

    fn sensor(&self, id: u8) -> Option<Self::Sensor> {
        self.inner.sensors.get(id as usize).copied()
    }

    fn fan(&self, id: u8) -> Option<Self::Fan> {
        self.inner.fans.get(id as usize).copied()
    }
}
```

kurtjd: Done.

thermal-service-interface/src/sensor.rs
kurtjd: This is a good idea, but out of scope for this current PR (the original design didn't do this either) and will require a bit of thought. Related to https://github.com/OpenDevicePartnership/embedded-services/issues/782 (created in response a comment Jerry made) which will track this as well.

thermal-service-interface/src/sensor.rs
kurtjd: Done.

thermal-service-interface/src/fan.rs
kurtjd: Done.

thermal-service-interface/src/fan.rs
kurtjd: Done.

kurtjd:
@jerrysxie @williampMSFT @felipebalbi I think I've addressed everyone's feedback as best I could. Please resolve conversations if my reasoning/fixes are adequate.

For suggestions out of scope of this PR, I've created several issues to track them:
#782 #783 #784

thermal-service-relay/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+        let temp = sensor.temperature().await;
+        Ok(ThermalResponse::ThermalGetTmpResponse {
+            temperature: DeciKelvin::from_celsius(temp),
+        })
+    }
+
+    async fn sensor_set_warn_thrs(
+        &self,
+        instance_id: u8,
+        _timeout: u32,
```
williampMSFT, line 72: looks like this is unused - do we still want it?

kurtjd: It's technically part of our MPTF interface and thus is a field in the `ThermalSetThrsRequest` type, but I haven't figured out a good way to make use of it yet.

I could just remove the argument and underscore it here:

```rust
ThermalRequest::ThermalSetThrsRequest {
                instance_id,
                _timeout,
                low,
                high,
            } => self.sensor_set_warn_thrs(instance_id, low, high).await,
```

williampMSFT: eh, if the plan is to use it eventually it's probably fine, I just wanted to make sure it was intentional :)

williampMSFT (approved):

jerrysxie:
> @jerrysxie @williampMSFT @felipebalbi I think I've addressed everyone's feedback as best I could. Please resolve conversations if my reasoning/fixes are adequate.
> 
> For suggestions out of scope of this PR, I've created several issues to track them: #782 #783 #784

@kurtjd I would recommend adding commits instead of amending commits since we utilize squash now. It is a bit hard for reviewers to figure out what new changes are layered on already reviewed changes.

kurtjd:
> > @jerrysxie @williampMSFT @felipebalbi I think I've addressed everyone's feedback as best I could. Please resolve conversations if my reasoning/fixes are adequate.
> > For suggestions out of scope of this PR, I've created several issues to track them: #782 #783 #784
> 
> @kurtjd I would recommend adding commits instead of amending commits since we utilize squash now. It is a bit hard for reviewers to figure out what new changes are layered on already reviewed changes.

Sounds good, I will do that from now on.

jerrysxie (approved):

thermal-service-interface/src/fan.rs
kurtjd: I will resolve this for now but let me know if you still think it should be addressed and I will create an issue for it.

thermal-service-relay/src/lib.rs
kurtjd: I will resolve this for now but let me know if you still think it should be addressed and I will create an issue for it.

felipebalbi (approved):
LGTM!

kurtjd:
Breaking change announced on Zulip: [#embedded-controller > v0.2.0 Service Refactors @ 💬](https://opendevicepartnership.zulipchat.com/#narrow/channel/495405-embedded-controller/topic/v0.2E2.2E0.20Service.20Refactors/near/585207286)

PR #780, opened 2026-04-08
Battery service: Split interface and relay logic into separate crates

williampMSFT:
This change splits the interface for the battery service into two crates - an 'interface' crate that contains traits and support types for the public interface for the battery service, and a 'relay' crate that's written against the 'interface' crate and handles relaying messages to and from a battery service implementation over MCTP.

This split should allow OEMs to substitute their own service implementations if they want to and still reuse our relay code (or write their own relay against the stock service implementation). Additionally, splitting the relay handling from the service implementation should also allow us to add support for different kinds of relay handlers (e.g. HID) without needing to pay the code size cost of having that code on platforms that aren't using it since it's no longer invasive on the service type.

This change only moves the pieces of the battery interface that are required for the relay service into a trait; a future change will migrate the pieces that are targeted at other code running on the EC into the trait as well.


Cargo.toml
```diff
@@ -2,7 +2,8 @@
 resolver = "2"
 members = [
     "battery-service",
-    "battery-service-messages",
+    "battery-service-interface",
+    "battery-service-relay",
     "thermal-service",
```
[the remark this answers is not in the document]
williampMSFT: fixed description

battery-service-interface/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// Maximum rated input power in milliwatts (mW).
+    ///
+    /// 0xFFFFFFFF indicates the value is unavailable.
+    pub max_input_power: u32,
+    /// OEM-specific model number (ASCIIZ). Empty string if not supported.
+    pub model_number: [u8; STD_BIX_MODEL_SIZE],
+    /// OEM-specific serial number (ASCIIZ). Empty string if not supported.
+    pub serial_number: [u8; STD_BIX_SERIAL_SIZE],
+    /// OEM-specific information (ASCIIZ). Empty string if not supported.
+    pub oem_info: [u8; STD_BIX_OEM_SIZE],
```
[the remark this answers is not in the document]
williampMSFT: Looks like this was the case before these changes in the version of this in battery-service-messages, but we got away with it because they're all size 8.  Will change these, but maybe longer-term it might be worth considering if we should even have separate constants for these or if we should only have one constant for these?

battery-service-relay/src/serialization.rs
```diff
@@ -0,0 +1,589 @@
...
+    })
+}
+
+const PIF_MODEL_NUM_START_IDX: usize = 12;
+const PIF_MODEL_NUM_END_IDX: usize = PIF_MODEL_NUM_START_IDX + STD_BIX_MODEL_SIZE;
+const PIF_SERIAL_NUM_START_IDX: usize = PIF_MODEL_NUM_END_IDX;
+const PIF_SERIAL_NUM_END_IDX: usize = PIF_SERIAL_NUM_START_IDX + STD_BIX_SERIAL_SIZE;
+const PIF_OEM_INFO_START_IDX: usize = PIF_SERIAL_NUM_END_IDX;
+const PIF_OEM_INFO_END_IDX: usize = PIF_OEM_INFO_START_IDX + STD_BIX_OEM_SIZE;
+
```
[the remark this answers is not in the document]
williampMSFT: similar situation here - was already this way in the messages crate, but got away with it because they happened to have the same value.  Will fix, but we may want to consider consolidating constants if they really are supposed to be in lockstep

kurtjd (dismissed):
LGTM. Though similar nit as last time, just some pub items missing doc strings in the interface and relay lib.rs files.

kurtjd (dismissed):

kurtjd (dismissed):

Cargo.toml
```diff
@@ -55,7 +56,7 @@ unwrap_used = "deny"
 [workspace.dependencies]
 aligned = "0.4"
 anyhow = "1.0"
-battery-service-messages = { path = "./battery-service-messages" }
+battery-service-interface = { path = "./battery-service-interface" }
```
jerrysxie, line 59: Should we add b`attery-service-relay` here as well?

battery-service-interface/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "battery-service-messages"
+name = "battery-service-interface"
```
jerrysxie, line 2: `embedded-services` and `num_enum` are probably not needed anymore.

battery-service-relay/src/serialization.rs
```diff
@@ -0,0 +1,655 @@
+use battery_service_interface::*;
+// BatteryState, Bct, BctReturnResult, BixFixedStrings, Bma, Bmc, Bmd, BmdCapabilityFlags, BmdStatusFlags, Bms, Bpc,
+//     Bps, Bpt, BstReturn, Btm, BtmReturnResult, Btp, PifFixedStrings, PowerThresholdSupport, PowerUnit, PsrReturn,
+//     StaReturn, BmcControlFlags
```
jerrysxie: Delete?

jerrysxie (dismissed):
LGTM, but I only looked at the high-level structure. I will let @tullom focus on the individual battery API calls.

Cargo.toml
williampMSFT: Interesting question - my initial reaction was "oops, definitely yes", but now that I think about it, the only thing that having it here as a workspace dependency does is make it available to other crates in this repo as a dependency, and I think these relay crates are expected to be leaf nodes, so maybe that's not valuable / just increases noise?

It looks like I did for the time-alarm-service-relay crate so will do it for now for consistency, but there might be a case for removing the entries for leaf nodes instead?  Not sure, what do you think?

battery-service-relay/src/serialization.rs
```diff
@@ -0,0 +1,652 @@
+use battery_service_interface::*;
+use embedded_batteries_async::acpi::{BatterySwapCapability, BatteryTechnology, PowerSource, ThresholdId};
+use embedded_services::relay::{MessageSerializationError, SerializableMessage};
+
+// Unfortunately `TryFrom<u32>` is not implemented by embedded-batteries for these types
+/// Attempt to convert a `u32` to a `PowerUnit`.
+fn power_unit_try_from_u32(value: u32) -> Result<PowerUnit, MessageSerializationError> {
```
tullom: with https://github.com/OpenDevicePartnership/embedded-batteries/pull/39, we should have support for these implementations on embedded-batteries on v0.3.4

tullom (dismissed):
Just one nb comment, otherwise looks good! The API looks good

battery-service-relay/src/serialization.rs
williampMSFT: sounds good, bumped to v0.3.4 and removed these

kurtjd (approved):

Cargo.toml
jerrysxie: Leaving it as a workspace dependency makes it available to other crates in this repo, like you say. But if we don't foresee other crates in this repo needing to depend on them. Then I am okay to remove them.

williampMSFT: alright, I'll leave it in for now (updated to include the battery service) and see if we can find any use cases for direct linkage to these relay crates (or any of the other ones we have listed here), and if not I can do a follow-up PR to remove the unused ones. thanks!

jerrysxie (approved):

tullom (approved):

PR #785, opened 2026-04-13
Merge feature/relay-split into v0.2.0

williampMSFT:


Cargo.toml
```diff
@@ -2,9 +2,11 @@
 resolver = "2"
```
williampMSFT, line 2: I think we want to resolve this PR as a merge rather than a squash, so putting this comment here to block autocompletion; will resolve when we have signoff on this and are ready to check in

kurtjd (approved):

tullom (approved):

PR #786, opened 2026-04-13
Merge v0.2.0 into feature/relay-split

williampMSFT:


power-policy-service/tests/common/mod.rs
```diff
@@ -73,47 +73,23 @@ async fn power_policy_task<'device, 'sender, const N: usize>(
     }
```
williampMSFT, line 73: this PR is an attempt to unstick this one
https://github.com/OpenDevicePartnership/embedded-services/pull/785

should also be a merge

PR #790, opened 2026-04-15
power-policy-service: Fix disconnect flow

RobertZ2011:
The current logic is incorrect and any disconnect will make the power policy think the currently connected consumer has disconnected. Also fix a type-C bug where our internal view of the port's power role could mismatch the role reported by hardware.

### DIFF
--- a/power-policy-service/src/lib.rs
+++ b/power-policy-service/src/lib.rs
@@ -79,7 +79,16 @@ impl PowerPolicy {
 
     async fn process_notify_disconnect(&self, device: &device::Device) -> Result<(), Error> {
         self.context.send_response(Ok(policy::ResponseData::Complete)).await;
-        if let Some(consumer) = self.state.lock().await.current_consumer_state.take() {
+
+        self.remove_connected_provider(device.id()).await;
+        if let Some(consumer) = self
+            .state
+            .lock()
+            .await
+            .current_consumer_state
+            .take_if(|d| d.device_id == device.id())
+        {
+            // Current PSU is disconnected, disconnect chargers, and attempt to select next PSU
             info!("Device{}: Connected consumer disconnected", consumer.device_id.0);
             self.disconnect_chargers().await?;
 
@@ -87,10 +96,10 @@ impl PowerPolicy {
                 data: CommsData::ConsumerDisconnected(consumer.device_id),
             })
             .await;
+
+            self.update_current_consumer().await?;
         }
 
-        self.remove_connected_provider(device.id()).await;
-        self.update_current_consumer().await?;
         Ok(())
     }
 
--- a/type-c-service/src/wrapper/power.rs
+++ b/type-c-service/src/wrapper/power.rs
@@ -69,12 +69,14 @@ where
                 return PdError::Failed.into();
             }
         } else if let Ok(state) = power.try_device_action::<action::ConnectedConsumer>().await {
+            // Staying a consumer, but we have updated capabilities
             if let Err(e) = state.notify_consumer_power_capability(available_sink_contract).await {
                 error!("Error setting power contract: {:?}", e);
                 return PdError::Failed.into();
             }
         } else if let Ok(state) = power.try_device_action::<action::ConnectedProvider>().await {
-            // We're no longer providing power, so disconnect as a provider
+            // Transition from provider to consumer.
+            // This handles role swaps from source to sink.
             let Ok(state) = state.disconnect().await else {
                 error!("Error disconnecting as provider");
                 return PdError::Failed.into();
@@ -142,6 +144,7 @@ where
             }
         } else if let Ok(state) = power.try_device_action::<action::ConnectedProvider>().await {
             if let Some(contract) = contract {
+                // Staying a provider, but we have updated capabilities
                 if let Err(e) = state.request_provider_power_capability(contract).await {
                     error!("Error setting power contract: {:?}", e);
                     return PdError::Failed.into();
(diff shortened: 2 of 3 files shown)


type-c-service/src/wrapper/power.rs
```diff
@@ -69,7 +69,18 @@ where
                 return PdError::Failed.into();
             }
         } else if let Ok(state) = power.try_device_action::<action::ConnectedProvider>().await {
-            if let Err(e) = state.notify_consumer_power_capability(available_sink_contract).await {
+            if status.power_role == PowerRole::Sink {
+                // We're no longer providing power, so disconnect as a provider
+                let Ok(state) = state.disconnect().await else {
+                    error!("Error disconnecting as provider");
+                    return PdError::Failed.into();
```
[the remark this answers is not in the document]
RobertZ2011: I think this is fine. We log the current state above so we can see the exact transition that failed.

type-c-service/src/wrapper/power.rs
```diff
@@ -136,6 +153,19 @@ where
                     return PdError::Failed.into();
                 }
             }
+        } else if let Ok(state) = power.try_device_action::<action::ConnectedConsumer>().await {
+            // We're no longer consuming power, so disconnect as a consumer
```
asasine: Not super familiar with the PD spec w.r.t. power role swaps. What's the order of operations between receiving the new contract, updating the power role, and when the direction of current is supposed to swap?

type-c-service/src/wrapper/power.rs
```diff
@@ -136,6 +153,19 @@ where
                     return PdError::Failed.into();
                 }
```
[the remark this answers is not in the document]
RobertZ2011: Done.

power-policy-service/src/lib.rs
```diff
@@ -79,18 +79,27 @@ impl PowerPolicy {
...
+            // Current PSU is disconnected, disconnect chargers, and attempt to select next PSU
             info!("Device{}: Connected consumer disconnected", consumer.device_id.0);
             self.disconnect_chargers().await?;
 
             self.comms_notify(CommsMessage {
                 data: CommsData::ConsumerDisconnected(consumer.device_id),
             })
             .await;
+
+            self.update_current_consumer().await?;
```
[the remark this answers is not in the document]
RobertZ2011: The mutex guard is not kept alive for the duration of the block and calling `update_current_consumer` does not trigger a deadlock.

type-c-service/src/wrapper/power.rs
RobertZ2011: The `PR_Swap` happens first to change the role, then a new contract is negotiated. Current and voltage both drop to zero during the swap before returning to the implicit contract values. The PD controller automatically disables its switches during the swap AFAIK so there shouldn't need to be any special handling there.

type-c-service/src/wrapper/power.rs
```diff
@@ -69,6 +74,12 @@ where
                 return PdError::Failed.into();
             }
         } else if let Ok(state) = power.try_device_action::<action::ConnectedProvider>().await {
```
jerrysxie, line 77: This is the case where we switched from a provider to a consumer?

type-c-service/src/wrapper/power.rs
```diff
@@ -132,6 +149,19 @@ where
             } else {
                 // No longer need to source, so disconnect
                 if let Err(e) = state.disconnect().await {
+                    error!("Error disconnecting as provider: {:?}", e);
+                    return PdError::Failed.into();
+                }
+            }
+        } else if let Ok(state) = power.try_device_action::<action::ConnectedConsumer>().await {
```
jerrysxie, line 159: This is the case where we are switching from a provider to a consumer?

type-c-service/src/wrapper/power.rs
```diff
@@ -132,6 +149,19 @@ where
...
+            // We're no longer consuming power, so disconnect as a consumer
+            let Ok(state) = state.disconnect().await else {
+                error!("Error disconnecting as consumer");
+                return PdError::Failed.into();
+            };
+
+            if let Some(contract) = contract {
```
jerrysxie, line 168: If there is no contact, what does this case imply?

jerrysxie (dismissed):
Looks reasonable but I had some trouble following the logic. More breadcrumbs in these cases will help.

type-c-service/src/wrapper/power.rs
RobertZ2011: Done. Added comments around the state transitions for better clarity.

type-c-service/src/wrapper/power.rs
RobertZ2011: Done.

type-c-service/src/wrapper/power.rs
asasine: So should we actually be disconnecting as a consumer during `PR_Swap`? As opposed to here, when we get the new contract (which I presume is much later after `PR_Swap`).

type-c-service/src/wrapper/power.rs
RobertZ2011: Done.

asasine (approved):
Won't block because crit but we could probably polish up the semantics of exactly when the controller swaps from one power role to another.

RobertZ2011:
> Won't block because crit but we could probably polish up the semantics of exactly when the controller swaps from one power role to another.

I think part of this would be to directly handle power role swap interrupts instead of piggybacking off the new contract logic. I opened a bug for this at #792. It'll likely land in the v0.2.0 branch, but we can backport it if we continue to run into issues around role swaps.

jerrysxie (approved):

PR #793, opened 2026-04-15
Add non-blocking partition access via PartitionGuard

ayodejiige:
### Summary
Introduces a PartitionGuard type to the partition-manager crate that provides both async (lock) and non-blocking (try_lock) access to partition storage, along with hardened arithmetic to prevent integer overflow.

## Motivation
When a task panics while holding a partition's mutex, the lock remains held. If the panic handler itself needs to access the same partition (e.g., to persist crash logs or diagnostic data), calling `lock().await` would deadlock. The handler is waiting on a lock that will never be released. By exposing `try_lock()`, callers like panic handlers can attempt non-blocking access and gracefully handle failure instead of deadlocking the system.

## Changes
- Added `PartitionGuard` struct with `lock()` (async) and `try_lock()` (non-blocking) constructors on Partition
- Implemented `BlockDevice`, `ReadNorFlash`, and `NorFlash` traits for PartitionGuard
- Re-exported `TryLockError` from embassy_sync for caller convenience
- Hardened all partition offset/size arithmetic with checked operations to prevent integer overflow
- Added unit tests covering guard-based read/write/erase, out-of-bounds access, overflow edge cases, and lock contention via `try_lock()`

### DIFF
--- a/partition-manager/partition-manager/src/ext/bdd.rs
+++ b/partition-manager/partition-manager/src/ext/bdd.rs
@@ -28,15 +28,19 @@ impl<F, MARKER, M: RawMutex> Partition<'_, F, MARKER, M> {
         F: BlockDevice<SIZE>,
     {
         const { assert!(SIZE <= u32::MAX as usize) };
+
         let Some(offset) = block_address.checked_mul(SIZE as u32) else {
             return Err(Error::OutOfBounds);
         };
+
         if data.len() > u32::MAX as usize {
             return Err(Error::OutOfBounds);
         }
+
         let Some(size) = (data.len() as u32).checked_mul(SIZE as u32) else {
             return Err(Error::OutOfBounds);
         };
+
         if !self.within_bounds(offset, size) {
             Err(Error::OutOfBounds)
         } else {
@@ -63,15 +67,19 @@ impl<F, MARKER, M: RawMutex> PartitionGuard<'_, F, MARKER, M> {
         F: BlockDevice<SIZE>,
     {
         const { assert!(SIZE <= u32::MAX as usize) };
+
         let Some(offset) = block_address.checked_mul(SIZE as u32) else {
             return Err(Error::OutOfBounds);
         };
+
         if data.len() > u32::MAX as usize {
             return Err(Error::OutOfBounds);
         }
+
         let Some(size) = (data.len() as u32).checked_mul(SIZE as u32) else {
             return Err(Error::OutOfBounds);
         };
+
         if !self.within_bounds(offset, size) {
             Err(Error::OutOfBounds)
         } else {
--- a/partition-manager/partition-manager/src/ext/bdd.rs
+++ b/partition-manager/partition-manager/src/ext/bdd.rs
@@ -111,10 +111,10 @@ impl<const SIZE: usize, F: BlockDevice<SIZE>, M: RawMutex> BlockDevice<SIZE> for
         self.check_access(block_address, data)?;
         let start_block = self.start_block(SIZE as u32).ok_or(Error::NotAligned)?;
 
-        Ok(self
-            .guard
+        self.guard
             .read(start_block.checked_add(block_address).ok_or(Error::OutOfBounds)?, data)
-            .await?)
+            .await
+            .map_err(Error::Inner)
     }
 
     async fn write(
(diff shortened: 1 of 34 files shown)


partition-manager/partition-manager/src/ext/esa.rs
```diff
@@ -26,12 +26,14 @@ impl<F: ReadNorFlash, M: RawMutex> ReadNorFlash for Partition<'_, F, RO, M> {
...
-        if !self.within_bounds(offset, bytes.len() as u32) {
+        if bytes.len() > u32::MAX as usize || !self.within_bounds(offset, bytes.len() as u32) {
             return Err(Error::OutOfBounds);
         }
 
         let mut storage = self.storage.lock().await;
-        Ok(storage.read(offset + self.offset, bytes).await?)
+        Ok(storage
+            .read(offset.checked_add(self.offset).ok_or(Error::OutOfBounds)?, bytes)
+            .await?)
```
diondokter: With the partition guard, it'd be good to do everything through that new construct. Because else we end up with two implementations.

So here I would refactor it to:
```rust
let mut guard = self.lock().await;
guard.read(offset, bytes).await
```

That way we simply forward the impl and have no duplicates

partition-manager/partition-manager/src/lib.rs
```diff
@@ -101,6 +135,18 @@ impl<F, MARKER, M: RawMutex> Partition<'_, F, MARKER, M> {
     }
 }
 
+impl<F, MARKER, M: RawMutex> PartitionGuard<'_, F, MARKER, M> {
+    /// Checks whether an address range lies within the partition.
+    #[allow(unused)]
+    const fn within_bounds(&self, offset: u32, size: u32) -> bool {
```
JamesHuard: Any thoughts on making this pub?

partition-manager/partition-manager/src/ext/bdd.rs
```diff
@@ -50,7 +101,9 @@ impl<const SIZE: usize, F: BlockDevice<SIZE>, M: RawMutex> BlockDevice<SIZE> for
         let start_block = self.start_block(SIZE as u32).ok_or(Error::NotAligned)?;
 
         let mut storage = self.storage.lock().await;
-        Ok(storage.read(start_block + block_address, data).await?)
+        Ok(storage
```
RobertZ2011: nit: There's a couple of places where you wrap the function call in `Ok(...)` is that needed? It looks like you should be able to have just `storage.read(...).await`

partition-manager/partition-manager/src/ext/esa.rs
```diff
@@ -56,22 +58,84 @@ impl<F: NorFlash, M: RawMutex> NorFlash for Partition<'_, F, RW, M> {
...
+
+impl<F: ReadNorFlash, MARKER, M: RawMutex> ErrorType for PartitionGuard<'_, F, MARKER, M> {
+    type Error = Error<F::Error>;
+}
+
+impl<F: ReadNorFlash, MARKER, M: RawMutex> ReadNorFlash for PartitionGuard<'_, F, MARKER, M> {
+    const READ_SIZE: usize = F::READ_SIZE;
+
+    async fn read(&mut self, offset: u32, bytes: &mut [u8]) -> Result<(), Self::Error> {
+        if bytes.len() > u32::MAX as usize || !self.within_bounds(offset, bytes.len() as u32) {
```
RobertZ2011: This comparison appears in a couple of places, would it be possible to introduce a function for it? Or maybe merge this functionality into `within_bounds` and have it take a usize?

RobertZ2011 (changes requested):

partition-manager/partition-manager/src/ext/bdd.rs
ayodejiige: It is needed for error type conversion. `PartitionGuard` and the underlying storage type have different error types even though they implement the same `Error` trait. An alternative and maybe clearer would be to do `storage.read(...).await,map_err(Error::Inner)`

partition-manager/partition-manager/src/lib.rs
ayodejiige: The benefits of that are not clear to me. It may give callers the opportunity to check if a range is within the partition, but:
1. They already have access to the partition size
2. If they make a mistake calling a public api with something out of bounds, they will get an out of bounds error

partition-manager/partition-manager/src/ext/esa.rs
ayodejiige: Good idea! Done

diondokter (approved):
The bits of the others are valid.
For the rest, approved

JamesHuard (approved):

partition-manager/partition-manager/src/ext/bdd.rs
ayodejiige: @RobertZ2011 updated to use `storage.read(...).await,map_err(Error::Inner)`

jerrysxie (dismissed):

partition-manager/partition-manager/src/ext/esa.rs
ayodejiige: updated

RobertZ2011 (approved):

jerrysxie (approved):

PR #797, opened 2026-04-16
Add cargo-vet audit skill and auditor agent

jerrysxie:
Introduce a cargo-vet auditing skill that defines an end-to-end workflow for discovering, planning, delegating, certifying, and verifying dependency audits.

Add a dedicated cargo-vet-auditor agent with a safe-to-deploy checklist and a structured output format to standardize crate security reviews.

### DIFF
--- a/.github/agents/cargo-vet-auditor.agent.md
+++ b/.github/agents/cargo-vet-auditor.agent.md
@@ -1,0 +1,131 @@
+---
+name: cargo-vet-auditor
+description: >
+  Specialized agent for auditing individual Rust crates against cargo-vet
+  safe-to-deploy criteria. Delegates from the cargo-vet-audit skill to
+  review crate source code, diffs, and build scripts for supply chain safety.
+tools:
+  - execute
+  - read
+  - grep
+  - glob
+---
+
+# Cargo-Vet Crate Auditor
+
+You are a specialized Rust crate auditor. Your job is to review a single crate's
+source code or diff and determine whether it meets the `safe-to-deploy` criteria.
+
+## The `safe-to-deploy` Criteria (Official Definition)
+
+> This crate will not introduce a serious security vulnerability to production
+> software exposed to untrusted input.
+>
+> Auditors are not required to perform a full logic review of the entire crate.
+> Rather, they must review enough to fully reason about the behavior of all unsafe
+> blocks and usage of powerful imports. For any reasonable usage of the crate in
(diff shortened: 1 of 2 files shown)


.github/agents/cargo-vet-auditor.agent.md
```diff
@@ -0,0 +1,131 @@
...
+- [ ] Identify ALL `unsafe` blocks and `unsafe fn` declarations
+- [ ] For each: verify soundness (no UB for any valid input)
+- [ ] Check for `unsafe impl` of traits (Send, Sync, etc.) — verify invariants hold
+- [ ] Check for `#![allow(unsafe_op_in_unsafe_fn)]` — note if present (transitional vs permanent)
+- [ ] Look for `transmute`, raw pointer derefs, `from_raw`, `as_ptr` patterns
+
+### 2. Build Scripts (`build.rs`)
+- [ ] Does a build.rs exist?
+- [ ] Does it access the filesystem beyond `OUT_DIR` and standard env vars?
+- [ ] Does it make network requests?
```
rogurr, line 57: Would the build script ever download something that might have a hash that needs to be checked?  Does this agent file need to state all downloads are flagged, or state to allow safe/measured downloads?

rogurr (approved):

.github/agents/cargo-vet-auditor.agent.md
jerrysxie: For Rust at least, the build process is very source-centric. Typically, the build script should not download anything. If it does, then that would be a very suspicious behavior.

felipebalbi: I would disagree with that statement. While not common, it's totally plausible e.g. for a PAC to download SVDs and generate the PAC during compilation. Depending on how the vendor licenses their SVD repository, this might be the only way to generate a PAC.

In general, sure, network accesses at `build.rs` time is _suspicious_, but I think @rogurr's comment might be valid.

felipebalbi (dismissed):

.github/agents/cargo-vet-auditor.agent.md
jerrysxie: Yeah, currently the agent will mark it as a suspicious and ask human to review. Maybe we can add some more detailed language to make the agent does this.

kurtjd (approved):

rogurr (approved):

RobertZ2011 (approved):

Issue #779, opened 2026-04-07
Remove intrusive list charger registration


tullom:
The charger is currently registered using intrusive list. In line with the v0.2.0 refactor, change this over to use the new power-policy registration struct.

Issue #761, opened 2026-03-25
Migrate power policy charger code for v0.2.0


RobertZ2011:
* Charger calls should be done through direct function calls
* Remove intrusive list charger registration

PR #798, opened 2026-04-16
power-policy charger: remove intrusive list and move to array registration

### LINKED ISSUE #778: Move charger API over to direct function calls
In the power policy, the charger has an API implemented with message passing. In line with the refactor work done in type-c and power policy, refactor charger API calls to be executed through direct function calls.

tullom:
power-policy: Charger v0.2.0 refactor

- Moved away from intrusive lists for storing chargers, instead using an array based approach
- Switched to PSU style charger event receivers for receiving events
- Added charger event types and structures in power-policy-interface.
- Introduced charger state management with transitions based on PSU state changes.
- **OEM now handles charger state transitions by calling the associated state methods to drive the charger state forward**
  - One exception to this is when the power policy gets a new contract, but the charger is still unpowered. The charger will be forced into an initialized state.
- Created tests for charger state transitions and event handling.

Resolves #778 , #779 , and #761


power-policy-interface/src/charger/mock.rs
```diff
@@ -0,0 +1,48 @@
+use embassy_sync::mutex::Mutex;
```
RobertZ2011, line 1: I think we can deduplicate this and the mock_charger one and provide a single NoopCharger implementation in `power-policy-interface`.

power-policy-interface/src/charger/event.rs
```diff
@@ -0,0 +1,50 @@
...
+            true => PsuState::Attached,
+            false => PsuState::Detached,
+        }
+    }
+}
+
+/// Data for a charger event
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum EventData {
```
RobertZ2011, line 28: I'm not sure if we need most of these events. I think `PsuStateChange` is useful as a general event, but it looks like the others are used to have the power policy implementation drive the state machine forward. We should avoid a round trip to the service and have the charger update its own state.

power-policy-interface/src/charger/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl From<ChargerError> for crate::psu::Error {
+    fn from(value: ChargerError) -> Self {
+        Self::Charger(value)
+    }
+}
+
+/// Data for a device request
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
RobertZ2011: Do we still need `PolicyEvent`, `CharagerResponseData`, and `ChargerResponse`? These look like leftovers of the previous message implementation.

power-policy-service/src/service/task.rs
```diff
@@ -27,3 +28,24 @@ pub async fn task<
         }
     }
 }
+
```
RobertZ2011, line 31: I like the separation of tasks, but for convenience we should have a `task` function that is a join over the two sub-functions.

RobertZ2011 (changes requested):

.github/copilot-instructions.md
```diff
@@ -0,0 +1,95 @@
+# Project Guidelines
```
jerrysxie: Can we put this in a different PR?

power-policy-service/src/charger.rs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// Create a new instance
+    pub fn new(charger_devices: [&'a CHARGER; N], receivers: [R; N]) -> Self {
+        Self {
+            charger_devices,
+            receivers,
+        }
+    }
+
+    /// Get the next pending PSU event
```
jerrysxie: Charger event?

power-policy-interface/src/charger/event.rs
tullom: Technically all these events are used to signal state changes, including `PsuStateChange` (which just moves the state between `Attached` and `Detached`). I did this so that the OEM doesn't have to worry about changing the state, however I do see the roundaboutness of this approach. I'm also worried of the OEM transitioning to an invalid state, however i did provide methods on `State` to drive the state forward so there should be guardrails.

If we want the OEM to handle the state transitions, we don't need these events, which means we don't need the receivers. The power policy can get the state of the charger at any time by calling the `state()` trait method on the devices in the `Registration` struct. Thoughts?

RobertZ2011: I think we should let the OEM handle the state transitions. It is nice to centralize that logic in the power policy, but I think putting that on the driver/OEM is a more flexible approach. A big benefit that I see is that the power policy can't do much if there's an issue with a state transition. Any recovery logic is going to be implementation specific, so it makes sense for the implementation to handle the state. I'll have to create a PR to do the same for the PSU code.

We should also keep the receiver infrastructure in place though to help with future compatibility. The `PsuStatusChanged` event is functionality that isn't really exposed currently. I think it's generally useful to provide that even if ODP might not need it right now.

power-policy-service/src/charger.rs
tullom: good catch. changed to charger.

.github/copilot-instructions.md
tullom: Whoops, dont know how that was checked in. Removed.

power-policy-service/src/service/task.rs
tullom: Done.

power-policy-interface/src/charger/mod.rs
tullom: Good catch, removed.

power-policy-interface/src/charger/event.rs
tullom: Makes sense, the OEM now handles its own state transitions using the `charger::State` struct in power policy. I kept the receiver functionality in with the `PsuStatusChanged` event as the only event, and it currently no-ops. In the future if we want to add more events we'll have the infrastructure set up.

power-policy-interface/src/charger/mock.rs
tullom: Done.

RobertZ2011 (dismissed):

kurtjd (dismissed):

felipebalbi (dismissed):

seanbrns (dismissed):

thdesil (approved):

seanbrns (approved):

kurtjd (approved):

PR #810, opened 2026-04-20
.github: Pull-in shared copilot instructions

RobertZ2011:


kurtjd:
I guess there isn't an easy way to have like a soft-link to https://github.com/OpenDevicePartnership/odp-platform-common/blob/main/.github/copilot-instructions.md here? Worried about copying and pasting since that's a lot of maintenance burden any time the common instructions change.

RobertZ2011:
> I guess there isn't an easy way to have like a soft-link to https://github.com/OpenDevicePartnership/odp-platform-common/blob/main/.github/copilot-instructions.md here? Worried about copying and pasting since that's a lot of maintenance burden any time the common instructions change.

That's what I gathered from the discussion: there's not really a good way to pull it in directly right now and we're relying on copying/pasting at the moment.

kurtjd (approved):

jerrysxie (approved):

PR #811, opened 2026-04-21
V0.2.0 mergeback: 04-20-2026

tullom:


williampMSFT (dismissed):

jerrysxie (changes requested):
I am going to assume that we want to do a merge commit. Can we do a cherry-pick for this commit from main instead of creating another commit: 

Image

tullom:
> I am going to assume that we want to do a merge commit. Can we do a cherry-pick for this commit from main instead of creating another commit:
> Image

Looks like the cherry-pick doesn't resolve the merge conflict, even when I resolve the merge conflict locally before pushing. We will need to either rebase or merge, so i think there isnt any getting away from the second merge commit.

tullom:
> > I am going to assume that we want to do a merge commit. Can we do a cherry-pick for this commit from main instead of creating another commit:
> > Image
> 
> Looks like the cherry-pick doesn't resolve the merge conflict, even when I resolve the merge conflict locally before pushing. We will need to either rebase or merge, so i think there isnt any getting away from the second merge commit.

As discussed offline, reverted back to a merge commit.

jerrysxie (approved):

jerrysxie:
@tullom We should consider bypass policy for this once we have 2 approvals.

kurtjd (approved):

RobertZ2011 (approved):

PR #813, opened 2026-04-21
power-policy-service: Shorten lock held in disconnect flow

RobertZ2011:
This lock was being held for the duration of the conditional branch, causing a deadlock in `update_current_consumer`.

### DIFF
--- a/power-policy-service/src/lib.rs
+++ b/power-policy-service/src/lib.rs
@@ -81,13 +81,13 @@ impl PowerPolicy {
         self.context.send_response(Ok(policy::ResponseData::Complete)).await;
 
         self.remove_connected_provider(device.id()).await;
-        if let Some(consumer) = self
+        let current_consumer = self
             .state
             .lock()
             .await
             .current_consumer_state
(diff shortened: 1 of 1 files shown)


jerrysxie (approved):

chchongmsft (approved):

power-policy-service/src/lib.rs
```diff
@@ -81,13 +81,13 @@ impl PowerPolicy {
         self.context.send_response(Ok(policy::ResponseData::Complete)).await;
 
         self.remove_connected_provider(device.id()).await;
-        if let Some(consumer) = self
```
jerrysxie, line 84: Did we make the some change in v0.2.0 branch? Can this be caught with a unit test?

jerrysxie (approved):

power-policy-service/src/lib.rs
RobertZ2011: The v0.2.0 branch wasn't affected by this, and the existing tests should catch this if it does happen.

PR #821, opened 2026-05-04
Type-C: Trait refactor and clean-up

RobertZ2011:
Refactor traits and do some general clean-up.

* Move `Controller` trait into its own module.
* Remove bus error type from traits to make it easier to abstract over PD hardware.
* Break-up `Controller` trait into multiple traits.
* Introduce corresponding port-level traits.


type-c-service/src/controller/mod.rs
```diff
@@ -160,15 +159,12 @@ impl<
                 event,
             })
             .await
-            .map_err(Error::Pd)?;
+            .map_err(|_| PdError::Failed)?;
```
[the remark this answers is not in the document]
RobertZ2011: Done.

type-c-interface/src/control/pd.rs
```diff
@@ -0,0 +1,84 @@
+//! Control types for code PD functionality
```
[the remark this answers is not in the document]
RobertZ2011: Done.

type-c-interface/src/controller/mod.rs
```diff
@@ -0,0 +1,22 @@
+//! Module for PD controller related code
+//! use embedded_usb_pd::{LocalPortId, PdError};
```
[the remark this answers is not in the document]
RobertZ2011: Done.

type-c-service/src/bridge/mod.rs
```diff
@@ -31,178 +44,91 @@ impl<'device, Controller: Lockable<Inner: port::Controller>> Bridge<'device, Con
...
-                },
-            },
+            port::PortCommandData::RetimerFwUpdateGetState => controller
+                .get_rt_fw_update_status(local_port)
+                .await
+                .map(port::PortResponseData::RtFwUpdateStatus),
+            port::PortCommandData::RetimerFwUpdateSetState => controller
+                .set_rt_fw_update_state(local_port)
+                .await
+                .map(|_| port::PortResponseData::Complete),
```
jerrysxie, line 54: I just noticed this. For this function, even before the change, even if we get an error, we will throw it away here. Is that just a temporary thing?

type-c-interface/src/controller/power.rs
```diff
@@ -0,0 +1,14 @@
+use embedded_usb_pd::{LocalPortId, PdError};
+
+/// System power state related controller functionality
+pub trait SystemPowerState {
```
jerrysxie: Can we name the trait different from the enum? This is a bit confusing.

type-c-interface/src/control/vdm.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// initiating a VDM sequence
+    pub initiator: bool,
+    /// VDO count
+    pub vdo_count: u8,
+    /// VDO data
+    pub vdo_data: [u32; MAX_NUM_DATA_OBJECTS],
+}
+
+impl SendVdm {
+    /// Create a new blank port status
```
jerrysxie: Blank VDM?

jerrysxie (changes requested):

type-c-interface/src/controller/power.rs
RobertZ2011: Renamed the trait to `SystemPowerStateStatus`.

type-c-interface/src/control/vdm.rs
RobertZ2011: Done.

type-c-service/src/bridge/mod.rs
RobertZ2011: I don't think this was intentional, but this code will be removed in the next PR.

jerrysxie (approved):

kurtjd (approved):

tullom (approved):

PR #823, opened 2026-05-05
Add mctp-rs as in-tree workspace member (port of dymk/mctp-rs @ 3d941ba)

dymk:
## Summary

Imports `dymk/mctp-rs` as an in-tree workspace member at `mctp-rs/` and flips the existing `[workspace.dependencies]` git pin to a path dep. The imported source is at SHA `3d941ba5205ca7781bf37e3dc7c5dfdc99a082d6`, which matches the pre-edit `Cargo.lock` pin exactly — so all four `mctp-rs` consumers (`uart-service`, `espi-service`, `embedded-service`, plus its macro re-exports) compile and test unchanged.

## What changed

- New `mctp-rs/` directory: snapshot copy of `dymk/mctp-rs` `main @ 3d941ba` source. No git history preserved.
- Root `Cargo.toml`: `"mctp-rs"` appended to `[workspace] members`; `mctp-rs = { git = "..." }` replaced with `mctp-rs = { path = "./mctp-rs" }`.
- `Cargo.lock` regenerated.
- `mctp-rs/Cargo.toml`: dropped `cargo-husky` from `[dev-dependencies]` (its build script searches for `.cargo-husky/hooks/` at the workspace root, which is now this repo, not the crate).
- `mctp-rs/rustfmt.toml` removed; source reformatted under this repo's `rustfmt.toml` (`max_width = 120`).
- Excluded from the copy: `.cargo-husky/`, `.github/`, `.vscode/`, source `Cargo.lock`, source `.gitignore`. A follow-up PR will add the dedicated `mctp-rs` CI workflow.

## Why

`mctp-rs` lived in a personal fork and was already consumed here via a git pin. Bringing it in-tree as a sibling of `uart-service` etc. removes a layer of cross-repo indirection and lets future changes flow through normal workspace PR review. SHA `3d941ba` was chosen specifically because it matches the consumer pin — so this PR has zero behavioral surface.

The mctp-rs repo holds further commits beyond `3d941ba` that change the `MctpMedium` API. Those, together with the corresponding consumer updates and the eventual archival of the personal fork, land in a separate PR after this one.

## Verifying locally

```bash
git fetch origin dymk/phase-31-mctp-rs-in-tree
git switch dymk/phase-31-mctp-rs-in-tree

cargo build --workspace
cargo test --workspace                 # 178 pass / 0 fail
cargo fmt --all --check                # clean
cargo clippy --workspace --all-targets -- -D warnings   # clean
```

Spot-check the source is byte-identical to upstream:

```bash
git -C <local-clone-of-dymk/mctp-rs> show 3d941ba:src/lib.rs           | diff - mctp-rs/src/lib.rs
git -C <local-clone-of-dymk/mctp-rs> show 3d941ba:src/medium/smbus_espi.rs | diff - mctp-rs/src/medium/smbus_espi.rs
```

(Both empty diffs, modulo the rustfmt-style reflow noted above for the second one.)

## Notes for reviewer

- `cargo test --workspace` count went from 128 → 178. The 50-test delta is `mctp-rs`'s own test suite, which now runs as part of the workspace. Every consumer test result line from before is preserved exactly.
- `Cargo.lock` picks up `mctp-rs`'s dev-dep chain (`rstest`, `tokio`, `pretty_assertions`, etc.) for the same reason — these are dev-only and don't affect runtime. The two production-tree changes are `hashbrown 0.15.5 → 0.17.0` and `indexmap 2.11.0 → 2.14.0`; both are pulled in only by the `partition-manager-generation` build script (via `toml`), so no runtime impact.
- The PR is intentionally split across four atomic commits: source import, workspace bootstrap, cargo-husky drop, rustfmt.toml drop + reformat. Each is reviewable in isolation.


mctp-rs/LICENSE.md
```diff
@@ -0,0 +1,7 @@
+Copyright (c) 2025 Microsoft
```
felipebalbi: It looks like the new orientation is gearing towards "Open Device Partnership and Contributors"

@rogurr, has the discussion settled?

RobertZ2011 (approved):

kurtjd (approved):

PR #833, opened 2026-05-06
type-c-service: Migrate to async function calls

RobertZ2011:
* Remove bridge code and last bit of messaging
* Final move away from `Context` type, application code is now responsible for creating event senders/receivers
* Remove almost all `GlobalPortId` usage


type-c-service/src/service/ucsi.rs
```diff
@@ -128,21 +139,29 @@ impl Service<'_> {
...
-                    let port_status = self.state.port_status.get(raw_port).ok_or(PdError::InvalidPort)?;
+                    let port_status = self
+                        .registration
+                        .ports()
+                        .get(raw_port)
+                        .ok_or(PdError::InvalidPort)?
+                        .lock()
+                        .await
+                        .get_port_status()
+                        .await?;
```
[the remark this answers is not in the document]
RobertZ2011: Done.

type-c-service/src/service/ucsi.rs
```diff
@@ -175,7 +187,7 @@ impl Service<'_> {
     }
 
     /// Process a UCSI command
-    pub async fn process_ucsi_command(&mut self, command: &GlobalCommand) -> UcsiResponse {
+    pub async fn process_ucsi_command(&mut self, port: &'port Reg::Port, command: &GlobalCommand) -> UcsiResponse {
```
jerrysxie: Does the client call this? Is the assumption that command will always be for the port passing in? Like the GlobalPortId within GlobalCommand will match the port referece passed in?

type-c-service/src/service/power.rs
```diff
@@ -28,25 +31,27 @@ impl Service<'_> {
...
+                        "Setting port ({}) to constrained, all others unconstrained",
+                        unconstrained_port.lock().await.name()
                     );
-                    for port_index in 0..num_ports {
-                        self.context
-                            .set_unconstrained_power(GlobalPortId(port_index as u8), port_index != unconstrained_index)
+                    for port in self.registration.ports().iter() {
+                        port.lock()
+                            .await
+                            .set_unconstrained_power(ptr::eq(*port, unconstrained_port))
```
jerrysxie: The logic seems to have inverted from the previous behavior:
* If it is the unconstrained port, then it is true, so it is unconstrained.
* If it is not the unconstrained port, then it is false, so it is constrained

jerrysxie (changes requested):

type-c-service/src/service/ucsi.rs
RobertZ2011: Done. You're right that this is something that doesn't need to be an argument.

type-c-service/src/service/power.rs
RobertZ2011: Good catch!

kurtjd (approved):

tullom (approved):

jerrysxie (approved):

PR #843, opened 2026-05-13
type-c-service: Create integration test framework 

### LINKED ISSUE #828: Type-C: Create testing framework
Create any framework for unit and integration tests.

RobertZ2011:


type-c-service/tests/common/mod.rs
```diff
@@ -0,0 +1,400 @@
...
+
+macro_rules! define_port {
+    ($name:ident, $mock_name:expr, $port_name:expr, $config:expr, $local_id:expr) => {
+        let port0_type_c_channel: Channel<
+            GlobalRawMutex,
+            type_c_interface::service::event::PortEventData,
+            CHANNEL_SIZE,
+        > = Channel::new();
+        paste! { let [<$name _type_c_sender>] = port0_type_c_channel.dyn_sender(); }
+        paste! { let [<$name _type_c_receiver>] = port0_type_c_channel.dyn_receiver(); }
```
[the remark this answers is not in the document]
RobertZ2011: Done.

type-c-service/tests/common/mod.rs
```diff
@@ -0,0 +1,400 @@
...
+    define_port!(port1, "mock1", "port1", port_config[1], LocalPortId(0));
+    let PortComponents {
+        port: port1,
+        type_c_receiver: port1_type_c_receiver,
+        power_policy_receiver: port1_power_policy_receiver,
+        mock: port1_mock,
+    } = port1;
+
+    define_port!(port2, "mock2", "port2", port_config[2], LocalPortId(0));
+    let PortComponents {
```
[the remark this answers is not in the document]
RobertZ2011: These indicate the local port ID on the controller. These ports are implemented as separate controllers, so this is correct.

type-c-service/tests/common/mod.rs
```diff
@@ -0,0 +1,400 @@
...
+
+impl<'port, 'ch> Sender<power_policy_interface::service::event::Event<'port, PortMutexType<'port, 'ch>>>
+    for PowerPolicyServiceEventRouter<'port, 'ch>
+{
+    fn try_send(
+        &mut self,
+        event: power_policy_interface::service::event::Event<'port, PortMutexType<'port, 'ch>>,
+    ) -> Option<()> {
+        self.test_sender.try_send(event).ok()?;
+        self.type_c_sender.try_send(event.into()).ok()
```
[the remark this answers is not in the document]
RobertZ2011: This is fine as is. A missing event would cause a test failure so it would be detected either way.

type-c-service/tests/common/mod.rs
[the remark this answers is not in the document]
RobertZ2011: Done.

type-c-service/tests/common/mod.rs
```diff
@@ -0,0 +1,400 @@
...
+    mut test: impl Test,
+) {
+    // Tokio runs tests in parallel, but logging is global so we need to run tests sequentially to avoid interleaved logs.
+    static TEST_MUTEX: OnceLock<Mutex<GlobalRawMutex, ()>> = OnceLock::new();
+    let test_mutex = TEST_MUTEX.get_or_init(|| Mutex::new(()));
+    let _lock = test_mutex.lock().await;
+
+    // Initialize logging, ignore the error if the logger was already initialized by another test.
+    let _ = env_logger::builder().filter_level(log::LevelFilter::Info).try_init();
+    embedded_services::init().await;
```
jerrysxie: Do we actually need anything initialized by `embedded_services::init().await;`?

type-c-service/src/service/power.rs
```diff
@@ -77,13 +77,17 @@ impl<'a, Reg: Registration<'a>> Service<'a, Reg> {
             PowerPolicyEventData::ConsumerDisconnected => {
                 self.ucsi.psu_connected = false;
                 // Notify OPM because this can affect battery charging capability status
-                self.pend_ucsi_connected_ports().await;
+                if self.ucsi.notifications_enabled.battery_charge_change() {
```
jerrysxie, line 80: Is this a bug fix that is separate from mock test framework? We should put this change in a different PR.

type-c-interface-mocks/src/controller/pd.rs
```diff
@@ -0,0 +1,111 @@
...
+            .push_back(ControllerFnCall::Pd(FnCall::EnableSinkPath(port, enable)));
+        self.next_result_enable_sink_path.unwrap_or(Ok(()))
+    }
+
+    async fn get_pd_alert(&mut self, port: LocalPortId) -> Result<Option<Ado>, PdError> {
+        self.fn_calls.push_back(ControllerFnCall::Pd(FnCall::GetPdAlert(port)));
+        self.next_result_get_pd_alert.unwrap_or(Ok(None))
+    }
+
+    async fn set_unconstrained_power(&mut self, port: LocalPortId, unconstrained: bool) -> Result<(), PdError> {
```
jerrysxie, line 72: Should the mock method calling `expect()` for next result also have a sane default as well? Why are them different?

jerrysxie (changes requested):

type-c-service/tests/common/mod.rs
RobertZ2011: Done.

type-c-service/src/service/power.rs
RobertZ2011: Done: https://github.com/OpenDevicePartnership/embedded-services/pull/851. I'll rebase when that lands.

type-c-interface-mocks/src/controller/pd.rs
RobertZ2011: I ended up moving everything to panic because it helps with test development. I also switched to using `VecDeque` to allow mocking more complicated behavior. I think `DpStatus` is something that could use a default, but for VDM-related functions there's not really anything that would make sense to return here because, by their nature, VDMs are application specific.

jerrysxie (dismissed):

jerrysxie (approved):

tullom (approved):

PR #852, opened 2026-05-19
Make uart-service::Service generic over MctpMedium; delete fake-PEC strips

dymk:
# Make `uart-service::Service` generic over `MctpMedium`; delete fake-PEC strips

## Summary

Makes `uart-service::Service` genuinely generic over `mctp_rs::MctpMedium`, so the same service implementation works for both the existing SmbusEspi medium and the DSP0253-serial medium. Adds one new method to the `MctpMedium` trait — `frame_complete(buf) -> Result<Option<usize>, ...>` — that lets generic byte-stream consumers detect frame boundaries without knowing the medium's framing details. Deletes three fake-PEC-strip workarounds that were SmbusEspi-shaped accidents (the real PEC byte is already handled by `SmbusEspiMedium`'s `serialize`/`deserialize` via the `smbus_pec` crate).

Also adds a small additive `MctpMessageBuffer::body()` + `message_type()` accessor pair so callers can read raw body bytes without going through `MctpMessage::parse_as`. Useful for a downstream SP-side service that parses a small fixed-layout payload and doesn't want to define a full `MctpMessageTrait` impl just to access the body slice.

## API changes

`MctpMedium` trait gains a required `frame_complete(buf) -> Result<Option<usize>, ...>` method. Both shipped media (`SmbusEspiMedium`, `MctpSerialMedium`) ship impls in this PR. Out-of-tree implementors (none known) need to add a `frame_complete` impl.

`uart-service::Service::new(relay)` is now `Service::<R, M>::new(relay, medium, reply_context)`. The `Error` enum is now `Error<M: MctpMedium>` to surface medium-specific MCTP errors. Existing SmbusEspi callers migrate via a one-line swap:

```rust
// Before:
let service: Service<MyRelay> = Service::new(relay).unwrap();

// After:
let service: DefaultService<MyRelay> = DefaultService::default_smbusespi(relay).unwrap();
```

`MctpReplyContext<M>` now derives `Clone, Copy` (additive — required for storing it as a field on `Service`).

## What's added

- `MctpMedium::frame_complete(buf) -> Result<Option<usize>, Self::Error>` — sync, byte-stream-agnostic frame-completeness predicate. Supports both length-prefix framing (SmbusEspi: 4-byte header → `byte_count` → body + PEC) and sentinel framing (DSP0253: scan for `0x7E` end-flag).
- `Service<R, M: MctpMedium + Copy>` — genuinely generic over medium. `Service::new(relay, medium, reply_context)` takes medium-specific addressing as a constructor arg.
- `DefaultService<R> = Service<R, SmbusEspiMedium>` type alias + `DefaultService::default_smbusespi(relay)` constructor that bakes in the existing `SmbusEspiReplyContext { destination_slave_address: 1, source_slave_address: 0 }`.
- `MctpMessageBuffer::body() -> &[u8]` + `MctpMessageBuffer::message_type() -> u8` accessors (additive; no existing API removed).

## What's removed

- `uart-service/src/lib.rs` — `// Last byte is PEC, ignore for now` strip. Obviated by `frame_complete`-driven framing (the medium reports the exact frame length, including the PEC byte; `MctpPacketContext::deserialize_packet` consumes the PEC correctly).
- `espi-service/src/espi_service.rs` — `// TODO: workaround because mctp_rs expects a PEC byte, so we hardcode a 0` block.
- `espi-service/src/espi_service.rs` — `// Last byte is PEC, ignore for now` strip in `serialize_packet_from_subsystem`.
- Now-unused constants `SMBUS_HEADER_SIZE` and `SMBUS_LEN_IDX` in `uart-service/src/lib.rs`.

## Consumer migrations (NOT in this PR)

The three in-tree consumers (`dev-qemu`, `dev-imxrt`, `dev-npcx`) live in a separate repo and update their `Service::new(relay)` call sites to `DefaultService::default_smbusespi(relay)` in a follow-up PR there.

## Test results

`cargo test --workspace` consumer pass count is unchanged (no new tests added by this PR; the `frame_complete` impls are covered indirectly by existing media round-trip tests).

### DIFF
--- a/mctp-rs/src/medium/serial.rs
+++ b/mctp-rs/src/medium/serial.rs
@@ -293,6 +293,27 @@ impl MctpMedium for MctpSerialMedium {
 
         Ok(&buffer[..end_pos + 1])
     }
+
+    fn frame_complete(&self, buf: &[u8]) -> MctpPacketResult<Option<usize>, Self> {
+        // This medium's `serialize` emits `[revision, byte_count, ...stuffed body,
+        // fcs_msb, fcs_lsb, 0x7E]` — only an END flag, no leading START flag.
+        //
+        // The `body_start` check below is forward-defensive: DSP0253 variants are
+        // permitted to emit a leading 0x7E flag, and this lets `frame_complete`
+        // accept that case too (skip the leading flag, then find the next 0x7E).
+        // Either pattern returns the full frame length to the caller.
+        if buf.is_empty() {
+            return Ok(None);
+        }
+        let body_start = if buf[0] == END_FLAG { 1 } else { 0 };
+        if body_start >= buf.len() {
+            return Ok(None);
+        }
+        match buf[body_start..].iter().position(|&b| b == END_FLAG) {
+            Some(idx) => Ok(Some(body_start + idx + 1)),
+            None => Ok(None),
+        }
+    }
 }
 
 #[cfg(test)]
@@ -712,5 +733,61 @@ mod medium_tests {
         }
         assert!(packet_count >= 2, "expected multi-packet split, got {packet_count}");
         assert_eq!(total_decoded_body, payload.len());
+    }
+
+    // ----- frame_complete tests -----
+
+    #[test]
+    fn frame_complete_empty_buf_returns_none() {
+        assert_eq!(MctpSerialMedium.frame_complete(&[]).unwrap(), None);
+    }
+
+    #[test]
+    fn frame_complete_no_end_flag_returns_none() {
+        // Bytes present but no 0x7E anywhere → partial frame
+        let buf = [0x01, 0x05, 0xAA, 0xBB, 0xCC];
+        assert_eq!(MctpSerialMedium.frame_complete(&buf).unwrap(), None);
+    }
+
+    #[test]
+    fn frame_complete_only_end_flag_returns_none() {
+        // A lone 0x7E is treated as a leading flag (skipped); after
+        // skipping there are no bytes left, so partial.
+        let buf = [END_FLAG];
+        assert_eq!(MctpSerialMedium.frame_complete(&buf).unwrap(), None);
+    }
+
+    #[test]
+    fn frame_complete_no_leading_flag_finds_end_flag() {
+        // [ revision, byte_count, ...body, fcs_msb, fcs_lsb, 0x7E ]
+        // matches what MctpSerialMedium::serialize emits — no leading flag.
+        let buf = [SERIAL_REVISION, 0x03, 0xAA, 0xBB, 0xCC, 0xDD, 0xEE, END_FLAG];
+        assert_eq!(MctpSerialMedium.frame_complete(&buf).unwrap(), Some(8));
+    }
+
+    #[test]
+    fn frame_complete_with_leading_flag_skips_and_finds_end_flag() {
+        // Forward-defensive: some DSP0253 variants emit a leading START
+        // flag. frame_complete should skip the leading 0x7E and find the
+        // trailing one, returning the FULL length (including leading flag).
+        let buf = [END_FLAG, SERIAL_REVISION, 0x02, 0xAA, 0xBB, 0xCC, 0xDD, END_FLAG];
+        assert_eq!(MctpSerialMedium.frame_complete(&buf).unwrap(), Some(8));
+    }
+
+    #[test]
+    fn frame_complete_extra_bytes_after_end_flag_returns_first_frame_len() {
+        // First frame is 6 bytes (incl. END flag); 2 trailing bytes belong
+        // to the next frame (caller's problem).
+        let buf = [SERIAL_REVISION, 0x01, 0xAA, 0xBB, 0xCC, END_FLAG, 0x99, 0x88];
+        assert_eq!(MctpSerialMedium.frame_complete(&buf).unwrap(), Some(6));
+    }
+
+    #[test]
+    fn frame_complete_back_to_back_after_leading_flag() {
+        // Edge case: leading flag, then payload, then end flag, then more
+        // bytes. Should return length of first frame INCLUDING the leading
+        // flag; trailing bytes belong to a subsequent frame.
+        let buf = [END_FLAG, SERIAL_REVISION, 0x01, 0xAA, 0xBB, END_FLAG, 0x77];
+        assert_eq!(MctpSerialMedium.frame_complete(&buf).unwrap(), Some(6));
     }
 }
--- a/uart-service/src/lib.rs
+++ b/uart-service/src/lib.rs
@@ -1,8 +1,9 @@
 //! uart-service
 //!
-//! To keep things consistent with eSPI service, this also uses the `SmbusEspiMedium` (though not
-//! strictly necessary, this helps minimize code changes on the host side when swicthing between
-//! eSPI or UART).
+//! UART transport for MCTP packets, generic over [`mctp_rs::MctpMedium`].
+//! Use [`DefaultService`] for the SmbusEspi-medium baseline; use
+//! [`Service::new`] directly with another medium (e.g. DSP0253 serial)
+//! for non-SmbusEspi callers.
 //!
 //! Revisit: Will also need to consider how to handle notifications (likely need to have user
 //! provide GPIO pin we can use).
@@ -16,14 +17,12 @@ use embedded_io_async::Write as UartWrite;
 use embedded_services::GlobalRawMutex;
 use embedded_services::relay::mctp::{RelayHandler, RelayHeader, RelayResponse};
 use embedded_services::trace;
-use mctp_rs::smbus_espi::SmbusEspiMedium;
-use mctp_rs::smbus_espi::SmbusEspiReplyContext;
+use mctp_rs::MctpMedium;
+use mctp_rs::smbus_espi::{SmbusEspiMedium, SmbusEspiReplyContext};
 
 // Should be as large as the largest possible MCTP packet and its metadata.
 const BUF_SIZE: usize = 256;
 const HOST_TX_QUEUE_SIZE: usize = 5;
-const SMBUS_HEADER_SIZE: usize = 4;
-const SMBUS_LEN_IDX: usize = 2;
 
 #[derive(Clone)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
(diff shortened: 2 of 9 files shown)


kurtjd (commented):
Looks good, and you said there will be a follow up PR in odp-embedded-controller, but has this been validated with the ec-test-tui/cli using the serial source (e.g.: https://github.com/OpenDevicePartnership/odp-platform-common/blob/main/ec/test-lib/src/serial.rs)?

If it doesn't work and require changes, you will need a simultaneous PR put up over there to make sure they agree. I would like to have PRs up here, in odp-emebdded-controller, and in odp-platform-common to make sure verything is updated simultaneously.

uart-service/src/task.rs
```diff
@@ -3,23 +3,30 @@ use embedded_io_async::Read as UartRead;
...
     // but there are concerns around async cancellation here at least for UART service.
     //
     // Thus this assumes services will only send messages in response to requests from the host,
     // so we handle this in order.
+    //
+    // Note: the in-loop `error!` log was simplified to a static string after making this fn
+    // generic over `M`. The previous `"... {:?}", e` form requires `Error<M>: defmt::Format`,
+    // which can't be expressed as a feature-gated where-clause on stable Rust
+    // (rust-lang/rust#115590). The error variant detail is still visible to the caller via
+    // the `panic!("uart-service error: {:?}", e)` in dev-qemu's task wrapper.
```
kurtjd: Nit: Shouldn't reference dev-qemu here since the uart-service is platform agnostic.

uart-service/src/lib.rs
```diff
@@ -84,37 +83,47 @@ impl<R: RelayHandler> Service<R> {
...
+            filled += n;
+            match self
+                .medium
+                .frame_complete(buf.get(..filled).ok_or(Error::Serialize("buffer overrun"))?)
+                .map_err(Error::Mctp)?
+            {
+                Some(len) => break len,
+                None => continue,
+            }
+        };
```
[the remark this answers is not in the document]
dymk: Confirmed correct observation. The current v1.7/v1.8 traffic pattern is synchronous request/response (peer sends one frame per turn, then waits), so over-read can't occur in practice — but the API doesn't guarantee correctness if a streaming/batched consumer arrives later.

Documented the assumption + its implications in the function-level doc comment (commit `9d89778`). Captured the proper ring-buffer refactor as a follow-up — appropriate for a separate PR when a real consumer needs it.

mctp-rs/src/medium/serial.rs
```diff
@@ -293,6 +293,22 @@ impl MctpMedium for MctpSerialMedium {
...
     }
+
+    fn frame_complete(&self, buf: &[u8]) -> MctpPacketResult<Option<usize>, Self> {
+        // DSP0253 framing: [ 0x7E (start flag) ] [ stuffed body + FCS ] [ 0x7E (end flag) ]
+        // Skip leading flag if present, then find next 0x7E end-flag.
+        if buf.is_empty() {
+            return Ok(None);
+        }
+        let body_start = if buf[0] == END_FLAG { 1 } else { 0 };
+        if body_start >= buf.len() {
```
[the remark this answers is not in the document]
dymk: Fixed in commit `b166067`. Comment now accurately describes that `MctpSerialMedium::serialize` emits `[revision, byte_count, ...stuffed body, fcs_msb, fcs_lsb, 0x7E]` (END flag only, no leading START flag), with a separate note explaining that the `body_start` skip is forward-defensive for DSP0253 variants that DO emit a leading flag.

uart-service/src/task.rs
```diff
@@ -3,23 +3,30 @@ use embedded_io_async::Read as UartRead;
...
-            error!("uart-service request error: {:?}", e);
+        if uart_service.wait_for_request(&mut uart).await.is_err() {
+            error!("uart-service request error");
         } else {
             let host_msg = uart_service.wait_for_response().await;
-            if let Err(e) = uart_service.process_response(&mut uart, host_msg).await {
-                error!("uart-service response error: {:?}", e)
+            if uart_service.process_response(&mut uart, host_msg).await.is_err() {
+                error!("uart-service response error");
             }
```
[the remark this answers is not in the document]
dymk: Fixed in commit `0941c85`. Restored per-variant diagnostic signal by matching on `Error<M>` and emitting a fixed static string per variant (`Comms`, `Uart`, `Mctp`, `Serialize(s)`, `Buffer`). The `Serialize(s)` arm interpolates its `&'static str` payload. Also split into `log_request_error` / `log_response_error` so the request vs response site is clear in logs.

uart-service/src/lib.rs
```diff
@@ -84,37 +83,47 @@ impl<R: RelayHandler> Service<R> {
...
+        let mut buf = [0u8; BUF_SIZE];
+        let mut filled = 0usize;
+        let packet_len = loop {
+            let dst = buf.get_mut(filled..).ok_or(Error::Serialize("buffer overrun"))?;
+            if dst.is_empty() {
+                return Err(Error::Serialize("frame exceeds BUF_SIZE"));
+            }
+            let n = uart.read(dst).await.map_err(|_| Error::Uart)?;
+            if n == 0 {
+                return Err(Error::Comms);
```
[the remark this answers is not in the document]
dymk: Confirmed: `MctpSerialMedium` worst-case wire is `2 (header) + 251×2 (each body byte stuffed) + 5 (trailer) = 509 bytes`, well over the hardcoded 256-byte `BUF_SIZE`. v1.8 Battery traffic is well within (request ≈ 12 wire bytes, response ≈ 28 wire bytes), but the "generic over MctpMedium" claim is leaky at full MTU.

Documented the limitation in the `wait_for_request` doc comment (commit `9d89778`) with concrete worst-case math + two suggested designs (const-generic `BUF_SIZE` on `Service`, or a new `MctpMedium::max_wire_frame_size()` trait method). Holding off on the refactor for a follow-up PR — would prefer to ship the medium-aware sizing together with whatever real consumer first pushes the limit.

uart-service/src/lib.rs
```diff
@@ -84,37 +83,81 @@ impl<R: RelayHandler> Service<R> {
...
 
-    async fn wait_for_request<T: UartRead>(&self, uart: &mut T) -> Result<(), Error> {
+    /// Read one complete medium-framed packet from `uart` + dispatch it
+    /// through the relay handler.
+    ///
+    /// ## Known limitation: single-frame-per-read assumption
+    ///
+    /// `embedded_io_async::Read::read` can return more bytes than the
+    /// caller would prefer (it returns whatever is in the OS/HW RX
+    /// buffer). If the peer sends back-to-back frames in one
```
kurtjd: I'm not sure I understand this comment. This does not seem true based on the description: https://docs.rs/embedded-io-async/latest/embedded_io_async/trait.Read.html

As in, I don't see anything mentioning that `read` might return MORE bytes than the length of the provided buffer (only that it can return LESS). There is nothing in the description that states "`read` always returns the full hardware FIFO".

uart-service/src/lib.rs
```diff
@@ -84,37 +83,81 @@ impl<R: RelayHandler> Service<R> {
...
+    /// kernel-level read, the trailing bytes of the second frame would
+    /// be in this function's stack buffer but discarded — only the
+    /// first `packet_len` bytes are passed to `deserialize_packet`,
+    /// then the buffer is dropped on return.
+    ///
+    /// This is FINE for the v1.7/v1.8 traffic pattern (synchronous
+    /// request/response — the peer sends exactly one frame per turn and
```
kurtjd: Where do v.1.7 and v1.8 come from?

uart-service/src/lib.rs
```diff
@@ -84,37 +83,81 @@ impl<R: RelayHandler> Service<R> {
...
+    /// then waits) but would corrupt streaming or async/batched
+    /// scenarios. A persistent ring buffer or `read_exact`-after-header
+    /// would lift this assumption — out of scope for the initial
+    /// genericization; appropriate for a follow-up PR if/when a
+    /// streaming consumer arrives.
```
kurtjd: Not really sure I understand this either. There is a real concern that the hardware FIFO might overrun if the `read` implementation is not backed by a ring-buffered UART or throttled with hw flow control, but this is a HAL implementation detail and not something the uart-service is concerned with.

uart-service/src/lib.rs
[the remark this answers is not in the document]
dymk: Walking back my earlier reply. After kurtjd's pushback (see [his comment on lib.rs:101](https://github.com/OpenDevicePartnership/embedded-services/pull/852#discussion_r3269523346)), I verified `embedded_io_async::Read::read`'s contract — it reads `1..=buf.len()` bytes, never MORE than the buffer. The framing of "over-read past the end of the first complete frame" was based on a misreading of the trait contract.

The narrower latent risk that DOES exist (peer batches frames into a single OS buffer, `read` returns bytes spanning both frames, only frame 1 gets dispatched) requires both a peer that batches AND a buffer big enough to hold multiple frames — neither happens in current synchronous request/response patterns.

Deleted the doc-comment that was based on the wrong framing (commit `15f2611`). Not documenting the narrower risk in the source — it's captured in our internal seeds file for the day a streaming consumer appears.

uart-service/src/lib.rs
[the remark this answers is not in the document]
dymk: Walking back the "documented as limitation" part. The doc-comment I added was correctly flagged by kurtjd as mixing layers (HAL details in a generic library) and leaking internal milestone vocabulary. Deleted in commit `15f2611`.

The underlying observation (BUF_SIZE = 256 vs MctpSerialMedium MTU = 509 worst-case) is real, but not exploitable in the current SmbusEspi-heavy traffic shape, and the right fix (medium-aware buffer sizing via a `max_wire_frame_size()` trait method) is bigger than this PR. Captured for a follow-up; the buffer overflow itself fails cleanly with `frame exceeds BUF_SIZE` rather than corrupting wire bytes.

uart-service/src/lib.rs
dymk: Right — deleted the entire doc-comment block in commit `15f2611`. You were correct that `Read::read` doesn't over-read its buffer (the previous text was based on a misreading of the trait contract), the `v1.7`/`v1.8` tokens were internal milestone references that shouldn't leak into upstream docs, and HAL-level reasoning (PL011 RX FIFO) doesn't belong in a transport-agnostic library. The doc-comment was net-negative — better to have nothing there than the wrong story.

Thanks for the careful read.

uart-service/src/lib.rs
dymk: Right — deleted the entire doc-comment block in commit `15f2611`. You were correct that `Read::read` doesn't over-read its buffer (the previous text was based on a misreading of the trait contract), the `v1.7`/`v1.8` tokens were internal milestone references that shouldn't leak into upstream docs, and HAL-level reasoning (PL011 RX FIFO) doesn't belong in a transport-agnostic library. The doc-comment was net-negative — better to have nothing there than the wrong story.

Thanks for the careful read.

uart-service/src/lib.rs
dymk: Right — deleted the entire doc-comment block in commit `15f2611`. You were correct that `Read::read` doesn't over-read its buffer (the previous text was based on a misreading of the trait contract), the `v1.7`/`v1.8` tokens were internal milestone references that shouldn't leak into upstream docs, and HAL-level reasoning (PL011 RX FIFO) doesn't belong in a transport-agnostic library. The doc-comment was net-negative — better to have nothing there than the wrong story.

Thanks for the careful read.

uart-service/src/task.rs
dymk: Already addressed — that comment was in the old code path I rewrote in commit `1a38784`. The block that referenced dev-qemu is gone; replaced with per-variant `log_request_error` / `log_response_error` helpers (no platform references).

kurtjd:
> @kurtjd — validated. Three findings, all on your side:
> 
> **1. PR #852 makes zero API changes that `ec-test-lib` / `ec-test-cli` / `ec-test-tui` depend on.**
> 
> Source audit of `odp-platform-common`'s `ec/` workspace:
> 
> * They use only `battery_service_messages`, `time_alarm_service_messages`, `thermal_service_messages`, and `embedded_services::relay::{MessageSerializationError, SerializableMessage}`.
> * No usage of `uart_service::*`, `mctp_rs::MctpMedium`, `MctpMessageBuffer`, `frame_complete`, or `DefaultService` (the surfaces this PR changes).
> 
> ```shell
> $ grep -rn "uart_service\|MctpMedium\|MctpMessageBuffer\|frame_complete\|DefaultService" odp-platform-common/ec --include="*.rs"
> (no output)
> ```
> 
> **2. The PR for `odp-embedded-controller` will not affect `ec-test-tui` either.**
> 
> The followup PR is scoped to `platform/dev-qemu`, which is the QEMU dev platform — not a target `ec-test-tui` exercises. `ec-test-tui` drives real EC hardware (`dev-imxrt`, `dev-mcxa`, `dev-npcx`) over USB-serial; those platforms continue to speak SmbusEspi-shaped MCTP unchanged.
> 
> **3. Heads-up: `odp-platform-common` is already broken against current `v0.2.0` HEAD, independent of PR #852.**
> 
> Upstream PRs #780 (battery) and #776 (time-alarm) split `*-service-messages` crates into `*-interface` + `*-relay`. `odp-platform-common/ec/Cargo.toml` still references the old crate names; it builds today only because `Cargo.lock` pins to SHA `5a0ed0d`.
> 
> ```shell
> $ cd odp-platform-common/ec && cargo update
> error: no matching package named `battery-service-messages` found
> location searched: Git repository https://github.com/OpenDevicePartnership/embedded-services?branch=v0.2.0
> required by package `ec-test-cli`
> ```
> 
> So a simultaneous-PRs coordination is unnecessary for #852 / #odp-embedded-controller-followup specifically, but `odp-platform-common` does need its own update for the unrelated upstream rename — happy to file an issue or open a PR there if you want.

This doesn't address my concern. The ec-tets-tui/cli currently manually push bytes over serial in a way that should agree with what the uart-service is expecting. I just wanted to ensure that the `serial` source on the ec test app was validated against this change. If it's broken, then a PR should be opened there simultaneously to update it. It's not a matter of the Rust interface breaking that I'm referring to, jjust the format on the wire.

Also, the followup PR shouldn't just be scoped to dev-qemu. All the dev-platforms instantiate the uart-service, and the ec-test-tui is in fact expected to talk to all of them over the `serial` flag. In a nutshell, I want to ensure the test app can speak to all the existing dev platforms (including dev-qemu) over `--source serial`.

dymk:
@kurtjd re: breaking the TUI app - I'll put up a PR to make the changes there as well to fix the smbus/mctp workaround. Does it support being ran outside of a WinVOS image and talking to a virtual serial port? If so it'd be nice to have it as part of the E2E tests that run in CI.

kurtjd:
> @kurtjd re: breaking the TUI app - I'll put up a PR to make the changes there as well to fix the smbus/mctp workaround. Does it support being ran outside of a WinVOS image and talking to a virtual serial port? If so it'd be nice to have it as part of the E2E tests that run in CI.

Yes, the ec-test-app can be run from any host machine with the `--source serial` flag a provided port (e.g. `COM21` or `/dev/tty/USB0` or `/dev/pty/2`).

I recently integrated it into CI on the odp-embedded-controller side here: https://github.com/OpenDevicePartnership/odp-embedded-controller/pull/14

uart-service/src/task.rs
```diff
@@ -3,24 +3,53 @@ use embedded_io_async::Read as UartRead;
...
 }
+
+/// Emit a per-variant static log message for a `wait_for_request` error.
+///
+/// We can't use `{:?}` because `Error<M>` only implements `defmt::Format`
+/// when `M::Error` does too, and that bound can't be expressed as a
+/// feature-gated where-clause on stable Rust (rust-lang/rust#115590).
+/// Matching on the variant + emitting a fixed string preserves the
+/// per-variant signal without that bound.
+fn log_request_error<M: MctpMedium>(e: &Error<M>) {
```
kurtjd: Looks good and seems like everything is in place to ensure nothing breaks.

Minor nit: I think it's better to have a single:

`fn log_error<M: MctpMedium>(msg: &str, e: &Error<M>)`

Then just call it like:
`log_error("uart-service request:", &e)`
`log_error("uart-service response:", &e)`

kurtjd (dismissed):

kurtjd (approved):

tullom (approved):
Assuming everything works on the odp-e-c and ec-test-app side, lgtm

felipebalbi (approved):

PR #866, opened 2026-05-27
Add relayable service event plumbing

### LINKED ISSUE #733: embedded-services: Add support for notifications
We need to add a transport agnostic means of representing notification messages (that tasks can send asynchronously to relay services).

kurtjd:
As discussed, a grand unified notification handling system is not really feasible, so this PR does a few things.

First, it adds some plumbing though that should make it easier for all relay services to handle notifications to host:
- Introduces several new generic Receiver types: A MapReceiver which is self-explanatory and a MuxReceiver which allows an arbitrary number of heterogeneous Receiver types to be combined into a single Receiver. The relay handler macro for MCTP relay services uses these to construct a mapped and muxed receiver of all relayable service events which relay service tasks can make use of. Also added a FilterReceiver for reasons I'll discuss in next paragraph.
- Adds associated types to service relay traits for specifying the event that relayable service emits (added a ThermalEvent which just wraps Sensor and Fan events, and left the other services unit type for now until we know what they want to emit).

Second, updated the uart service to turn events into notifications. I had to make the `wait_for_request` method cancel-safe, so split it up and added an external struct for storing read state in case it is cancelled. This now allows us to select over requests from the host and events from the services.

The uart service now takes a "notifiable event" receiver. This is the part I'm not entirely happy with and can't think of a very clean approach. The problem is a relay service like the uart service doesn't know what services ahead of time will be declared relayable by the relay macro, and additionally it doesn't know which events the caller wants to consider notifiable.

And the way the MCTP uart service might choose to handle notifications over the wire is different from say HID I3C. So, I added an additional FilterReceiver type with the idea that the caller first needs to take the MUXed receiver produced by the macro, filter it for events they consider notifiable on their platform, then finally map them to a single u8 which the uart service will TX to the host (with the assumption that the caller ensures host agrees on what these u8s mean). Right now that part is TODO because we have to get the plumbing all the way through otherwise this can cause issues.

The reason I'm not too happy with this is because the caller has to handle creating a channel for each relayable service, ensuring the receiver passed to the `event_mux` function belongs to the same channel that the sender which was passed to the service belongs to, and then they have to set up the filter+map of events to something the relay service understands. But again, a clean solution to this is not coming to me that still allows flexibility for all kinds of relay serivces... You can see an example of how it's setup in this branch on odp-ec: https://github.com/kurtjd/odp-embedded-controller/blob/baa55f1bc6af1839f27e092d9602e27dbd8a1aa3/platform/dev-qemu/src/main.rs#L37

Resolves #733 

Assisted-by: GitHub Copilot:claude-opus-4.6

### DIFF
--- a/embedded-service/src/event.rs
+++ b/embedded-service/src/event.rs
@@ -291,7 +291,7 @@ impl<E> Receiver<E> for NeverReceiver<E> {
     }
 }
 
-/// Combines multiple receivers into one by racing them and returning
+/// Combines multiple receivers into one by racing them (with left-bias) and returning
 /// the first event that becomes available mapped to a common event type.
 pub struct MuxReceiver<E, L: Receiver<E>, R: Receiver<E>> {
     left: L,
--- a/embedded-service/src/event.rs
+++ b/embedded-service/src/event.rs
@@ -191,3 +191,157 @@ impl<I, O, S: Sender<O>, F: FnMut(I) -> O> Sender<I> for MapSender<I, O, S, F> {
         self.sender.send((self.map_fn)(event))
     }
 }
+
+/// Applies a function on events received from the wrapped receiver.
+pub struct MapReceiver<I, O, R: Receiver<I>, F: FnMut(I) -> O> {
+    receiver: R,
+    map_fn: F,
+    _phantom: PhantomData<(I, O)>,
+}
+
+impl<I, O, R: Receiver<I>, F: FnMut(I) -> O> MapReceiver<I, O, R, F> {
+    /// Create a new MapReceiver.
+    pub fn new(receiver: R, map_fn: F) -> Self {
+        Self {
+            receiver,
+            map_fn,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<I, O, R: Receiver<I>, F: FnMut(I) -> O> Receiver<O> for MapReceiver<I, O, R, F> {
+    fn try_next(&mut self) -> Option<O> {
+        self.receiver.try_next().map(&mut self.map_fn)
+    }
+
+    async fn wait_next(&mut self) -> O {
+        (self.map_fn)(self.receiver.wait_next().await)
+    }
+}
+
+/// Filters events from the wrapped receiver, only yielding events that pass the predicate.
+///
+/// Events that do not pass the filter are consumed and discarded.
+pub struct FilterReceiver<E, R: Receiver<E>, F: FnMut(&E) -> bool> {
+    receiver: R,
+    filter_fn: F,
+    _phantom: PhantomData<E>,
+}
+
+impl<E, R: Receiver<E>, F: FnMut(&E) -> bool> FilterReceiver<E, R, F> {
+    /// Create a new FilterReceiver.
+    pub fn new(receiver: R, filter_fn: F) -> Self {
+        Self {
+            receiver,
+            filter_fn,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<E, R: Receiver<E>, F: FnMut(&E) -> bool> Receiver<E> for FilterReceiver<E, R, F> {
+    fn try_next(&mut self) -> Option<E> {
+        loop {
+            match self.receiver.try_next() {
+                Some(e) if (self.filter_fn)(&e) => return Some(e),
+                Some(_) => continue,
+                None => return None,
+            }
+        }
+    }
+
+    async fn wait_next(&mut self) -> E {
+        loop {
+            let e = self.receiver.wait_next().await;
+            if (self.filter_fn)(&e) {
+                return e;
+            }
+        }
+    }
+}
+
+/// A receiver that never produces events.
+///
+/// This is mainly used to make it easier to construct a `MuxReceiver`
+/// via macro since we don't need to handle the special start case
+/// when chaining `with` calls.
+pub struct NeverReceiver<E>(PhantomData<E>);
+
+impl<E> NeverReceiver<E> {
+    /// Create a new NeverReceiver.
+    pub fn new() -> Self {
+        Self(PhantomData)
+    }
+}
+
+impl<E> Default for NeverReceiver<E> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<E> Receiver<E> for NeverReceiver<E> {
+    fn try_next(&mut self) -> Option<E> {
+        None
+    }
+
+    async fn wait_next(&mut self) -> E {
+        core::future::pending().await
+    }
+}
+
+/// Combines multiple receivers into one by racing them and returning
+/// the first event that becomes available mapped to a common event type.
+pub struct MuxReceiver<E, L: Receiver<E>, R: Receiver<E>> {
+    left: L,
+    right: R,
+    _phantom: PhantomData<E>,
+}
+
+impl<E> MuxReceiver<E, NeverReceiver<E>, NeverReceiver<E>> {
+    /// Create an empty MuxReceiver.
+    ///
+    /// Use `.with()` to add receivers.
+    pub fn new() -> Self {
+        Self {
+            left: NeverReceiver::new(),
+            right: NeverReceiver::new(),
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<E> Default for MuxReceiver<E, NeverReceiver<E>, NeverReceiver<E>> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<E, L: Receiver<E>, R1: Receiver<E>> MuxReceiver<E, L, R1> {
+    /// Add another receiver to multiplex with this one.
+    pub fn with<I, R2: Receiver<I>, F: FnMut(I) -> E>(
+        self,
+        receiver: R2,
+        map_fn: F,
+    ) -> MuxReceiver<E, Self, MapReceiver<I, E, R2, F>> {
+        MuxReceiver {
+            left: self,
+            right: MapReceiver::new(receiver, map_fn),
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<E, L: Receiver<E>, R: Receiver<E>> Receiver<E> for MuxReceiver<E, L, R> {
+    fn try_next(&mut self) -> Option<E> {
+        self.left.try_next().or_else(|| self.right.try_next())
+    }
+
+    async fn wait_next(&mut self) -> E {
+        match embassy_futures::select::select(self.left.wait_next(), self.right.wait_next()).await {
+            embassy_futures::select::Either::First(e) => e,
+            embassy_futures::select::Either::Second(e) => e,
+        }
+    }
+}
--- a/embedded-service/src/event.rs
+++ b/embedded-service/src/event.rs
@@ -32,6 +32,11 @@ pub trait Receiver<E> {
     /// Return none if there are no pending events
     fn try_next(&mut self) -> Option<E>;
     /// Receive an event
+    ///
+    /// # Cancel Safety
+    ///
+    /// The implementation MUST be cancel-safe as the `wait_next` method
+    /// is typically called inside a `select!` and thus may be cancelled.
     fn wait_next(&mut self) -> impl Future<Output = E>;
 }
 
--- a/embedded-service/src/relay/mod.rs
+++ b/embedded-service/src/relay/mod.rs
@@ -116,6 +116,9 @@ pub mod mctp {
 
         /// The result type that this service handler processes
         type ResultType: super::SerializableResult;
+
+        /// The event type that this service emits.
+        type EventType;
     }
 
     /// Trait for a service that can be relayed over an external bus (e.g. battery service, thermal service, time-alarm service)
@@ -448,6 +451,14 @@ pub mod mctp {
                     }
 
 
+                    /// A common event type wrapper for all relayable service events.
+                    #[derive(Debug)]
+                    pub enum ServiceEvent {
+                        $(
+                            $service_name(<$service_handler_type as $crate::relay::mctp::RelayServiceHandlerTypes>::EventType),
+                        )+
+                    }
+
                     pub struct $relay_type_name {
                         $(
                             [<$service_name:snake _handler>]: $service_handler_type,
@@ -465,6 +476,33 @@ pub mod mctp {
                                     [<$service_name:snake _handler>],
                                 )+
                             }
+                        }
+
+                        /// Build an event multiplexer from the provided relayable services.
+                        ///
+                        /// This is generic over the receiver type used for each service, so callers
+                        /// can decide at the call site which concrete `Receiver<EventType>` impl to
+                        /// supply for each relayed service (e.g. `NeverReceiver`, an embassy channel
+                        /// `Receiver`, a `DynamicReceiver`, or a custom impl).
+                        ///
+                        /// The caller will then need to further filter this event mux to whatever
+                        /// events they consider worth notifying the host about.
+                        ///
+                        /// Lastly, the caller will then need to map the events into a format the
+                        /// relay service understands (e.g. a single u8 for uart-service).
+                        pub fn event_mux<
+                            $(
+                                [<$service_name Rx>]: $crate::event::Receiver<
+                                    <$service_handler_type as $crate::relay::mctp::RelayServiceHandlerTypes>::EventType
+                                >,
+                            )+
+                        >(
+                            $(
+                                [<$service_name:snake _event_rx>]: [<$service_name Rx>],
+                            )+
+                        ) -> impl $crate::event::Receiver<ServiceEvent> {
+                            $crate::event::MuxReceiver::new()
+                                $(.with([<$service_name:snake _event_rx>], ServiceEvent::$service_name))+
                         }
                     }
 
@@ -494,6 +532,7 @@ pub mod mctp {
 
                 // Allows this generated relay type to be publicly re-exported
                 pub use [< _odp_impl_ $relay_type_name:snake >]::$relay_type_name;
+                pub use [< _odp_impl_ $relay_type_name:snake >]::ServiceEvent;
 
             } // end paste!
         }; // end macro arm
--- a/embedded-service/src/relay/mod.rs
+++ b/embedded-service/src/relay/mod.rs
@@ -172,6 +172,11 @@ pub mod mctp {
             &'a self,
             message: Self::RequestEnumType,
         ) -> impl core::future::Future<Output = Self::ResultEnumType> + 'a;
+
+        /// Returns a mutable reference to the multiplexed receiver of all relayable service events.
+        ///
+        /// Each service's events are mapped to the service's u8 ID.
+        fn receiver(&mut self) -> &mut impl crate::event::Receiver<u8>;
     }
 
     /// This macro generates a relay type over a collection of message types, which can be used by a relay service to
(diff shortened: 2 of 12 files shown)


RobertZ2011 (dismissed):

uart-service/src/task.rs
```diff
@@ -1,26 +1,59 @@
...
+            }
+            Either::First(Err(e)) => {
+                log_error("request", &e);
+                read_state.reset();
+            }
+            Either::Second(event) => {
+                warn!(
+                    "uart-service received notifiable event ({}) from relayable service",
+                    event
+                );
```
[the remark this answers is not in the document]
kurtjd: Notifiable events are intended to be relatively infrequent, and the caller is responsible for correctly filtering events before passing the event receiver to the uart-service.

embedded-service/src/relay/mod.rs
```diff
@@ -466,6 +477,33 @@ pub mod mctp {
...
+
+                        /// Build an event multiplexer from the provided relayable services.
+                        ///
+                        /// This is generic over the receiver type used for each service, so callers
+                        /// can decide at the call site which concrete `Receiver<EventType>` impl to
+                        /// supply for each relayed service (e.g. `NeverReceiver`, an embassy channel
+                        /// `Receiver`, a `DynamicReceiver`, or a custom impl).
+                        ///
+                        /// The caller will then need to further filter this event mux to whatever
+                        /// events they consider worth notifying the host about.
```
williampMSFT: I think all this filtering stuff belongs in the relay handlers?

williampMSFT:
>     }

I think the relay handlers need to take something that emits an event at construction time (or have a way to interrogate the service they're passed for a pubsubclient but I think that's incompatible with the DIY broadcast thing we're doing)

---
Refers to: time-alarm-service-relay/src/lib.rs:17 in 4651f80. [](commit_id = 4651f800531adeb493fc55bf057d7644c5c00aad, deletion_comment = False)

embedded-service/src/relay/mod.rs
kurtjd: Thought about that, but I was trying to maximize flexibility in that I'm not sure if we should assume that all users will agree on what events are considered "notifiable". Maybe some users have platforms where they want to notify the host in case of sensor error where others don't.

embedded-service/src/relay/mod.rs
```diff
@@ -494,6 +532,7 @@ pub mod mctp {
 
                 // Allows this generated relay type to be publicly re-exported
                 pub use [< _odp_impl_ $relay_type_name:snake >]::$relay_type_name;
+                pub use [< _odp_impl_ $relay_type_name:snake >]::ServiceEvent;
```
williampMSFT: >ServiceEvent [](http://example.com/codeflow?start=65&length=12)

I don't think we want to emit a hardcoded name into the surrounding namespace - we may want this to be an associated type on the RelayHandler trait to avoid this problem

kurtjd:
> > ```
> > }
> > ```
> 
> I think the relay handlers need to take something that emits an event at construction time (or have a way to interrogate the service they're passed for a pubsubclient but I think that's incompatible with the DIY broadcast thing we're doing)
> 
> Refers to: time-alarm-service-relay/src/lib.rs:17 in [4651f80](https://github.com/OpenDevicePartnership/embedded-services/commit/4651f800531adeb493fc55bf057d7644c5c00aad). [](commit_id = [4651f80](https://github.com/OpenDevicePartnership/embedded-services/commit/4651f800531adeb493fc55bf057d7644c5c00aad), deletion_comment = False)

Understood, but the challenge was coming up with a solution for this while still supporting the design we are heading toward: services only take a list of senders and don't own any backing channel. So I haven't been able to come up with a way to merry both these approaches together unfortuantely.

embedded-service/src/relay/mod.rs
kurtjd: Hm, though with the current approach: the caller has to filter and map the MUXed event receiver, the caller needs to know about the ServiceEvent type no?

williampMSFT:
sorry, to clarify, I'm asserting that the relay handler *should* own a backing channel. The services can continue to do the DIY-broadcast thing if they want, but a relay handler has exactly one client and the shape of the messages that it emits are different than the shape of the messages that the services emit. It can emit messages to its channel based on one of the DIY channels from the underlying service, and it'll need a handle to that at construction time

---
In reply to: [4566389330](https://github.com/OpenDevicePartnership/embedded-services/pull/866#issuecomment-4566389330) [](http://example.com/codeflow?ancestors=4566389330)

kurtjd:
> sorry, to clarify, I'm asserting that the relay handler _should_ own a backing channel. The services can continue to do the DIY-broadcast thing if they want, but a relay handler has exactly one client and the shape of the messages that it emits are different than the shape of the messages that the services emit. It can emit messages to its channel based on one of the DIY channels from the underlying service, and it'll need a handle to that at construction time
> 
> In reply to: [4566389330](https://github.com/OpenDevicePartnership/embedded-services/pull/866#issuecomment-4566389330)

I can play around with this but that means we now need a task for the relay service that is dedicated to listening to all the receivers for each service, then sending them onto a new channel, which the relay serivce then also needs to listen on (as opposed to the current approach I have where there is no middle man like that). Not exactly sure what that will look like but can experiment.

williampMSFT:
> > sorry, to clarify, I'm asserting that the relay handler _should_ own a backing channel. The services can continue to do the DIY-broadcast thing if they want, but a relay handler has exactly one client and the shape of the messages that it emits are different than the shape of the messages that the services emit. It can emit messages to its channel based on one of the DIY channels from the underlying service, and it'll need a handle to that at construction time
> > In reply to: [4566389330](https://github.com/OpenDevicePartnership/embedded-services/pull/866#issuecomment-4566389330)
> 
> I can play around with this but that means we now need a task for the relay service that is dedicated to listening to all the receivers for each service, then sending them onto a new channel, which the relay serivce then also needs to listen on (as opposed to the current approach I have where there is no middle man like that). Not exactly sure what that will look like but can experiment.

It might not need to be a literal embassy `Channel` - maybe just a 'next_notification(&self)` function on the relayhandler that blocks until a notification is ready, and that can be implemented by waiting on the DIY channel or something like that?  I don't think you need an extra task for this, just a select over "host wants me to do something" and "notification came in"

kurtjd:
> > > sorry, to clarify, I'm asserting that the relay handler _should_ own a backing channel. The services can continue to do the DIY-broadcast thing if they want, but a relay handler has exactly one client and the shape of the messages that it emits are different than the shape of the messages that the services emit. It can emit messages to its channel based on one of the DIY channels from the underlying service, and it'll need a handle to that at construction time
> > > In reply to: [4566389330](https://github.com/OpenDevicePartnership/embedded-services/pull/866#issuecomment-4566389330)
> > 
> > 
> > I can play around with this but that means we now need a task for the relay service that is dedicated to listening to all the receivers for each service, then sending them onto a new channel, which the relay serivce then also needs to listen on (as opposed to the current approach I have where there is no middle man like that). Not exactly sure what that will look like but can experiment.
> 
> It might not need to be a literal embassy `Channel` - maybe just a 'next_notification(&self)` function on the relayhandler that blocks until a notification is ready, and that can be implemented by waiting on the DIY channel or something like that? I don't think you need an extra task for this, just a select over "host wants me to do something" and "notification came in"

I think I get what you're saying... I originally wanted to not return a separate `event_mux` function that the user had to call and just tie it directly to the relay service handler like you described, but it gets tricky because the `Receiver` trait takes `&mut` which makes it hard to select over two futures which both borrow `self`. Would have to come up with some interior mutability approach I guess using mutexes. I'll try to think of a good way around this some more but nothing immediately came to mind when I thought about it originally.

uart-service/src/lib.rs
```diff
@@ -107,37 +125,52 @@ impl<R: RelayHandler, M: MctpMedium + Copy> Service<R, M> {
...
-        let mut buf = [0u8; BUF_SIZE];
-        let mut filled = 0usize;
-        let packet_len = loop {
-            let dst = buf.get_mut(filled..).ok_or(Error::Serialize("buffer overrun"))?;
+    // Read bytes from UART until a complete request is assembled.
+    //
+    // # Cancel Safety
+    //
+    // This method is cancel-safe because partial read progress is stored in `state`
+    // and will be resumed on the next call.
```
felipebalbi: I'm not 100% aligned with this claim because it's only half true. Let's say the user chose a DMA-backed UART driver. The embedded-io-async trait makes no cancel-safety guarantee for read. Many UART drivers will have already moved bytes out of the FIFO/DMA buffer into the destination slice, and dropping the future loses those bytes. `state.filled` won't be incremented because the await never returned.

Maybe you should document the dependency explicitly, that is "cancel-safe iff T::read is cancel-safe"?

embedded-service/src/event.rs
```diff
@@ -143,3 +143,157 @@ impl<I, O, S: Sender<O>, F: FnMut(I) -> O> Sender<I> for MapSender<I, O, S, F> {
...
+    fn try_next(&mut self) -> Option<E> {
+        self.left.try_next().or_else(|| self.right.try_next())
+    }
+
+    async fn wait_next(&mut self) -> E {
+        match embassy_futures::select::select(self.left.wait_next(), self.right.wait_next()).await {
+            embassy_futures::select::Either::First(e) => e,
+            embassy_futures::select::Either::Second(e) => e,
+        }
+    }
```
felipebalbi, line 351: Are we sure `L` and `R` are *always* cancel safe? I'm asking because `Receiver` does not impose a cancel-safety contract:

Image

embedded-service/src/relay/mod.rs
```diff
@@ -116,6 +116,9 @@ pub mod mctp {
 
         /// The result type that this service handler processes
         type ResultType: super::SerializableResult;
+
+        /// The event type that this service emits.
+        type EventType;
```
felipebalbi, line 121: `EventType` needs a `Debug` trait bound here, no...

embedded-service/src/relay/mod.rs
```diff
@@ -448,6 +451,14 @@ pub mod mctp {
                     }
 
 
+                    /// A common event type wrapper for all relayable service events.
+                    #[derive(Debug)]
```
felipebalbi: ... otherwise you may be unable to derive `Debug`. Additionally, should you add:

```rust
```

for consistency?

embedded-service/src/event.rs
```diff
@@ -143,3 +143,157 @@ impl<I, O, S: Sender<O>, F: FnMut(I) -> O> Sender<I> for MapSender<I, O, S, F> {
...
+            right: MapReceiver::new(receiver, map_fn),
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<E, L: Receiver<E>, R: Receiver<E>> Receiver<E> for MuxReceiver<E, L, R> {
+    fn try_next(&mut self) -> Option<E> {
+        self.left.try_next().or_else(|| self.right.try_next())
+    }
```
felipebalbi, line 344: This will always drain `left` first before looking at `right`. Could we starve `right` if `left` is always hot? There's a silent discrepancy between `try_next` and `next` below. Because `next` uses `select`, both futures will be polled, whereas `try_next` may never poll `right` at all.

uart-service/src/lib.rs
kurtjd: Yeah good catch, re-read the embedded-io-async doc and saw the specific mention about DMA.

I suppose the best approach is just to document above the `uart_service` function that `T: UartRead` must be cancel-safe, though would like to figure out a way to not make it possible to provide a embedded-io-async impl that is not cancel safe...

embedded-service/src/event.rs
kurtjd: Good point. I'll add documentation to the Receiver trait that methods must be cancel-safe and that should not be problematic because all `embassy-sync` futures are cancel-safe (see: https://github.com/embassy-rs/embassy/issues/5484#issuecomment-3921041927).

embedded-service/src/relay/mod.rs
kurtjd: I'll revisit this, did want to add defmt here originally but ran into some issues. Thanks.

embedded-service/src/event.rs
kurtjd: I might be wrong, but I think `wait_next` has the same left-biased problem as well. Looking at the source for `select` it will also return immediately if left is ready without polling right, so it seems right can get starved if left is hot in `wait_next` as well.

I did document that the `MuxReceiver` is left-biased, but I didn't bother to attempt to make this all more "fair" as at the moment I don't see any downsides to the bias (though if I'm later proven wrong suppose will have to revisit then).

embedded-service/src/relay/mod.rs
kurtjd: Good point will add.

uart-service/src/lib.rs
felipebalbi: I suppose to guarantee that, you would have to add an `unsafe trait ThisIsCancelSafePinkyPromise` and expect users to implement it. Not sure you gain much from that :(

kurtjd:
@williampMSFT So I was able to improve things a bit:

- We discussed the idea of creating a Receiver "on-demand" but it still didn't seem feasible in this case: since we don't own the backing channel, we have to own a Receiver. So what I did was create a field to hold the `MuxReceiver` created during construction then add a trait method `receiver()` to just get a mutable reference to it (we also discussed the idea of a "give me a receiver trait" but didn't seem necessary here when we can just add a method to the existing Relay trait.
- To get around borrow issues, separated the uart service into a `ServiceInner` separate from the relay_handler field so there's no conflict during the `select` call. However, this requires making `uart_service` mutably borrow or own the `Service` handle (since the Receiver belonging to the relay_handler which belongs to `Service` requires `&mut self`). Decided to make it own it since the `Service` handle doesn't really provide any sort of interface that some other task/function would want to make use of.
- So now we don't need to manually pass in a receiver into the uart_service, it just gets it from the relay_handler.
- Also made the macro handle the mapping from service event to u8 as discussed. If user wants to filter service events, they should create a FilterReceiver for that service before passing the Receiver to the relay handler constructor. The macro will always map the service to the same service_id that is passed into the macro (which should be fine for MCTP, we just make the rule that notification id = mctp service id). The idea is still that the uart-service would TX this ID to the host so the host knows what service needs attention.
- Example use seen here: https://github.com/kurtjd/odp-embedded-controller/blob/668a0da3005b56df88ba22db613e7d7f602034a1/platform/platform-common/src/mock/mod.rs#L35 (now the relay macro constructor takes each service's receiver in addition to its Service handler)

uart-service/src/lib.rs
kurtjd: Added docs describing the uart must have a cancel-safe `Read` implementation. Yeah unfortunately no real way to force this via the type system.

embedded-service/src/relay/mod.rs
kurtjd: Ended up not needing this since https://github.com/OpenDevicePartnership/embedded-services/pull/866#issuecomment-4595245063 removes the `ServiceEvent` type.

embedded-service/src/relay/mod.rs
kurtjd: Ended up not needing this since https://github.com/OpenDevicePartnership/embedded-services/pull/866#issuecomment-4595245063 removes the `ServiceEvent` type.

examples/rt685s-evk/src/bin/time_alarm.rs
```diff
@@ -51,7 +51,10 @@ async fn main(spawner: embassy_executor::Spawner) {
         TimeAlarm, 0x0B, crate::TimeAlarmServiceRelayHandlerType;
     );
 
-    let _relay_handler = EspiRelayHandler::new(TimeAlarmServiceRelayHandlerType::new(time_service));
+    let _relay_handler = EspiRelayHandler::new(
```
williampMSFT, line 54: re: this:
> So now we don't need to manually pass in a receiver into the uart_service, it just gets it from the relay_handler.

I think what I'm pushing for is to have the shape of this be something like this instead:
```rust
    let _relay_handler = EspiRelayHandler::new(
        TimeAlarmServiceRelayHandlerType::new(time_service, embedded_services::event::NeverReceiver::new()),
    );
```

and then the espiRelayHandler can interrogate the TimeAlarmServiceRelayHandlerType for a receiver to its notification channel, which will likely be of a different type than the receiver that's passed into the TimeAlarmServiceRelayHandlerType's constructor because the shape of notifications is going to be different across different transport protocols (e.g. MCTP might only support a `()`-typed message, whereas HID has to support messages with payloads).

This approach has a few advantages:
- It's not possible to mix up the notifications for different services (i.e. you can't accidentally wire up the battery service's notification handler to the thermal service) (at least at this layer - the DIY pubsub thing between service <-> relay handler still has this problem, which is part of why I don't like the DIY pubsub thing, but I think fixing that is out of scope for this PR)
- It's easier to read / reason about what shape a relay handler has because all that information is present in the trait definition, rather than being strewn about in the implementations of macros that consume instances of the trait - sort of the same difference between traditional C++ templates and C++20 concepts

Based on the work I've done on HID stuff I'm pretty convinced it's possible to do this and satisfy the borrow checker, it just requires a few weird-looking (to my eye, anyway) constructs around when you listen to the channel - happy to chat more if you're running into problems there :)

embedded-service/src/event.rs
```diff
@@ -191,3 +196,157 @@ impl<I, O, S: Sender<O>, F: FnMut(I) -> O> Sender<I> for MapSender<I, O, S, F> {
...
+pub struct MuxReceiver<E, L: Receiver<E>, R: Receiver<E>> {
+    left: L,
+    right: R,
+    _phantom: PhantomData<E>,
+}
+
+impl<E> MuxReceiver<E, NeverReceiver<E>, NeverReceiver<E>> {
+    /// Create an empty MuxReceiver.
+    ///
+    /// Use `.with()` to add receivers.
```
jerrysxie, line 310: What does `.with() to add receivers` mean?

uart-service/src/task.rs
```diff
@@ -1,26 +1,64 @@
...
+
+                // TODO: Here we would do something like:
+                //
+                // if let Err(_e) = uart.write_all(&[0x42, event]).await {
```
jerrysxie, line 51: So in term of ec->host notification, the desire is to keep the notification simple without data payload and have the host to come back to ask what the notification is about?

jerrysxie (commented):
Looks reasonable to me, couple of quick questions.

embedded-service/src/event.rs
kurtjd: So when you call `MuxReceiver::new()` it basically returns an empty receiver. To "multiplex" additional receivers you need to call `MuxReceiver::with(...)` which consumes `self` and spits out a new `MuxReceiver` with the additional receiver multiplexed into it. You can keep chaining `with` calls indefinitely to keep adding receivers.

Designed it this way because it was easier than trying to just take a list of N receivers statically upfront or just a slice of indeterminate length. This also feels a bit more ergonomic to select over because an array or slice of Receivers means I need to first map the array of receivers to an array of futures, then perform the select, etc.

uart-service/src/task.rs
kurtjd: Yep this is something I was discussing with Billy a lot, and part of the problem why it was difficult to create a "one size fits all" notification system.

So ultimately it comes down to the specific relay service and how it decides it wants to handle notifications (either as a doorbell or asynchronously sending a payload to the host).

The flexibility is there for future services, but for our MCTP relay services I decided on doorbell style only notifications. This is what we originally had awhile back in the eSPI service (basically just using VWIRE and the ALERT line  to doorbell the host that the specified service needs attention, and then the host makes a request to that service like you mentioned).

So TL;DR: Yes our MCTP services are doorbell only, but future relay services can choose payload notifications if they wish.

uart-service/src/task.rs
```diff
@@ -1,26 +1,64 @@
...
+
     loop {
-        if let Err(e) = uart_service.wait_for_request(&mut uart).await {
-            log_error("request", &e);
-        } else {
-            let host_msg = uart_service.wait_for_response().await;
-            if let Err(e) = uart_service.process_response(&mut uart, host_msg).await {
-                log_error("response", &e);
+        match select(
+            service.inner.wait_for_request(&mut uart, &mut read_state),
```
RobertZ2011, line 23: I would split the `Service` struct vertically, i.e. move the event receiving code into its own struct so that the end result here would look something like:
```rust
let event = event_receiver.wait_next(&mut uart, &mut read_state).await;

service.process_event(event).await;
```

This is more flexible because the core service logic/state doesn't have to be locked while you're waiting for an event. It also makes the processing logic slightly simpler because `inner.wait_for_response()` would now be handled by `event_receiver` so you wouldn't have the mixed waiting/processing logic. The type-C code already does something very similar with a `Lookback` receiver: https://github.com/OpenDevicePartnership/embedded-services/blob/576034c1f6ae347064e3124192cf8038b56bef16/type-c-service/src/controller/event_receiver.rs#L91

But I also think you could remove `host_tx_queue` entirely. It seems like the host messages get immediately sent after `process_request` so there's not really a need to queue them up.

RobertZ2011 (changes requested):

uart-service/src/task.rs
kurtjd: Good catch regarding `host_tx_queue`, yeah I can definitely remove that and just have `process_request` return the response directly.

Though I'm confused a bit by the rest of your comment. By removing host_tx_queue we don't need `wait_for_response`, but `wait_for_response` is unrelated to events, it's only related to host requests, so it wouldn't be handled by event_receiver. Likewise `process_response` is related to host requests, NOT service events.

Right now what this is doing is waiting for an event from a service OR a request from the host so I am unsure what you mean by the core logic being locked while waiting for an event.

Could you clarify a bit what you mean?

examples/rt685s-evk/src/bin/time_alarm.rs
kurtjd: I'll try to spend some more time thinking about this, but I'm worried there is some fundamental misunderstanding between us and I'm not sure the best way to resolve it. I might have to just wait until the hid services are a bit further along so I have something tangible to help me understand your concerns better.

kurtjd:
Closing this for now since there might still be some fundamental issues with the design. Will need to go back to the drawing board and think of some other approach.

PR #875, opened 2026-05-29
feat: canonical ODP wire format in mctp-rs + odp-client crate

dymk:
Consolidates three duplicate copies of the ODP MCTP wire format and relay scaffolding into two new homes inside this repo.

## Changes

1. **`mctp-rs` gains an `odp` feature** exposing `mctp::message_type::odp` with the canonical wire format:
   - `ODP_MESSAGE_TYPE = 0x7D`, `OdpService` enum, `OdpHeader` BE bitfield, `OdpMessage<'a>`
   - Implements `MctpMessageHeaderTrait` and `MctpMessageTrait<'buf>`
   - Wire-format snapshot tests pinned to the pre-migration byte layout (Battery / Thermal / TimeAlarm)

2. **New `odp-client` crate (`no_std`)** wraps the wire format so consumers depend only on `odp-client`, never on `mctp-rs` directly:
   - `OdpTransport` — pure byte-message trait (`send_message`/`recv_message`); no `MctpMedium` leakage
   - `MctpSerialTransport<U: embedded_io::Read + Write>` — concrete impl that encapsulates MCTP serial framing internally
   - `LoopbackTransport` — in-memory test double backed by `heapless::Vec`
   - `OdpClient<T: OdpTransport>` + `Relay` trait — sync client with closure-based `invoke` that hands the caller a borrowed response slice
   - `OdpError` — `Copy`/`Eq` error type with structured `UnexpectedMessageId { expected, got }` variant
   - `impl_odp_relay_handler!` macro — async server-side relay handler (moved verbatim from `embedded-service::relay`, renamed from `impl_odp_mctp_relay_handler!`)
   - `SerializableMessage` / `SerializableResult` traits — moved from `embedded-service::relay`

3. **Deletes `embedded-service/src/relay/` entirely** and drops `bitfield`, `paste`, and `mctp-rs` from `embedded-service/Cargo.toml`. All in-tree consumers (espi-service, uart-service, thermal-service-relay, battery-service-relay, time-alarm-service-relay, debug-service, debug-service-messages, examples/rt685s-evk) are migrated in this PR.

## Motivation

Before this change the ODP MCTP wire format (`ODP_MESSAGE_TYPE = 0x7D`, the `OdpService` enum, the `OdpHeader` big-endian bitfield, and the per-service request/response encoding) was duplicated across **three** copies in the wider ODP ecosystem:

| Repo | Location | LOC |
| --- | --- | --- |
| `embedded-services` | `embedded-service/src/relay/mctp.rs` | ~230 |
| `odp-secure-services` | `ec-service-lib/src/ec_relay.rs` | ~529 |
| `odp-platform-common` | `ec-test-lib/src/serial.rs` | ~150 |

Three copies of essentially the same on-the-wire bytes meant every protocol tweak required coordinated edits across three repos — and the copies had already drifted (decoded structure varied, endianness layered behind a bitfield trait, error types had different shapes). This PR makes `mctp-rs::odp` the single source of truth and provides `odp-client` as the sole API consumers depend on. Downstream consumers (`odp-secure-services`, `odp-platform-common`, `odp-platform-qemu-sbsa`, `odp-embedded-controller`) have follow-up PRs (already open on personal forks, full E2E green) that delete their duplicate copies once this lands upstream.

This is a **clean break**: no shims, no `#[deprecated]`, no backwards-compat layer. The renamed macro path (`impl_odp_mctp_relay_handler!` → `impl_odp_relay_handler!`) and the deleted `embedded-service::relay` module are both intentional consolidation moves.

## Verification

- `cargo build --workspace` ✅
- `cargo test --workspace` ✅ — all suites green, includes 15 new `odp-client` integration tests + 13 `mctp-rs::odp` wire-format snapshot tests pinned to the pre-migration byte layout
- `cargo clippy --workspace --all-targets -- -D warnings` ✅
- `cargo fmt --all -- --check` ✅
- `cd examples/rt685s-evk && cargo build --bin time_alarm` ✅ — exercises the relocated `impl_odp_relay_handler!` macro end-to-end

**End-to-end validation against downstream consumers:** all 4 downstream PRs (open on the contributor's personal forks against this branch via `[patch]`) have been verified live on QEMU SBSA + dev-qemu:
- Legacy FFA-only path (`make e2e-test`): 27/27 pass
- Full MCTP path (`make e2e-test-mctp`): UEFI ↔ FFA ↔ SP ↔ MCTP/serial ↔ EC round-trip with marker `EC_MCTP_OK service_id=8 msg_id=GetBst battery_status=...` matched
- Standalone EC integration (`mod/ec/scripts/integration-test.sh`): thermal/battery/RTC commands all decoded correctly
- Host ec-test-cli ↔ live QEMU SP ↔ EC: 4/4 commands round-trip clean

## Downstream PRs

These will be opened against their respective upstream repos after this PR lands and the `[patch]` redirections can be replaced with normal version pins:

- `odp-secure-services`: adopt `odp-client`; delete the 529-LOC `ec_relay.rs`
- `odp-embedded-controller`: 3-LOC swap of `embedded_services::relay::mctp::impl_odp_mctp_relay_handler!` → `odp_client::impl_odp_relay_handler!`
- `odp-platform-common`: rewrite `ec-test-lib/src/serial.rs` to use `OdpClient<SerialTransport<…>>`; delete the ~150-LOC duplicate framing shim
- `odp-platform-qemu-sbsa`: bump submodule pins + 5-line SP relay constructor swap

Net impact across the ecosystem: **~700 LOC of duplicate wire-format implementations deleted**, replaced by a single canonical source under `mctp-rs::odp` + a typed client crate at `odp-client`.


examples/rt685s-evk/src/bin/time_alarm.rs
```diff
@@ -45,8 +45,8 @@ async fn main(spawner: embassy_executor::Spawner) {
     )
     .expect("Failed to spawn time alarm service");
 
-    use embedded_services::relay::mctp::impl_odp_mctp_relay_handler;
-    impl_odp_mctp_relay_handler!(
+    use odp_client::impl_odp_relay_handler;
```
williampMSFT, line 48: >impl_odp_relay_handler [](http://example.com/codeflow?start=20&length=22)

We're going to need a similar but incompatible thing for HID - I don't think we want to take "MCTP" out of the name for this macro

mctp-rs/Cargo.toml
```diff
@@ -10,8 +10,9 @@ ignored = ["embedded-batteries", "espi-device", "uuid", "crc"]
 
 [features]
 default = []
-espi = ["dep:espi-device"]
 defmt = ["dep:defmt", "embedded-batteries/defmt"]
+espi = ["dep:espi-device"]
+odp = []
```
williampMSFT, line 15: This seems odd to me - why is this in mctp-rs? It seems like we wouldn't want mctp-rs to have explicit knowledge of its clients, right?

mctp-rs/src/message_type/odp.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub fn from_u8(raw: u8) -> Option<Self> {
+        match raw {
+            0x08 => Some(Self::Battery),
+            0x09 => Some(Self::Thermal),
+            0x0B => Some(Self::TimeAlarm),
+            _ => None,
+        }
+    }
+}
+
```
williampMSFT, line 30: We explicitly do not want this knowledge in this repository - this is something that the application has to provide (edit: because OEMs may want to add their own services to this list, remove services from this list, or put ODP services on different addresses)

mctp-rs/src/message_type/odp.rs
```diff
@@ -0,0 +1,332 @@
...
+        pub _reserved_b24:     u8 => [24],
+        pub is_request:        u8 => [25],
+        pub _reserved_b26_31:  u8 => [26:31],
+    }
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct OdpHeader {
+    pub is_request: bool,
+    pub service: OdpService,
```
williampMSFT, line 46: The shape of this field is not knowable without input from the OEM - this has to be codegenned
edit: it looks like this is still being codegenned too? it's on line 174 of server.rs - probably delete this version?

mctp-rs/Cargo.toml
dymk: It seemed like too much separation to me for the size of the code. I took a middle ground route, and put it behind a feature flag so that `mctp-rs` can be compiled & tested without any ODP knowledge If you think that it should be a fully separate crate, I can split it out though.

odp-client/src/client.rs
```diff
@@ -0,0 +1,78 @@
...
+
+impl<T: OdpTransport> Relay for OdpClient<T> {
+    fn invoke<'a>(&'a mut self, header: OdpHeader, body: &[u8]) -> Result<OdpResponse<'a>, OdpError> {
+        let need = 4 + body.len();
+        if need > self.buf.len() {
+            return Err(OdpError::BufferTooSmall);
+        }
+
+        // Compose request: 4-byte BE header + body.
+        self.buf[..4].copy_from_slice(&header.to_be_bytes());
```
williampMSFT, line 53: >[..4] [](http://example.com/codeflow?start=16&length=5)

I think there's a no-panic lint that's going to reject this; if it isn't, that may mean your lints aren't configured correctly for the new crates

williampMSFT:
> 

I was going to need to split the relay stuff into MCTP and HID sub-crates anyway - if you're already moving stuff, consider putting this stuff under an MCTP sub-crate so we don't have to move it again when the HID stuff is ready

---
Refers to: odp-client/src/lib.rs:34 in b00c3a7. [](commit_id = b00c3a7a3f1710fe725a0bbe29c1b6bb936817f1, deletion_comment = False)

dymk:
Parking this. The two-QEMU FFA serial E2E this was supporting only needs `MctpSerialMedium`, which has since landed in upstream `embedded-services` main — so the canonical-wire / `odp-client` refactor (which deletes `embedded-service/src/relay/` and is a sizable breaking change) isn't required for that goal.

Closing rather than merging, but **keeping the `dymk/odp-canonical-wire-and-odp-client` branch** so the work isn't lost if we revisit the shared-ODP-code factoring later (tracked as `sp-ec-relay-e2e` issue 10). Superseded by a fresh-eyes minimal approach.

PR #876, opened 2026-06-01
embedded-service: reliability, observability, and soundness hardening

felipebalbi:
Four commits hardening the foundational embedded-service crate against silent failure modes surfaced by an architect / reliability / reviewer audit. All changes are additive or compile-time-only; no public function signatures change.

**Commits**

- 632bfc1 — reliability and correctness fixes
- cb5b412 — observability counters and unsafe-block lint
- 8f6b360 — require T: Send on CriticalSectionCell Sync impl
- a3ddfbc — require T: Send on ThreadModeCell Sync impl

**Highlights**

Silent failures eliminated. comms::Endpoint::process no longer swallows MailboxDelegateError, and now also warn-logs + counts the case where a routed message reaches an endpoint whose delegator slot is still None. comms::register_endpoint reorders push-then-init so duplicate registration cannot silently overwrite. hid::Device::receive uniformly gates on message.id == self.id and uses a Channel<_, _, 2> (was a single-slot Signal). activity::Publisher::publish skips-with-trace instead of panicking on a foreign NodeContainer.

_Cancel safety_. ipc::deferred::Channel::execute is now cancel-safe via a CommandDropGuard that resets the command slot on caller-drop; previously a cancelled command would be silently overwritten by the next caller.

_Correctness_. hid::command::encode_data off-by-2 (latent) fixed. Buffer::SharedRef accepts empty trailing slices like len..len. intrusive_list::push is now atomic inside a single critical section.

_Observability_. New embedded_services::metrics module with 8 saturating AtomicUsize counters wired from every previously-silent failure path (comms.{delegator_errors,routed_unknown,delivered_to_uninit}, ipc_deferred.{dropped_commands,id_mismatches}, hid.request_overflows, broadcaster.lag, buffer.poisons).

_Soundness_. Both cell types now require T: Send for Sync (matching std::sync::Mutex). The cascade was resolved through IntrusiveNode, NodeContainer: Any + Send + Sync, and documented manual unsafe impl Send + Sync blocks on broadcaster::Receiver, comms::Endpoint, and activity::Subscriber justified by the single Cortex-M / single Embassy executor model.

_Guard-rails_. embedded-service enforces #![deny(clippy::undocumented_unsafe_blocks)]. Workspace-wide const _ Send guards added at every SyncCell<T> instantiation in the workspace.

**Test coverage**

- 34 → 61 unit tests, each substantive fix driven test-first (TDD red→green observed).
- 1 → 3 passing doctests, 0 → 5 compile_fail doctests at the cascade boundaries.
- Counter regression tests with snapshot-before/snapshot-after assertions.
- Cross-thread spawn tests on the new manual Send + Sync impls.

**Verification**

All gates green: workspace build; cargo test -p embedded-services (61 + 3 + 5); workspace tests (excluding pre-existing Windows-env failures in fw-update-interface{,-mocks}); cargo clippy --all-targets -- -D warnings; thumbv8m + --features defmt + --features log builds.

**Deferred (documented inline or in commit messages)**

Timeouts on ipc::deferred::execute / relay::process_request / hid::wait_request; typing comms::send/route to a real error type; typed Registry<T> / CommsRouter<'hw>; extraction of hid / relay / keyboard / activity from the foundation crate; loom-based concurrency proofs.

### DIFF
--- a/.opencode/agent/architect.md
+++ b/.opencode/agent/architect.md
@@ -1,0 +1,86 @@
+---
+description: Use when the task needs system design, API or module-boundary decisions, state-machine or invariant design, failure-domain analysis, or long-term architectural strategy. Produces specifications, not implementations. Trigger for "design", "architecture", "spec", "module boundary", "API shape", "state machine", "invariant", "should we", "how should this look".
+mode: subagent
+permission:
+  edit: deny
+  bash: ask
+  webfetch: allow
+  task: deny
+---
+
+# Architect
+
+You are the **Architect**: systems designer and long-term technical
+strategist for this project. Your primary goal is **robust, coherent,
+maintainable system designs** — not code.
+
+## Stance
+
+- Thoughtful, conservative, structured, big-picture oriented.
+- Resistant to unnecessary complexity. Calm under ambiguity.
+- Read `.github/copilot-instructions.md` and `docs/api-guidelines.md`
+  before proposing structural change. If your proposal contradicts
+  either, either say so explicitly and recommend updating the doc, or
+  revise the proposal.
+
+## What you do
+
+- API and interface design — especially the boundary between a
+  service crate and its `-interface` crate. Runtime APIs that callers
+  depend on should live in the `-interface` crate (mockable,
+  customizable); internal implementation details stay in the service
+  crate.
+- Module boundaries and layering across the EC service stack: HAL
+  (embedded-hal traits) → peripheral drivers → EC subsystem
+  abstractions → services → top-level composition.
+- Service-trait design: shape of `Service<'hw>::Resources`,
+  `InitParams`, `ErrorType`, and the `Runner` event loop. Lifetime
+  parameterisation (`'hw`) rather than `'static` references.
+- State machines, protocols, invariants — including IPC topology
+  (channels, signals, deferred request/response, broadcaster,
+  relay) and which mechanism fits each information shape.
+- Failure-domain analysis: what breaks what, what is recoverable,
+  what is not, what crosses an await point and what doesn't.
+- Trade-off articulation: name the alternatives you rejected and why.
+
+## How you work
+
+- Think before acting. Use your tools to read, search, and understand
+  before proposing.
+- Produce **specifications**: short documents, interface sketches,
+  pseudocode, state diagrams (as text), invariant lists, decision
+  records. Not patch series.
+- Prefer clean abstractions, but only when they earn their keep.
+  "Two call sites" is not enough motivation to extract.
+- Question architectural drift when you see it. Name it — including
+  drift away from the patterns documented in `docs/api-guidelines.md`
+  (no `'static` refs, no internal `OnceLock` singletons, trait-based
+  public APIs in `-interface` crates).
+- Avoid premature optimization. Correctness and clarity first.
+
+## What you do NOT do
+
+- You do **not** write large code patches. If small illustrative
+  snippets help, they are pseudocode in your spec, not files on disk.
+- You do **not** micromanage implementation choices that fall inside
+  a module's private surface.
+- You do **not** bikeshed names past one round.
+- You do **not** reach for a new abstraction every time a problem
+  appears.
+
+## Output format
+
+Every response you produce should be structured roughly like:
+
+1. **Context** — what you understood the problem to be.
+2. **Proposal** — the design, in spec form.
+3. **Invariants & failure modes** — what must always hold, what can
+   fail and how it is handled.
+4. **Alternatives considered** — at least one, with the reason it was
+   rejected.
+5. **Open questions** — anything the caller must decide before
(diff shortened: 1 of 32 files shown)


embedded-service/src/ipc/deferred.rs
```diff
@@ -2,9 +2,34 @@
...
 
-use crate::debug;
+use crate::warn;
 use embassy_sync::{blocking_mutex::raw::RawMutex, mutex::Mutex, signal::Signal};
 
+/// Drop guard for [`Channel::execute`].
+///
+/// On drop (i.e., the caller's future was cancelled before the responder
+/// picked up the command), resets the command signal. This prevents the
+/// scenario where a cancelled caller's command is silently overwritten by
```
RobertZ2011, line 12: This comment seems to be saying the opposite of what the code is doing. If the guard is clearing the command signal on drop, then that will prevent a cancelled call from executing if the responder hasn't grabbed it yet. Previously, the command would persist until the responder grabbed it, or it was overwritten, allowing execution to happen even if the call was cancelled.

embedded-service/src/ipc/deferred.rs
```diff
@@ -45,19 +70,49 @@ impl<M: RawMutex, C, R> Channel<M, C, R> {
...
     /// Send a command and return the response
     /// This locks to ensure that commands are executed atomically
+    ///
+    /// # Cancel safety
+    ///
+    /// This future is cancel-safe with respect to the channel state. If this
+    /// future is dropped (e.g. via `select`, `tokio::timeout`, or task abort)
+    /// after `command.signal(...)` but before the responder picks the command
+    /// up, the dropped command would otherwise be silently overwritten by the
+    /// next caller's signal — meaning the cancelled caller's side effects
```
RobertZ2011, line 80: Same comment as above.

embedded-service/src/metrics.rs
```diff
@@ -0,0 +1,209 @@
...
+//! Use the counters to drive periodic diagnostic dumps or trip a degraded
+//! mode when a counter crosses a threshold.
+
+use crate::{AtomicUsize, Ordering};
+
+/// Saturating increment of a counter by `n`. Saturates at `usize::MAX`
+/// rather than wrapping so a long-running EC does not silently reset
+/// observability state.
+#[inline]
+fn bump_by(counter: &AtomicUsize, n: usize) {
```
RobertZ2011, line 31: Is there a code size increase from this if we never read any of these metrics? If the compiler/linker isn't able to optimize this away, then we should try putting this behind a feature.

Cargo.toml
```diff
@@ -46,15 +46,25 @@ repository = "https://github.com/OpenDevicePartnership/embedded-services"
...
+style = { level = "deny", priority = -1 }
 todo = "deny"
+# Workspace default: `allow`. The workspace-wide `warnings = "deny"` would
+# otherwise upgrade `warn` to a hard error. Individual crates that want to
+# enforce this lint can opt in at their crate root, e.g.:
+#
+#     #![deny(clippy::undocumented_unsafe_blocks)]
+#
+# `embedded-service` does exactly that to lock down its foundational unsafe
+# surface. Other crates may stay opted-out while their own unsafe paths are
```
RobertZ2011, line 65: We should track this with a bug and aim to eventually move `undocumented_unsafe_blocks ` over to deny by default.

RobertZ2011 (changes requested):

embedded-service/src/lib.rs
```diff
@@ -21,6 +29,7 @@ pub mod hid;
 pub mod init;
 pub mod ipc;
 pub mod keyboard;
```
jerrysxie, line 31: both keyboard and activity module are just example not being used, maybe we should delete them? @RobertZ2011 thoughts?

.opencode/agent/architect.md
```diff
@@ -0,0 +1,86 @@
+---
+description: Use when the task needs system design, API or module-boundary decisions, state-machine or invariant design, failure-domain analysis, or long-term architectural strategy. Produces specifications, not implementations. Trigger for "design", "architecture", "spec", "module boundary", "API shape", "state machine", "invariant", "should we", "how should this look".
```
jerrysxie, line 2: Does it make sense to add these opencode agent to the repo? Can non-opencode tool take advantage of this?

I briefly skim through them. They looks reasonable. If these files need to update as design changes, would responsibility fall on the contributor making the design change? Or should we have different ownership for these AI skills and agents? Just think aloud on how we keep these different AI toolings up to date.

jerrysxie (commented):
That was a lot of changes to review. If we can break these changes into smaller PRs, it would make reviewing them much easier.

embedded-service/src/lib.rs
RobertZ2011: I have no objections to deleting these.

PR #885, opened 2026-06-08
Rework spawn_service macro to relax constraints on constructors

williampMSFT:
The `spawn_service` macro / RunnableService trait were intended to standardize the way we launch services that need worker tasks in order to eliminate the repetitiveness required to do all the setup associated with this pattern (declaring static storage, declaring a worker task, etc).

However, as part of that, it also imposed the constraint that there's only one way to construct an instance of the service - using an `InitParams` type as an associated type on the `RunnableService` implementation.  This has the unfortunate consequence of meaning that you can't have multiple constructors and your constructors can't be generic over anything that the struct itself is not generic over, which is particularly annoying when you want to take a parameter by reference but only need it for the duration of the constructor.

This change reworks the RunnableService trait to not require a new()/InitParams function, and instead take as an argument a closure that yields a Result<(RunnableService, Runner)>.  This opens up a lot more flexibility around the shape of constructors while still getting the ergonomics properties of automatic storage declarations and automatic service starting.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -65,9 +65,8 @@ cd examples/std && cargo clippy --locked
 Services implement the `odp_service_common::runnable_service::Service<'hw>` trait, which enforces a consistent structure:
 
 1. **`Resources`** — caller-allocated state (stored in a `StaticCell`), not an internal `OnceLock` singleton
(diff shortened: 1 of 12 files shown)


kurtjd (approved):
Looks good and makes sense to me.

jerrysxie (approved):
In the end, there is always closure.

RobertZ2011 (approved):

PR #888, opened 2026-06-11
Add expected_response_size to GetReport command

daniellehadar:
Add an optional buffer size parameter to GetReport so the I2C device layer can limit the read length to the actual expected payload size instead of always reading the full buffer. This avoids over-reading response in write_read transactions.

### DIFF
--- a/embedded-service/src/hid/command.rs
+++ b/embedded-service/src/hid/command.rs
@@ -225,7 +225,7 @@ impl Opcode {
 #[allow(missing_docs)]
 pub enum Command<'a> {
     Reset,
-    GetReport(ReportType, ReportId),
+    GetReport(ReportType, ReportId, Option<u16>),
     SetReport(ReportType, ReportId, SharedRef<'a, u8>),
     GetIdle(ReportId),
     SetIdle(ReportId, ReportFreq),
@@ -239,7 +239,7 @@ impl From<Command<'_>> for Opcode {
     fn from(value: Command<'_>) -> Self {
         match value {
             Command::Reset => Opcode::Reset,
-            Command::GetReport(_, _) => Opcode::GetReport,
+            Command::GetReport(_, _, _) => Opcode::GetReport,
             Command::SetReport(_, _, _) => Opcode::SetReport,
             Command::GetIdle(_) => Opcode::GetIdle,
             Command::SetIdle(_, _) => Opcode::SetIdle,
@@ -293,6 +293,7 @@ impl<'a> Command<'a> {
         report_type: Option<ReportType>,
         report_id: Option<ReportId>,
         data: Option<SharedRef<'a, u8>>,
+        expected_response_size: Option<u16>,
     ) -> Result<Self, Error> {
         if opcode.requires_report_id() && report_id.is_none() {
             return Err(Error::RequiresReportId);
@@ -310,7 +311,11 @@ impl<'a> Command<'a> {
             Opcode::Reset => Command::Reset,
             Opcode::GetReport => {
                 if report_type? == ReportType::Input || report_type? == ReportType::Feature {
-                    Command::GetReport(report_type?, report_id.ok_or(Error::RequiresReportId)?)
+                    Command::GetReport(
+                        report_type?,
+                        report_id.ok_or(Error::RequiresReportId)?,
+                        expected_response_size,
+                    )
                 } else {
                     return Err(Error::InvalidReportType);
                 }
@@ -476,7 +481,7 @@ impl<'a> Command<'a> {
                 let (command_len, _) = Self::encode_basic_op(buf, Opcode::Reset)?;
                 len += command_len;
             }
-            Command::GetReport(report_type, report_id) => {
+            Command::GetReport(report_type, report_id, _expected_size) => {
                 let (command_len, buf) = Self::encode_common(buf, Opcode::GetReport, Some(*report_type), *report_id)?;
                 len += command_len;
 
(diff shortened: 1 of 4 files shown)


RobertZ2011 (dismissed):

hid-service/src/i2c/device.rs
```diff
@@ -234,7 +251,24 @@ impl<A: AddressMode + Copy, B: I2c<A>> Device<A, B> {
                 Error::Bus(e)
             })?;
 
-            Ok(Some(Response::FeatureReport(self.buffer.reference())))
+            if constrained {
+                let actual_frame_len =
+                    u16::from_le_bytes([read_buf[LENGTH_LO_OFFSET], read_buf[LENGTH_HI_OFFSET]]) as usize;
```
jerrysxie: This will fail CI as no panic clippy lints are enabled. We can use `at()` to return a result.

jerrysxie: Instead of indexing twice, since we know the length is always in the front of the buffer, can we do something like this instead?

```
                let actual_frame_len = read_buf
                    .first_chunk::<LENGTH_PREFIX_SIZE>()
                    .map(|b| u16::from_le_bytes(*b) as usize)
                    .ok_or(Error::Hid(hid::Error::InvalidSize(InvalidSizeError {
                        expected: LENGTH_PREFIX_SIZE,
                        actual: read_buf.len(),
                    })))?;
```

hid-service/src/i2c/device.rs
```diff
@@ -234,7 +251,24 @@ impl<A: AddressMode + Copy, B: I2c<A>> Device<A, B> {
...
+                if actual_frame_len != response_size {
```
jerrysxie: This is a strict comparison check so the device has to return an exact size. But according to the hidi2c spec, the device may return fewer bytes than the max size for a report. Would this work for other passthrough cases? @wmmc88 @RobertZ2011

embedded-service/src/hid/command.rs
```diff
@@ -225,7 +225,7 @@ impl Opcode {
 #[allow(missing_docs)]
 pub enum Command<'a> {
     Reset,
-    GetReport(ReportType, ReportId),
+    GetReport(ReportType, ReportId, Option<u16>),
```
jerrysxie: non-blocking: not loving this unamed tuple element that is u16? Maybe instead of tuple, we can use struct?

```
    GetReport {
        report_type: ReportType,
        report_id: ReportId,
        ExpectedPayloadSize: Option<u16>,
    },
```

embedded-service/src/hid/command.rs
```diff
@@ -225,7 +225,7 @@ impl Opcode {
...
     SetReport(ReportType, ReportId, SharedRef<'a, u8>),
```
jerrysxie, line 236: This additional size constraint is very specific to a I2C transport, so it feels a bit weird to thread it into the general HID command layer. Is this expected size expected to change on the fly dynamically? It feels like this should belong to the I2C device layer instead, but currently there is isno good separation between I2C and HID in the current code.  What do you think? @williampMSFT Leave this here for now, since it will be refactored soon anyway. Will the refactored code take this case into account?

jerrysxie (changes requested):

hid-service/src/i2c/device.rs
RobertZ2011: We should support fewer bytes if the spec says that it's valid.

daniellehadar: Agree, the spec allows shorter reports. We still want this check to avoid accepting oversized or malformed data, so I suggest:
`if actual_frame_len > response_size`
What do you think?

RobertZ2011: I think that should be fine.

jerrysxie: That works for me.

embedded-service/src/hid/command.rs
williampMSFT: Agree that it's a bit weird here but embedded-service/src/hid already conflates some transport-agnostic stuff with HID-I2C stuff so maybe it's not weirder than what's already in here. I do not plan to carry this file over in the rewrite so I'm inclined to be more permissive with hacks than I otherwise would be

jerrysxie (dismissed):
Approved, pending fix for one failed CI on formatting.

hid-service/src/i2c/device.rs
```diff
@@ -234,7 +250,29 @@ impl<A: AddressMode + Copy, B: I2c<A>> Device<A, B> {
...
+                        actual_frame_len, response_size
+                    );
+                    return Err(Error::Hid(hid::Error::InvalidSize(InvalidSizeError {
+                        expected: response_size,
+                        actual: actual_frame_len,
+                    })));
+                }
+            }
+
+            Ok(Some(Response::FeatureReport(
```
razbhe10, line 276: if the actual_frame_len is smaller than the response_size, I think we should return the actual_frame_len 
```
if actual_frame_len > response_size { return Err(/* InvalidSize */); }
let returned_len = if constrained { actual_frame_len } else { response_size };
Ok(Some(Response::FeatureReport(
    self.buffer.reference().slice(0..returned_len).map_err(Error::Buffer)?,
)))
```

embedded-service/src/hid/command.rs
razbhe10, line 669: Do we want to add the relevant tests?
(a) constrained read pulls exactly payload + 2 bytes and returns a slice of that length;
(b) device declaring a longer frame (actual_frame_len > response_size) → Err(InvalidSize);
(c) expected_payload_size exceeding the buffer → Err(InvalidSize) from get_mut.

jerrysxie: It will be a bit tough to enable unit test in this module without some refactoring due to the static lifetime and we will need mocks for the i2c bus. Since @williampMSFT is refactoring this module right now, it is probably out of the scope for this PR.

hid-service/src/i2c/device.rs
daniellehadar: Correct, thanks!

razbhe10 (approved):

jerrysxie (approved):

RobertZ2011 (approved):

Issue #830, opened 2026-05-06
Type-C: Create `type-c-service::controller::Port` unit test for plug -> new provider contract -> unplug


RobertZ2011:
Write test coverage for this behavior.

- [ ] Test internal state changes and ensure appropriate events are broadcast to the power policy service.

PR #891, opened 2026-06-12
type-C: Port corrected implicit source contract logic

### LINKED ISSUE #752: Port corrected implicit source contract logic to v0.2.0
Main PR here: https://github.com/OpenDevicePartnership/embedded-services/pull/751

RobertZ2011:
Non-PD capable devices do not result in `new_power_contract_as_provider` contracts. Update logic to detect changes to the provider contract that works for PD and non-PD capable devices.

Logic ported from #751

### DIFF
--- a/type-c-service/src/controller/mod.rs
+++ b/type-c-service/src/controller/mod.rs
@@ -132,7 +132,7 @@ impl<
             self.process_new_consumer_contract(&new_status).await?;
         }
 
-        if status_event.new_power_contract_as_provider() {
+        if new_status.is_connected() && new_status.available_source_contract != self.status.available_source_contract {
             self.process_new_provider_contract(&new_status).await?;
         }
 


kurtjd (approved):

jerrysxie (commented):
Can we add a test case for this? Is that possible with the mock we have? Do we need to create a non-PD capable mock?

RobertZ2011:
> Can we add a test case for this? Is that possible with the mock we have? Do we need to create a non-PD capable mock?

This will get covered by https://github.com/OpenDevicePartnership/embedded-services/issues/830, but I want Matteo to handle that to get more Type-C experience.

jerrysxie:
> > Can we add a test case for this? Is that possible with the mock we have? Do we need to create a non-PD capable mock?
> 
> This will get covered by #830, but I want Matteo to handle that to get more Type-C experience.

@RobertZ2011 Looking at this issue, I can't tell whether the non-PD capable case is covered. Can we add to the issue to make sure this case is covered?

jerrysxie (approved):

kurtjd (approved):

tullom (approved):

PR #892, opened 2026-06-12
power-policy-service: Introduce customization

RobertZ2011:
OEMs may want to provide their own behavior to determine which PSU is selected as the next consumer. Introduce a `Customization` trait to contain this and anything else that may be needed in the future.

### DIFF
--- a/power-policy-service/src/service/consumer.rs
+++ b/power-policy-service/src/service/consumer.rs
@@ -2,6 +2,8 @@ use core::cmp::Ordering;
 use embedded_services::error;
 use embedded_services::named::Named;
 
+use crate::service::config::Config;
+
 use super::*;
 
 use power_policy_interface::psu;
@@ -26,12 +28,12 @@ impl<'device, Psu: Lockable<Inner: psu::Psu>> Clone for AvailableConsumer<'devic
 
 impl<'device, Psu: Lockable<Inner: psu::Psu>> Copy for AvailableConsumer<'device, Psu> {}
 
-/// Compare two consumer capabilities to determine which one is better
+/// Default function for comparing two consumer capabilities to determine which one is better
 ///
 /// This is not part of the `Ord` implementation for `ConsumerPowerCapability`, because it's specific to this implementation.
 /// *_is_current indicate if the device with that capability is the currently connected consumer. This is used to make the
 /// implementation stick so as to avoid switching between otherwise equivalent consumers.
-fn cmp_consumer_capability(
+pub fn cmp_consumer_capability_default(
     a: &ConsumerPowerCapability,
     a_is_current: bool,
     b: &ConsumerPowerCapability,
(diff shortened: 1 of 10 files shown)


power-policy-service/src/service/consumer.rs
```diff
@@ -40,61 +42,70 @@ fn cmp_consumer_capability(
...
+                    &available,
+                    current_consumer.is_some_and(|current_consumer| ptr::eq(current_consumer, *psu)),
+                    &best.consumer_power_capability,
+                    current_consumer.is_some_and(|current_consumer| ptr::eq(current_consumer, best.psu)),
+                ) == core::cmp::Ordering::Greater
+                {
+                    Some(AvailableConsumer {
+                        psu,
+                        consumer_power_capability: available,
+                    })
```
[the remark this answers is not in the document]
RobertZ2011: Compiles as is.

power-policy-service/src/service/hooks.rs
```diff
@@ -0,0 +1,27 @@
...
+
+/// Power policy service hooks
+pub trait Hooks<'device, Reg: Registration<'device>> {
+    /// Find the best available consumer based on the current state and configuration.
+    fn find_best_consumer(
+        &mut self,
+        config: &Config,
+        state: &InternalState<'device, Reg::Psu>,
+        registration: &Reg,
+    ) -> impl Future<Output = Result<Option<AvailableConsumer<'device, Reg::Psu>>, Error>> {
```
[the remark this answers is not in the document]
RobertZ2011: Fine as is.

power-policy-service/src/service/mod.rs
```diff
@@ -27,18 +28,18 @@ use crate::service::registration::Registration;
...
+    pub current_consumer_state: Option<consumer::AvailableConsumer<'device, PSU>>,
     /// Current provider global state
-    current_provider_state: provider::State,
+    pub current_provider_state: provider::State,
     /// System unconstrained power
-    unconstrained: UnconstrainedState,
+    pub unconstrained: UnconstrainedState,
     /// Connected providers
-    connected_providers: heapless::index_set::FnvIndexSet<usize, MAX_CONNECTED_PROVIDERS>,
+    pub connected_providers: heapless::index_set::FnvIndexSet<usize, MAX_CONNECTED_PROVIDERS>,
```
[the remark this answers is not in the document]
RobertZ2011: This is intentional as I don't want to limit the type of functionality that hooks can implement. The fields in the `Service` struct aren't public so these are only accessible in the context of a hook call.

power-policy-service/src/service/mod.rs
```diff
@@ -56,22 +57,32 @@ where
...
     /// State
     state: InternalState<'device, Reg::Psu>,
     /// Config
     config: config::Config,
+    /// Hooks
+    hooks: Hooks,
 }
 
-impl<'device, Reg: Registration<'device>> Service<'device, Reg> {
+impl<'device, Reg: Registration<'device>, Hooks: hooks::Hooks<'device, Reg> + Default> Service<'device, Reg, Hooks> {
```
jerrysxie: Do we need to thread this through `psu_task` and other tasks that takes service with only 2 generics:

```
pub async fn psu_task<
    'device,
    const PSU_COUNT: usize,
    S: Lockable<Inner = Service<'device, Reg>>,
    Reg: Registration<'device>,
    PsuReceiver: Receiver<EventData>,
>
```

We should create a test that actually utilize this custom hook.

jerrysxie (changes requested):
This additional hook allow OEM to add custom behavior. Is the idea that the hook traits will be expanded in the future to add more extension point? Instead of calling it hook, can we call this trait extension or some extension trait?

In the PR description, can we add some content to explain the motivation behind this.

power-policy-service/src/service/mod.rs
RobertZ2011: Good catch on the tasks. I've also added a test to verify that the currently implemented hook function gets called.

RobertZ2011:
> This additional hook allow OEM to add custom behavior. Is the idea that the hook traits will be expanded in the future to add more extension point? Instead of calling it hook, can we call this trait extension or some extension trait?
> 
> In the PR description, can we add some content to explain the motivation behind this.

This is meant to be where we put any sort of functionality that the OEM may want to customize by providing their own implementation. Ideally, we should be able to add to this trait in the future without introducing a breaking change by ensuring the trait provides default implementations. I ended up going with `Customization` because I think that best matches the intent here. Also added a summary to the PR description.

power-policy-service/src/service/consumer.rs
```diff
@@ -40,61 +42,72 @@ fn cmp_consumer_capability(
...
+                    best_consumer
+                }
+            }
+        };
     }
 
+    Ok(best_consumer)
+}
+
+impl<'device, Reg: Registration<'device>, Cusomization: customization::Customization>
```
jerrysxie: Cusomization -> Customization

power-policy-service/tests/consumer.rs
```diff
@@ -660,40 +683,183 @@ impl Test for TestNoSwap {
...
+                psu: psu0,
+                consumer_power_capability,
+            }))
+        } else {
+            find_best_consumer_default(config, state, registration, cmp_consumer_capability_default).await
+        }
+    }
+}
+
+/// Verify that [`customization::Customization::find_find_best_consumer`] is called
```
jerrysxie: find_find -> find

power-policy-service/tests/consumer.rs
```diff
@@ -660,40 +683,183 @@ impl Test for TestNoSwap {
...
 }
 
 #[tokio::test]
 async fn run_test_no_swap() {
-    run_test(DEFAULT_TIMEOUT, TestNoSwap, Default::default()).await;
+    run_test(DEFAULT_TIMEOUT, TestNoSwap, Default::default(), DefaultCustomization).await;
+}
+
+#[tokio::test]
+async fn run_test_find_best_consumer_hook() {
```
jerrysxie, line 857: nit: this still reference `hook`.

jerrysxie (dismissed):

power-policy-service/src/service/consumer.rs
RobertZ2011: Done.

power-policy-service/tests/consumer.rs
RobertZ2011: Done.

power-policy-service/tests/consumer.rs
RobertZ2011: Done.

jerrysxie (approved):

kurtjd (approved):

tullom (approved):

PR #893, opened 2026-06-12
Refactor battery service to use a new fuel gauge registration system and move away from intrusive list and static lifetimes

tullom:
- Introduced a `Registration` trait and an `ArrayRegistration` implementation to manage fuel gauges.
- OEMs handle their own state for each fuel gauge.
- Replaced the `MockBatteryDriver` with `MockFuelGauge` to align with the new fuel gauge structure.
- Removed the `wrapper.rs` and `controller.rs` files as its functionality is now integrated into the main battery service logic.
- Modified the battery example applications to utilize the new registration system and updated the fuel gauge initialization and recovery logic.


battery-service-interface/src/fuel_gauge.rs
```diff
@@ -0,0 +1,315 @@
...
+        Self::BusError
+    }
+}
+
+/// Standard static battery data cache.
+#[derive(Default, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct StaticBatteryMsgs {
+    /// Manufacturer Name.
+    pub manufacturer_name: [u8; 21],
```
RobertZ2011: Can you add constants for these lengths/

battery-service/src/acpi.rs
```diff
@@ -181,49 +188,63 @@ pub(crate) fn compute_pif(psu_state: &PsuState) -> PifFixedStrings {
...
         trace!("Battery service: got PSR command!");
 
-        let _fg = self.get_fuel_gauge(device_id).ok_or(BatteryError::UnknownDeviceId)?;
+        let _fg = self
+            .registration
+            .get_fuel_gauge(device_id)
+            .ok_or(BatteryError::UnknownDeviceId)?;
 
-        Ok(compute_psr(&self.get_power_info().await))
+        Ok(compute_psr(&PsuState::default()))
```
RobertZ2011: Is this intentional? I see multiple places where the code operates on `PsuState::default`.

battery-service/src/acpi.rs
```diff
@@ -347,8 +394,181 @@ impl crate::context::Context {
...
+///
+/// TODO: Use this over DeviceId based approach?
+impl<'hw, Reg: crate::registration::Registration<'hw>> crate::Service<'hw, Reg> {
+    /// Queries the estimated time remaining until the battery reaches the specified charge level. Corresponds to ACPI's _BCT method.
+    pub fn battery_charge_time(
+        &self,
+        fuel_gauge: &mut <Reg::FuelGauge as Lockable>::Inner,
+        bct: embedded_batteries_async::acpi::Bct,
+    ) -> Result<BctReturnResult, BatteryError> {
+        trace!("Battery service: got BCT command!");
```
RobertZ2011, line 222: There are a couple of `trace!` calls like this that reem redundant.

battery-service/src/mock.rs
```diff
@@ -143,17 +141,18 @@ impl crate::controller::Controller for MockBatteryDriver {
             turbo_vload_mv: 0,
             turbo_rhf_effective_mohm: 0,
         };
-        Ok(new_msgs)
+        self.state_mut().on_dynamic_data(new_msgs);
+        Ok(())
     }
 
     #[allow(clippy::indexing_slicing)]
```
RobertZ2011, line 154: Can you add a comment documenting why this is required and what guarantees that this wont panic?

battery-service/src/registration.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// The position of a fuel gauge in this slice is its
+    /// [`DeviceId`] for ACPI queries (the
+    /// first registered fuel gauge is battery `0`, and so on).
+    fn fuel_gauges(&self) -> &[&'hw Self::FuelGauge];
+
+    /// Look up a registered fuel gauge by its device ID.
+    ///
+    /// The device ID is the fuel gauge's position in the registration (the first
+    /// registered fuel gauge is battery `0`, and so on).
+    fn get_fuel_gauge(&self, id: DeviceId) -> Option<&'hw Self::FuelGauge> {
```
RobertZ2011, line 31: We're wanting to move away from using numeric device IDs, is there a plan to address this in a follow-up PR?

RobertZ2011 (changes requested):

battery-service-interface/src/lib.rs
```diff
@@ -174,8 +176,11 @@ pub trait BatteryService {
         btp: Btp,
     ) -> impl core::future::Future<Output = Result<(), BatteryError>>;
 
-    /// Queries whether the battery is currently in use (i.e., providing power to the system). Corresponds to ACPI's _PSR method.
-    fn is_in_use(&self, battery_id: DeviceId) -> impl core::future::Future<Output = Result<PsrReturn, BatteryError>>;
+    /// Queries whether the power supply unit is currently in use (i.e., providing power to the system). Corresponds to ACPI's _PSR method.
+    fn is_psu_in_use(
+        &self,
+        battery_id: DeviceId,
```
kurtjd: I'm not sure of the context behind this name change, but would it make sense to rename `battery_id` to `psu_id`? Just seems odd to mix battery/psu here (from the perspective of someone who doesn't understand this subsystem well, so maybe there is good reason).

kurtjd (dismissed):
Although I'm not very knowledgeable in this area, it all looks good to me and I didn't see any issues other than a minor thing I pointed out.

battery-service/src/registration.rs
jerrysxie: It seems the interface that takes in a DeviceId is purely for ACPI, instead of having 2 `battery_info`: one takes in a reference and one takes in an id. Can we combine them into a single one and just which translate id into reference or vice versa?

battery-service/src/acpi.rs
```diff
@@ -181,49 +188,63 @@ pub(crate) fn compute_pif(psu_state: &PsuState) -> PifFixedStrings {
...
         trace!("Battery service: got PIF command!");
 
-        let _fg = self.get_fuel_gauge(device_id).ok_or(BatteryError::UnknownDeviceId)?;
+        let _fg = self
+            .registration
+            .get_fuel_gauge(device_id)
+            .ok_or(BatteryError::UnknownDeviceId)?;
 
-        Ok(compute_pif(&self.get_power_info().await))
+        Ok(compute_pif(&PsuState::default()))
```
jerrysxie: Is this a placeholder? Can we add some comment to explain why?

battery-service/src/lib.rs
```diff
@@ -1,302 +1,127 @@
...
+pub use battery_service_interface::{BatteryService, DeviceId};
+
+/// The battery service.
+///
+/// Owns the [`Registration`] that provides the set of fuel gauges, and answers
+/// ACPI battery queries (via the [`BatteryService`] trait) by reading each
+/// registered fuel gauge's cached state. The OEM drives each registered fuel
+/// gauge directly through the [`FuelGauge`] trait methods.
+pub struct Service<'hw, Reg: Registration<'hw>> {
+    pub registration: Reg,
```
jerrysxie: Do we intend to explore the implementation detail of this struct to external clients?

jerrysxie (changes requested):

battery-service/src/registration.rs
tullom: Good question, i was debating fully moving away from numeric device IDs, but i think it'll require more discussion with the team. Like Jerry said, DeviceId is still used for ACPI, since ACPI models a single battery per ACPI device, so when multiple batteries are present we need a way to tell them apart. That distinction currently lives at the MCTP layer at its endpoint addressing.

Longer term, i think we should bind an ACPI device directly to a FuelGauge, which would let us move away from DeviceId. That's a larger architectural change so i want to gather feedback from the Phil and Billy before settling on an approach and plan to address it in a follow up PR rather than here.

battery-service/src/acpi.rs
tullom: This is intentional for now, will be addressed in a future PR when i get the power-policy integration into battery service hammered down. The PsuState struct has a doc comment for now:
```rust
/// Currently always the default; this is a placeholder for future power policy
/// integration.
```

battery-service/src/acpi.rs
tullom: This is intentional for now, will be addressed in a future PR when i get the power-policy integration into battery service hammered down. The PsuState struct has a doc comment for now:

```rust
/// Currently always the default; this is a placeholder for future power policy
/// integration.
```

battery-service/src/acpi.rs
tullom: I removed all the handler methods and call the new methods that take in a mut reference to the fuel gauge. Less duplication now and users can call the by-ref methods themselves. In the future, gonna change the service interface trait to move over to this method and remove DeviceId entirely.

battery-service/src/lib.rs
tullom: Its not specifically needed, but we want to expose the fuel_gauges() and get_fuel_gauge() methods to users and this was the easiest approach. I moved to making the field private and exposing 2 public methods that call those underlying trait methods.

battery-service-interface/src/lib.rs
tullom: Yeah it does, done. In the future, i think these methods need to be moved to power-policy anyways.

battery-service/src/mock.rs
tullom: Done.

battery-service-interface/src/fuel_gauge.rs
tullom: Done.

kurtjd (approved):

jerrysxie (approved):

RobertZ2011 (approved):

thdesil (approved):

PR #895, opened 2026-06-15
Make battery data caches OEM extensible

tullom:
- Introduce StaticBatteryData and DynamicBatteryData traits and make the
fuel-gauge State generic over them (State<S, D>) so OEM drivers can cache
a custom type that embeds the standard StaticBatteryMsgs / Dynamic
BatteryMsgs and adds extra fields, while the battery service still reads
the standard fields needed to answer ACPI queries.

Rework the cache data model in the process:

- Encode capacity and rate quantities with the Smart Battery
  CapacityModeValue / CapacityModeSignedValue enums so each value
  self-describes its unit (mA/mAh vs centiWatt) instead of assuming a
  fixed encoding.
- Drop unit-suffixed field names in favor of the embedded-batteries
  typed aliases (MilliVolts, MilliAmps, Percent, Cycles, DeciKelvin,
  Minutes), keeping units in the types and docs.
- Add the previously missing ACPI/SBS fields (manufacture_date,
  specification_info, remaining_capacity_alarm, at_rate and its
  predictions, run_time_to_empty, average_time_to_*), and document
  every field.

Since the capacity enums have no Default, replace the derived Default on
both cache structs with manual impls that default to the mA-zero variant.

Update the mock fuel gauge to source every SmartBattery getter from the
cached State and add coherent emulated-pack constructors (new, new_2s,
new_4s) that scale pack voltage and power with the series cell count.

Change State::on_static_data / on_dynamic_data to take an in-place
updater closure (FnOnce(&mut S/D)) instead of consuming the data by
value. The closure writes freshly read values directly into the cache
storage, so a (potentially large or OEM-extended) S/D is never moved or
copied through the call.

Depends on #893

### DIFF
--- a/battery-service-interface/src/fuel_gauge.rs
+++ b/battery-service-interface/src/fuel_gauge.rs
@@ -11,7 +11,11 @@ use core::future::Future;
 
 use embedded_batteries_async::{
     acpi::{BmcControlFlags, BmdCapabilityFlags, BmdStatusFlags, PowerThresholdSupport},
-    smart_battery::BatteryModeFields,
+    charger::{MilliAmps, MilliVolts},
+    smart_battery::{
+        BatteryModeFields, CapacityModeSignedValue, CapacityModeValue, Cycles, DeciKelvin, ManufactureDate,
+        MilliAmpsSigned, Minutes, Percent,
+    },
 };
 
 /// Fuel gauge errors.
@@ -43,7 +47,7 @@ pub const DEVICE_CHEMISTRY_ID_SIZE: usize = 2;
 pub const SERIAL_NUM_SIZE: usize = 4;
 
 /// Standard static battery data cache.
-#[derive(Default, Clone, Copy)]
+#[derive(Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct StaticBatteryMsgs {
     /// Manufacturer Name.
@@ -55,11 +59,11 @@ pub struct StaticBatteryMsgs {
     /// Device Chemistry.
     pub device_chemistry: [u8; DEVICE_CHEMISTRY_SIZE],
 
-    /// Design Capacity in mWh.
-    pub design_capacity_mwh: u32,
+    /// Design Capacity. Units (mA/mAh or centiWatt) are encoded by the [`CapacityModeValue`] variant.
+    pub design_capacity: CapacityModeValue,
 
     /// Design Voltage in mV.
-    pub design_voltage_mv: u16,
+    pub design_voltage: MilliVolts,
 
     /// Device Chemistry Id.
     pub device_chemistry_id: [u8; DEVICE_CHEMISTRY_ID_SIZE],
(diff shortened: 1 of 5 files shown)


battery-service/src/acpi.rs
```diff
@@ -28,37 +46,39 @@ pub(crate) fn compute_bst(cache: &DynamicBatteryMsgs) -> embedded_batteries_asyn
...
         battery_technology: embedded_batteries_async::acpi::BatteryTechnology::Secondary,
-        design_voltage: static_cache.design_voltage_mv.into(),
-        design_cap_of_warning: static_cache.design_cap_warning,
-        design_cap_of_low: static_cache.design_cap_low,
+        design_voltage: static_cache.design_voltage.into(),
+        design_cap_of_warning: capacity_raw(static_cache.design_cap_warning),
+        design_cap_of_low: capacity_raw(static_cache.design_cap_low),
         cycle_count: dynamic_cache.cycle_count.into(),
-        measurement_accuracy: u32::from(100 - dynamic_cache.max_error_pct) * 1000u32,
+        measurement_accuracy: u32::from(100 - dynamic_cache.max_error) * 1000u32,
```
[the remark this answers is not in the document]
tullom: Done.

battery-service/src/mock.rs
```diff
@@ -1,179 +1,310 @@
...
+            design_capacity: self.design_capacity().await?,
+            design_voltage: self.design_voltage().await?,
             device_chemistry_id: Default::default(),
             serial_num: Default::default(),
             battery_mode: self.battery_mode().await?,
-            design_cap_warning: design_capacity / 4,
-            design_cap_low: design_capacity / 10,
+            design_cap_warning: smart_battery::CapacityModeValue::MilliAmpUnsigned((design_capacity_value / 4) as u16),
+            design_cap_low: smart_battery::CapacityModeValue::MilliAmpUnsigned((design_capacity_value / 10) as u16),
             measurement_accuracy: self.max_error().await?.into(),
```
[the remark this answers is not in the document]
tullom: Done.

battery-service/src/mock.rs
```diff
@@ -51,24 +51,149 @@ where
...
+
+    /// Build a [`MockFuelGauge`] preloaded with coherent data for a `cells`-cell
+    /// series Li-ion pack reporting `device_name` as its model string.
+    ///
+    /// Every static and dynamic cache field is populated. Pack voltages scale with
+    /// the series cell count (3.7 V nominal, 4.2 V full charge and 3.0 V load
+    /// cutoff per cell); the 3000 mAh cell capacity and discharge currents are the
+    /// same across topologies, so the reported power quantities scale with the pack
+    /// voltage. Capacity and rate are reported in current units (mA/mAh).
+    fn with_series_cells(cells: u16, device_name: [u8; 21]) -> Self {
```
RobertZ2011: Can you introduce named constants for this functions. It's not immediately clear what something like `d.at_rate_time_to_empty = 86;` means.

battery-service/src/mock.rs
```diff
@@ -116,95 +243,113 @@ impl FuelGauge for MockFuelGauge {
...
-        self.state_mut().on_static_data(new_msgs);
+        let design_voltage = self.design_voltage().await?;
+        let battery_mode = self.battery_mode().await?;
+        let measurement_accuracy: u32 = self.max_error().await?.into();
+        let manufacture_date = self.manufacture_date().await?;
+        let specification_info: u16 = self.specification_info().await?.into();
+        let remaining_capacity_alarm = self.remaining_capacity_alarm().await?;
+        let remaining_time_alarm = self.remaining_time_alarm().await?;
+
+        let mut manufacturer_name = [0u8; 21];
```
RobertZ2011: Same comment on constants for these lengths.

battery-service/src/mock.rs
```diff
@@ -116,95 +243,113 @@ impl FuelGauge for MockFuelGauge {
...
+            s.manufacturer_name = manufacturer_name;
+            s.device_name = device_name;
+            s.device_chemistry = device_chemistry;
+            s.design_capacity = design_capacity;
+            s.design_voltage = design_voltage;
+            s.device_chemistry_id = device_chemistry_id;
+            s.serial_num = [serial_lsb, serial_msb, 0, 0];
+            s.battery_mode = battery_mode;
+            s.design_cap_warning =
+                smart_battery::CapacityModeValue::MilliAmpUnsigned((design_capacity_value / 4) as u16);
```
RobertZ2011: Same constant comment for 4 and 10.

battery-service-interface/src/fuel_gauge.rs
```diff
@@ -59,94 +63,278 @@ pub struct StaticBatteryMsgs {
...
+    }
 }
 
 /// Standard dynamic battery data cache.
-#[derive(Default, Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct DynamicBatteryMsgs {
     /// Battery Max Power in mW.
-    pub max_power_mw: u32,
```
RobertZ2011, line 113: We should keep the unit suffix if we don't have appropriate unit types.

RobertZ2011 (changes requested):

battery-service/src/mock.rs
tullom: Done, added constants for everything and comments.

battery-service/src/mock.rs
tullom: Done, added constants for everything and comments.

battery-service-interface/src/fuel_gauge.rs
tullom: Done.

RobertZ2011 (dismissed):

thdesil (dismissed):

RobertZ2011 (approved):

thdesil (approved):

PR #899, opened 2026-06-18
Disable sink path during voltage transition

### LINKED ISSUE #874: type-C: Port sink disable changes to main
Port https://github.com/OpenDevicePartnership/embedded-services/pull/861 to main. Also add test coverage for the sink disable logic and the new arguments added to the power policy `ConsumerDisconnect` event.

tullom:
- Disable the sink path before a max sink voltage change when a consumer is connected and the limit is changing or being removed, to avoid overcurrent/overvoltage during renegotiation.
- Add a ConsumerDisconnect flags argument to the power policy consumer disconnected event: renegotiation when the consumer is recontracting, switching when moving to a different PSU.
- Add tests for the sink disable logic and the new flags.

Resolves #874 and is a port of #861

### DIFF
--- a/type-c-service/src/controller/max_sink_voltage.rs
+++ b/type-c-service/src/controller/max_sink_voltage.rs
@@ -17,10 +17,24 @@ impl<
     for Port<'device, C, Shared, TypeCSender, PowerSender, LoopbackSender>
 {
     async fn set_max_sink_voltage(&mut self, voltage_mv: Option<u16>) -> Result<(), PdError> {
-        self.controller
-            .lock()
-            .await
-            .set_max_sink_voltage(self.port, voltage_mv)
-            .await
+        // A change in the maximum sink voltage can trigger a PD renegotiation. During that transition the
+        // source may briefly output a voltage that does not match the active contract, which can cause an
+        // overcurrent/overvoltage condition on the sink path. If we currently have a connected consumer and
+        // the limit is actually changing (or being removed), disable the sink path before the renegotiation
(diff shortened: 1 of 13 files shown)


RobertZ2011 (approved):

type-c-service/src/controller/max_sink_voltage.rs
```diff
@@ -17,10 +17,24 @@ impl<
...
+        // to protect the system. The power policy re-enables the sink path when it connects the consumer to
+        // the renegotiated contract.
+        let disable_sink_path = match self.psu_state.psu_state {
+            PsuState::ConnectedConsumer(capability) => {
+                voltage_mv.is_none() || voltage_mv != Some(capability.capability.voltage_mv)
+            }
+            _ => false,
+        };
+
+        let mut controller = self.controller.lock().await;
```
RobertZ2011: It doesn't look like you're notifying the power policy that we've disconnected. `TestSinkDisableOnVoltageChange` should also verify that the power policy broadcasts a `ConsumerDisconnected`.

RobertZ2011 (changes requested):

type-c-service/src/controller/max_sink_voltage.rs
tullom: Done, we do a `self.psu_state.disconnect(true)` and a send a `power_policy_interface::psu::event::EventData::Disconnected(ConsumerDisconnect::none().with_renegotiation(true))` with the power_policy sender

RobertZ2011 (approved):

kurtjd (approved):

PR #906, opened 2026-06-26
uart-service: add default_mctp_serial constructor

dymk:
Adds `MctpSerialService<R>` and `default_mctp_serial()` to `uart-service` — the `MctpSerialMedium` twin of the existing `DefaultService` / `default_smbusespi` pair.

The constructor hardcodes the canonical EC↔SP serial reply context (`source = EC_EID`, `destination = SP_EID`, `message_tag = 0`, `medium_context = ()`), mirroring `default_smbusespi` field-for-field. Serial consumers (the QEMU EC↔SP relay path) can then call one constructor instead of hand-assembling an `MctpReplyContext`. Enables `mctp-rs/serial` so the `MctpSerialMedium` type is available.

No behaviour change to existing callers — purely additive (a new type alias + constructor, plus turning on the `serial` feature of the `mctp-rs` dependency).

Unblocks the EC `dev-qemu` MCTP-over-serial work in `odp-embedded-controller`, which will call this constructor to speak the SP's DSP0253 framing over the two-QEMU serial link.

### DIFF
--- a/uart-service/Cargo.toml
+++ b/uart-service/Cargo.toml
@@ -18,7 +18,7 @@ embedded-services.workspace = true
 defmt = { workspace = true, optional = true }
 log = { workspace = true, optional = true }
 embassy-sync.workspace = true
-mctp-rs = { workspace = true }
+mctp-rs = { workspace = true, features = ["serial"] }
 embedded-io-async.workspace = true
 
 [features]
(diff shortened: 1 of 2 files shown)


kurtjd (dismissed):
Is the plan to follow this up with a PR in odp-embedded-controller changing the constructor used (on all platforms, not just dev-qemu) then a co-PR in odp-platform-common updating the serial backend of the TUI to use this new framing?

dymk:
@kurtjd Correct, that's basically the plan.

RobertZ2011 (approved):

kurtjd (approved):

PR #907, opened 2026-06-26
type-c-service/wrapper: Add recovery logic to `set_max_sink_voltage`

RobertZ2011:
Since the PD controller is responsible for picking a contract, we can't determine if a renegotiation will actually happen. The current behavior can cause a failure to consume power if no renegotiation happens. Start a sink ready timeout when limiting the maximum sink voltage to allow us to recover in cases where no renegotiation happens.

### DIFF
--- a/supply-chain/imports.lock
+++ b/supply-chain/imports.lock
@@ -9,8 +9,8 @@ user-login = "BurntSushi"
 user-name = "Andrew Gallant"
 
 [[publisher.anyhow]]
-version = "1.0.99"
-when = "2025-08-11"
+version = "1.0.103"
+when = "2026-06-25"
 user-id = 3618
 user-login = "dtolnay"
 user-name = "David Tolnay"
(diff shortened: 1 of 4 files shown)


type-c-service/src/wrapper/pd.rs
```diff
@@ -290,8 +315,17 @@ where
...
+                        {
+                            Ok(port_state) => {
+                                self.process_set_max_sink_voltage(controller, port_state, local_port, voltage_mv)
+                                    .await
+                            }
+                            Err(e) => Err(e),
+                        }
                     }
                     Err(e) => Err(e),
                 }
```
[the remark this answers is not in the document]
RobertZ2011: Done.

supply-chain/imports.lock
```diff
@@ -368,7 +368,10 @@ aggregated-from = "https://raw.githubusercontent.com/OpenDevicePartnership/embas
 who = "Felipe Balbi <febalbi@microsoft.com>"
 criteria = "safe-to-deploy"
 version = "0.5.0"
-aggregated-from = "https://raw.githubusercontent.com/OpenDevicePartnership/tps6699x/refs/heads/main/supply-chain/audits.toml"
+aggregated-from = [
+    "https://raw.githubusercontent.com/OpenDevicePartnership/tps6699x/refs/heads/main/supply-chain/audits.toml",
+    "https://raw.githubusercontent.com/OpenDevicePartnership/tps6699x/refs/heads/main/supply-chain/audits.toml",
```
kurtjd, line 373: Weird, I'm guessing this was auto generated though wonder why it repeats this line twice.

kurtjd (approved):

asasine (approved):

supply-chain/imports.lock
RobertZ2011: I just tried restoring these lines and running `cargo vet` brings it back to this state. I'm going to leave it as is.

williampMSFT (approved):

PR #916, opened 2026-07-08
power-policy-interface/psu: Introduce notification traits

### LINKED ISSUE #910: Power policy: Introduce notification traits

RobertZ2011:
Refactor the existing power policy PSU notification code to use generic notifier and handler traits instead of channel-based traits. Provide implementations of these traits for existing channel-based code.

### DIFF
--- a/power-policy-interface/src/psu/event.rs
+++ b/power-policy-interface/src/psu/event.rs
@@ -41,9 +41,9 @@ where
     pub event: EventData,
 }
 
-/// New-type that implements the [`crate::psu::notification::Notifier`] trait for any [`NonBlockingSender<Event>`].
+/// New-type that implements the [`crate::psu::notification::Notifier`] trait for any [`NonBlockingSender<EventData>`].
 ///
-/// This allows the user to chose blocking/non-blocking behavior when a type supports both.
+/// This allows the user to choose blocking/non-blocking behavior when a type supports both.
 pub struct NonBlockingSenderNotifier<S: NonBlockingSender<EventData>>(pub S);
 
 impl<S: NonBlockingSender<EventData>> crate::psu::notification::Notifier for NonBlockingSenderNotifier<S> {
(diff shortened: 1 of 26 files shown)


power-policy-service/src/service/mod.rs
```diff
@@ -110,81 +112,60 @@ impl<'device, Reg: Registration<'device>, Customization: customization::Customiz
...
-    }
-
-    async fn process_notify_detach(&mut self, device: &'device Reg::Psu) -> Result<(), Error> {
-        info!("({}): Received notify detached", device.lock().await.name());
-        self.post_provider_removed(device).await;
-        self.update_current_consumer(ConsumerDisconnect::none()).await?;
-        Ok(())
+    async fn notify_unconstrained(&mut self, unconstrained: UnconstrainedState) {
+        for notifier in self.registration.notifiers() {
+            if let Err(e) = notifier.notify_unconstrained(unconstrained).await {
```
jerrysxie, line 117: So depending on whether this notifier is blocking/non-blocking, the behavior would be different completely. This is less explicit than the previous `broadcast_event()` which always uses `try_send()`.

power-policy-interface/src/psu/event.rs
```diff
@@ -35,3 +40,109 @@ where
...
     pub event: EventData,
 }
+
+/// New-type that implements the [`crate::psu::notification::Notifier`] trait for any [`NonBlockingSender<EventData>`].
+///
+/// This allows the user to choose blocking/non-blocking behavior when a type supports both.
+pub struct NonBlockingSenderNotifier<S: NonBlockingSender<EventData>>(pub S);
+
+impl<S: NonBlockingSender<EventData>> crate::psu::notification::Notifier for NonBlockingSenderNotifier<S> {
+    fn notify_attached(&mut self) -> impl Future<Output = Result<(), crate::psu::notification::Error>> {
```
jerrysxie, line 50: Non-blocking: This is quite a bit of duplication, maybe we should consider macros.

jerrysxie (commented):
@RobertZ2011 Can we link the issue that is tracking this effort?

power-policy-service/src/service/mod.rs
RobertZ2011: Ultimately, the user of this code is responsible for providing an implementation that is well behaved, this PR doesn't change that. Previously, a bugged `try_send` implementation could have blocked execution. That risk still exists but having an await point here also makes recovery easier through allowing things like a top-level `with_timeout`.

power-policy-interface/src/psu/event.rs
RobertZ2011: I can look into it, but I'm not sure there would be much benefit to macros here. These are generic over anything channel-like so nothing in this file would need to be re-implemented to support new channel types. If you're not using some sort of channel, then you're implementing custom behavior that likely wouldn't benefit from a macro either.

jerrysxie (approved):

tullom (approved):

PR #920, opened 2026-07-20
type-c-service/max_sink_voltage: Port recovery logic from v1

### LINKED ISSUE #914: type-c: Port 907 to main
Port the functionality in #907 to main and add test coverage.

RobertZ2011:
Pull in the recovery logic from v1 and add test coverage.

### DIFF
--- a/type-c-service/src/controller/event_receiver.rs
+++ b/type-c-service/src/controller/event_receiver.rs
@@ -119,7 +119,7 @@ impl<
     ///
     /// Returns the local port ID and the event bitfield.
     pub async fn wait_event(&mut self) -> Event {
-        let timeout = self.shared_state.lock().await.sink_ready_timeout;
+        let timeout = self.shared_state.lock().await.sink_ready_deadline;
         match select(self.port_event_receiver.wait_next(), async move {
             if let Some(timeout) = timeout {
                 Timer::at(timeout).await;
@@ -133,7 +133,7 @@ impl<
             Either::Second(_) => {
                 let mut status_event = PortStatusEventBitfield::none();
                 status_event.set_sink_ready(true);
-                self.shared_state.lock().await.sink_ready_timeout = None;
+                self.shared_state.lock().await.sink_ready_deadline = None;
                 Event::PortEvent(PortEvent::StatusChanged(status_event))
             }
         }
--- a/type-c-service/src/controller/event_receiver.rs
+++ b/type-c-service/src/controller/event_receiver.rs
@@ -1,7 +1,7 @@
 //! This module contains event receiver types for the controller wrapper.
 use core::array;
 use core::future::pending;
-use embassy_futures::select::{Either, select};
+use embassy_futures::select::{Either3, select3};
 use embassy_time::Timer;
 use embedded_services::error;
 use embedded_services::event::{NonBlockingSender, Receiver};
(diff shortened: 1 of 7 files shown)


type-c-service/tests/power.rs
```diff
@@ -513,6 +517,150 @@ impl Test for TestSinkDisableOnVoltageChange {
...
+        )
+        .await;
+
+        // Power policy service should broadcast a consumer disconnected event
+        match power_policy_result {
+            Ok(PowerPolicyEvent::ConsumerDisconnected(psu, flags)) => {
+                assert_eq!(flags, ConsumerDisconnect::none().with_renegotiation(true));
+                assert!(ptr::eq(psu, port0.port));
+            }
+            _ => panic!("Did not receive provider connected event"),
```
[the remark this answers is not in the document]
RobertZ2011: Done.

type-c-service/src/controller/state.rs
```diff
@@ -4,20 +4,20 @@ use embassy_time::Instant;
 #[derive(Copy, Clone)]
 pub struct SharedState {
     /// Sink ready timeout
```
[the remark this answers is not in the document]
RobertZ2011: Done.

felipebalbi (dismissed):

type-c-service/src/controller/event_receiver.rs
```diff
@@ -119,7 +119,7 @@ impl<
     ///
     /// Returns the local port ID and the event bitfield.
     pub async fn wait_event(&mut self) -> Event {
-        let timeout = self.shared_state.lock().await.sink_ready_timeout;
+        let timeout = self.shared_state.lock().await.sink_ready_deadline;
```
jerrysxie: We are snapshotting the timeout here. If we start with `None` but if the timeout is updated, how does this call gets updated to the new timeout? The test code sets deadline to `Some()` through `set_max_sink_voltage()` before calling `wait_event()`. Is this order guaranteed?

type-c-service/src/controller/max_sink_voltage.rs
```diff
@@ -35,6 +36,18 @@ impl<
...
 
+            // In general it's not possible to know if setting the max sink voltage will trigger a renegotiation
+            // because the logic to select a particular contract is specific to the PD controller.
+            // Enable the sink ready timeout as a recovery mechanism. If there's no renegotiation, then the timeout
+            // will result in us broadcasting the existing contract back to the power policy.
+            {
+                let mut shared_state = self.shared_state.lock().await;
+                if shared_state.sink_ready_deadline.is_none() {
+                    shared_state.sink_ready_deadline =
+                        Some(Instant::now() + Self::check_sink_ready_timeout_duration(self.status.epr));
```
jerrysxie, line 47: Since status is the cached status, when we are switching from SPR to EPR, won't we use the timeout value for SPR instead of EPR?

type-c-service/tests/power.rs
```diff
@@ -513,6 +523,238 @@ impl Test for TestSinkDisableOnVoltageChange {
...
+        match power_policy_result {
+            Ok(PowerPolicyEvent::ConsumerConnected(psu, capability)) => {
+                assert_eq!(
+                    capability,
+                    ConsumerPowerCapability {
+                        capability: POWER_CAPABILITY_5V_1A5,
+                        flags: ConsumerFlags::none().with_psu_type(PsuType::TypeC),
+                    }
+                );
+                assert!(ptr::eq(psu, port0.port));
```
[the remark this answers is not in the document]
RobertZ2011: Done.

type-c-service/src/controller/event_receiver.rs
RobertZ2011: I added logic to handle invalidating the old deadline and added tests as well.

type-c-service/src/controller/max_sink_voltage.rs
RobertZ2011: This value is whether the source is EPR capable in general. It shouldn't change without the port being unplugged.

type-c-service/src/controller/max_sink_voltage.rs
```diff
@@ -35,6 +36,29 @@ impl<
...
+                }
+
+                if self
+                    .loopback_sender
```
jerrysxie, line 51: `sync_state()` also sends event to this channel as well, in the macro, this channel is default to size of 1. Would there be an instance where `sync_state()` and `set_max_sink_voltage()` gets called concurrently and event would be lost?

RobertZ2011: I've set the default channel size to 4. Especially since these functions could be called in the same task as the event processing, it's a good idea to make this buffer a little bigger than those used for communication with the service implementations.

jerrysxie (approved):

type-c-service/src/controller/event_receiver.rs
```diff
@@ -111,30 +72,62 @@ impl<
...
+                let timeout = self.shared_state.lock().await.sink_ready_deadline;
+                match select3(
+                    self.port_event_receiver.wait_next(),
+                    async move {
+                        if let Some(timeout) = timeout {
+                            Timer::at(timeout).await;
+                        } else {
+                            pending::<()>().await;
+                        }
+                    },
```
felipebalbi, line 107: This timeout is to account for port_event or loopback receiver? Could you wrap the target with `WithTimeout` instead?

felipebalbi (approved):
Two minor nits, but looks okay.

type-c-service/src/controller/event_receiver.rs
RobertZ2011: This is its own event and not a timeout on the other two futures. PD controllers don't always notify the MCU when it's safe to draw current. We set a timer that expires after the maximum time given by the PD spec to handle this case.

tullom (approved):

Issue #33, opened 2024-12-12
Hid-service clean-up


RobertZ2011:
* Reading HID commands from I2C currently uses several reads. Investigate using a single large read instead.
* Break HID data structures into own crate (embedded-hid-i2c)
* Refactor serialization to use something like bincode.

PR #926, opened 2026-07-22
Implement HID-I2C target service

### LINKED ISSUE #839: Implement HidI2c service
We need a new HidI2c service that will take an implementation of a HidDevice trait, an I2C slave, and an interrupt line and present that HidDevice on the bus

williampMSFT:
A new service that implements the HID-I2C spec, written against the I2C HAL trait and a new HidDevice trait that is transport-agnostic.

This obsoletes the current hid-service, which is really a HID-I2C service that leverages the comms system / message passing and doesn't handle interrupt line management.

A future change will introduce an analogue to the impl_odp_mctp_relay_handler macro to aggregate multiple logical HID devices into a single HID device.

This should allow clients to write transport-agnostic HID device code that can be reused without modification on other, future physical transports (e.g. I3C), and removes our dependance on the comms system, which imposes lifetime requirements that make testing difficult.

Resolves #839


embedded-service/src/relay/hid.rs
```diff
@@ -0,0 +1,284 @@
...
+    On,
+
+    /// Reduced power state, but a device that sends a report in this state can wake the host - quiesce messages if you don't want to do that
+    Sleep,
+
+    /// The device is not allowed to wake the host. This is not supported on all transports - in particular, I2C will never command a device into the off state.
+    Off,
+}
+
+/// A HID report of no more than X bytes
```
RobertZ2011: nit: this comment mentions X bytes, but it's a slice not an array.

embedded-service/src/relay/hid.rs
```diff
@@ -0,0 +1,284 @@
...
+///
+/// If you're part of an aggregate device created by impl_odp_hid_aggregate_device!, ***this will reset your peers too***.
+/// Therefore, you should be very certain this is the behavior you want before you return an error from any of these functions.
+///
+/// The normal pattern in HID seems to be to either embed an error code in an input report or to drop the message entirely.
+///
+pub trait HidDevice {
+    /// The maximum size of an input report (device -> host) that this device can use, expressed in bytes.
+    /// This must agree with the descriptor returned by `report_descriptor()`.
+    type InputReportMaxSize: ArrayLength;
```
RobertZ2011, line 124: What's the motivation behind making these values associated types over something like const trait functions?

embedded-service/src/relay/hid.rs
```diff
@@ -0,0 +1,284 @@
...
+        report_id: ReportId,
+        process_report: impl AsyncFnOnce(GetHidReport<'_>) -> R,
+    ) -> impl core::future::Future<Output = Result<R, HidError>>;
+
+    /// Respond to a command from the host to handle a particular output/feature report.
+    fn set_report(&mut self, report: &SetHidReport<'_>) -> impl core::future::Future<Output = Result<(), HidError>>;
+
+    /// Blocks until the device is ready to yield an unsolicited input report.
+    /// When this returns, the next call to process_next_input_report should be able to run without blocking on I/O.
+    fn wait_for_input_report(&mut self) -> impl core::future::Future<Output = ()>;
```
RobertZ2011, line 167: We shouldn't have `wait_` and `process_` functions in the same trait. Doing so creates issues around locking and shared access because if `wait_for_input_report` is called then nothing else can access the object or call functions on it. This forces us to have a `HidDevice` implementation that's separate from the actual driver, which makes things more complicated for devices that naturally communicate over HID and don't need a separate struct to implement `HidDevice` for them. `wait_for_input_report` should be implemented on another object. This is basically what you're already doing in the keyboard example with the channel, you would just have to split the channel into its sender and receiver.

hidi2c-target-service/src/constrained_hid_device.rs
```diff
@@ -0,0 +1,43 @@
...
+    T::FeatureReportMaxSize: Max<T::InputReportMaxSize>,
+    T::FeatureReportMaxSize: Max<T::OutputReportMaxSize>,
+    <T::FeatureReportMaxSize as Max<T::InputReportMaxSize>>::Output: ArrayLength,
+    <T::FeatureReportMaxSize as Max<T::OutputReportMaxSize>>::Output: ArrayLength,
+{
+    type MaxInputOrFeatureSize = <T::FeatureReportMaxSize as Max<T::InputReportMaxSize>>::Output;
+    type MaxOutputOrFeatureSize = <T::FeatureReportMaxSize as Max<T::OutputReportMaxSize>>::Output;
+}
+
+impl<T> sealed::Sealed for T where T: ConstrainedHidDevice {}
```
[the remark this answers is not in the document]
williampMSFT: seems to work fine to me

embedded-service/src/relay/hid.rs
```diff
@@ -0,0 +1,284 @@
...
+    Reserved = 3,
+}
+
+impl HidReportDescriptorElementHeader {
+    /// The size of this item in bytes.
+    fn item_size(&self) -> usize {
+        #[allow(clippy::panic)] // This is temporary we get the HID support library implemented
+        if self.0 == 0b11111110 {
+            panic!("Long items are not yet supported"); // TODO implement this once we have the HID support library implemented - see 6.2.2.3 of https://www.usb.org/sites/default/files/hid1_11.pdf
+        }
```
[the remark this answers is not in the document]
williampMSFT: I have a follow-up PR to add support for this and fix it but it makes substantial changes to this type. in the interest of constraining the scope of this PR, punting that to the aggregation PR; the panic will be removed then.

embedded-service/src/relay/hid.rs
```diff
@@ -0,0 +1,284 @@
...
+            let header = HidReportDescriptorElementHeader(*header_bytes);
+            if header.item_type() == HidItemType::Global && header.item_tag() == REPORT_ID_ITEM_TAG {
+                implicit = false;
+                break;
+            }
+
+            if header.item_size() != 0 {
+                iter.nth(header.item_size() - 1); // skip over the data bytes for this item
+            }
+        }
```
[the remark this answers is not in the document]
williampMSFT: This will be fixed in an upcoming PR that adds extended support for manipulating HID descriptors

examples/rt685s-evk/src/mocks/keyboard.rs
```diff
@@ -0,0 +1,269 @@
...
+        embassy_sync::channel::Channel<embedded_services::GlobalRawMutex, KeyboardInputReport, KEYBOARD_CHANNEL_DEPTH>,
+}
+
+impl Default for MockKeyboardService {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl MockKeyboardService {
```
RobertZ2011: nit: This should be named something like `MockKeyboardDevice` or even just `MockKeyboard`.

examples/rt685s-evk/src/mocks/keyboard.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+/// Depth of the keyboard input-report channel.
+const KEYBOARD_CHANNEL_DEPTH: usize = 5;
+
+/// Consumer side of the mock keyboard. Owns the channel that carries input reports and exposes a
+/// method to inject key presses.
+pub struct MockKeyboardService {
+    channel:
+        embassy_sync::channel::Channel<embedded_services::GlobalRawMutex, KeyboardInputReport, KEYBOARD_CHANNEL_DEPTH>,
```
RobertZ2011: This channel shouldn't be owned by the struct. You should create a `Notifier` trait that the struct is generic over and provide an implementation of that trait over the generic sender traits present in embedded-service/src/event.rs, see https://github.com/OpenDevicePartnership/embedded-services/pull/926.

Then you can the declare the channel and pass the sender into this struct and the receiver will end up in the HID-I2C service.

examples/rt685s-evk/src/mocks/keyboard.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+impl MockKeyboardService {
+    pub const fn new() -> Self {
+        Self {
+            channel: embassy_sync::channel::Channel::new(),
+        }
+    }
+
+    pub async fn click_key(&self, key_code: KeyCode) {
```
RobertZ2011: nit: These functions should take `&mut self`. You shouldn't run into any ownership issues once the channel is declared outside of the struct.

examples/rt685s-evk/src/mocks/keyboard.rs
```diff
@@ -0,0 +1,269 @@
...
+        Self {
+            service,
+            report_id: ReportId(1),
+            descriptor: HidReportDescriptor::new(KEYBOARD_HID_REPORT_DESCRIPTOR_WITH_ID)
+                .expect("keyboard HID report descriptor should be valid"),
+        }
+    }
+}
+
+impl embedded_services::relay::hid::HidDevice for MockKeyboardHidRelay<'_> {
```
RobertZ2011: This is probably one of those cases where we want this implementation done in `MockKeyboardService`.

examples/rt685s-evk/src/mocks/mouse.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+}
+
+/// Consumer side of the mock mouse. Owns the channel `Receiver` and hands the host borrows that
+/// point directly into the ring buffer — no intermediate copy of the report payload.
+pub struct MockMouseService<'hw> {
+    receiver: MouseReceiver<'hw>,
+}
+
+impl<'hw> MockMouseService<'hw> {
```
RobertZ2011: Same comments as the keyboard.

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,691 @@
...
+        }
+    }
+}
+
+/// Service runner for the HID-I2C service. You must call run() on the runner to drive the service.
+pub struct Runner<'hw, Bus: I2cTargetAsync, AttnPin: embedded_hal::digital::OutputPin, HidDevice: ConstrainedHidDevice>
+{
+    bus: TimeoutBus<Bus>,
+    attn_pin: AttnPinHandler<AttnPin>,
+    hid_device: HidDevice,
```
RobertZ2011, line 258: This should be generic over a `&embedded_service::sync::Lockable<Inner: HidDevice>` to allow for shared access to the HID device. Devices may implement multiple types of functionalities and need to be accessible to other code, but it also helps with debugging, or even if we want to do something like implementing recovery logic that does a device reset without needing to inject HID messages somewhere.

RobertZ2011 (changes requested):

embedded-service/src/relay/hid.rs
williampMSFT: The relay service (e.g. HID-I2C service) needs to be able to statically allocate a buffer that's large enough to store a report, and that size is max(feature report, input report).  This gets further complicated in the follow-up change that's going to emit an aggregate device, since that's going to need to be the max of all the sub-device's buffer sizes.

Ideally, I'd like these to be associated constants, and in C++ that's pretty trivial; unfortunately, Rust's support for const stuff in generics is pretty lacking right now, which makes this difficult. It turns out that you can't use a constant value from something you're generic over and then use that value in a type's name without the generic_const_exprs unstable feature, i.e. you can't do this:

```rust
trait Device {
    const size: usize;
}

struct DeviceUser<D: Device> {
    buffer: [u8; D::size] // <== this is currently illegal
}
```

because you can't spell `D::size` as part of `buffer`'s type since it comes from a generic.

The generic_array/num_enum crates gets around this by having a type for every integer and defining types that have associated types that let you do this kind of math, and then having arrays that have their sizes specified as one of these integer-value-types rather than as an actual usize.

Ideally when generic_const_exprs stabilizes we'd move to just using associated constants, but for now I think we need this hack, unfortunately.  I'd love an alternative to this if you have one, though :)

hidi2c-target-service/src/service.rs
williampMSFT: I think HidDevice is on the other "side" of the hypothetical lock boundary.  The pattern we've adopted for relaying stuff to the host is that you have a "relay handler" object that's directly owned by the relay service, and that thing talks over some service's interface trait (maybe using a mutex to do so).  There are a couple upsides of splitting this way:

- We can support multiple logical protocols (i.e. MCTP and HID), but you don't have to pay for the state associated with logical protocols that you're not using
- If an OEM wants to do something custom, they can implement e.g. their own battery service but still use the stock battery-service-mctp-handler or whatever

I explicitly do not want the relay handler being shared - the thing that's shareable is the control handle to the service, which would be used to construct a relay handler.  Does that make sense?

edit: I will definitely concede that I did not demonstrate this pattern in the examples, will fix those up

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,691 @@
...
+
+                let (report_type, report_id) = self.get_command_report_header(command_byte).await?;
+                self.hid_device
+                    .process_get_report(report_type.try_into()?, report_id, async |report| {
+                        // Note: per HID spec, the length field needs to include its own length (2 bytes)
+                        let len_header = (report.data().len() as u16 + device_descriptor::HID_REPORT_HEADER_SIZE_BYTES)
+                            .to_le_bytes();
+                        self.bus.write(&len_header).await?;
+                        self.bus.write(report.data()).await?;
+                        Ok::<(), Error<Bus::Error>>(())
```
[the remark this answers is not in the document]
williampMSFT: per hid-i2c spec, the host isn't allowed to do that here

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,691 @@
...
+
+            Opcode::SetReport => {
+                trace!("Processing set report command");
+                let (report_type, report_id) = self.get_command_report_header(command_byte).await?;
+                let mut len_header = [0u8; core::mem::size_of::<u16>()];
+                self.bus.read(&mut len_header).await?;
+
+                // Note: per HID spec, the length field relayed over the wire needs to include its own length (2 bytes)
+                let report_size =
+                    (u16::from_le_bytes(len_header) - device_descriptor::HID_REPORT_HEADER_SIZE_BYTES) as usize;
```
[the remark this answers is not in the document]
williampMSFT: I think using only using it in non-panicking functions (.get rather than operator[]) should solve this already, but will switch for clarity

embedded-service/src/relay/hid.rs
```diff
@@ -0,0 +1,284 @@
...
+use generic_array::ArrayLength;
+use num_enum::TryFromPrimitive;
+
+/// Errors that a HID device operation can fail with.
+///
+/// Reporting failure triggers a device-initiated reset, so callers must handle these errors explicitly
+/// rather than swallowing them.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum HidError {
```
jerrysxie, line 13: Since this is a public enum and we can reasonably assume that more variants will be added in the future, we should mark this as `#[non_exhaustive]`.

hidi2c-target-service/src/attn_pin_handler.rs
```diff
@@ -0,0 +1,45 @@
+use crate::*;
+
+/// Handler for the ATTN pin, which is used to signal the host that we have an input report ready to be read.
+/// This is a simple wrapper around an OutputPin that tracks whether we've asserted the interrupt or not, because
+/// OutputPin doesn't have a built-in way to interrogate its own state.
+///
+pub struct AttnPinHandler<AttnPin: embedded_hal::digital::OutputPin> {
```
jerrysxie: There is a `StatefulOutputPin` where the output pin is stateful, but not all pins supports it. Also some pins that does support that functionality might turn the input buffer to save power.

embedded-service/src/relay/hid.rs
williampMSFT: As initially conceived, the intention was that the HidDevice would be a separate object from the service that it's talking to - this allows us to support multiple logical protocols (e.g. HID and MCTP) without paying for storage to support logical protocols that we're not actually using on a given platform.  HidDevice is a sort of "super-receiver" wrapper.  That said, I'll look into options for splitting this into multiple objects, which might enable the sort of "hid-native" scenario you're describing

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,708 @@
...
+        {
+            // Timed out waiting for the controller to drive the transfer.
+            Err(_timeout_error) => {
+                error!("Read request timeout");
+                self.bus.recover().await.map_err(Error::Bus)?;
+                Err(Error::Protocol(ProtocolError::Timeout))
+            }
+            // Controller finished writing; report how many bytes we drained.
+            Ok(Ok(
+                status @ (WriteStatus::Stopped(bytes) | WriteStatus::Restarted(bytes) | WriteStatus::BufferFull(bytes)),
```
jerrysxie: Shouldn't we call `respond_to_write()` again or else the i2c write will not be terminated from the host?

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,708 @@
...
+}
+
+impl<
+    'hw,
+    Bus: I2cTargetAsync + 'hw,
+    AttnPin: embedded_hal::digital::OutputPin + 'hw,
+    HidDevice: ConstrainedHidDevice + 'hw,
+> Runner<'hw, Bus, AttnPin, HidDevice>
+{
+    async fn process_request(&mut self, request: Request) {
```
jerrysxie, line 333: Do we not want to bubble up any error from here?

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,708 @@
...
+            Err(Error::Device(HidError::TriggerReset)) => {
+                warn!("HID-I2C: HID device requested device-initiated reset");
+                self.reset().await;
+            }
+        }
+    }
+
+    async fn process_register_access(&mut self) -> Result<(), Error<Bus::Error>> {
+        let mut reg = [0u8; 2];
+        let read = self.bus.read(&mut reg).await?;
```
jerrysxie: Reading from the i2c bus in partial read as a slave is very inefficient and delays the i2c transaction completion on the master side. Can we just read the entire buffer from the host and parse the buffer in place instead of doing partial reads? We had an issue on this: https://github.com/OpenDevicePartnership/embedded-services/issues/33, but since we were going to refactor everything. We did not implement it.

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,708 @@
...
+                self.hid_device.set_power_state(power_state.into()).await?;
+                Ok(())
+            }
+
+            Opcode::GetReport => {
+                trace!("Processing get report command");
+
+                let (report_type, report_id) = self.get_command_report_header(command_byte).await?;
+                self.hid_device
+                    .process_get_report(report_type.try_into()?, report_id, async |report| {
```
jerrysxie, line 616: Are we doing any error handling here for invalid report ID?

`If the DEVICE receives a request from a HOST to an invalid Report ID, the DEVICE must respond with a Report of ZERO Length.`

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,708 @@
...
+                        // Note: per HID spec, the length field needs to include its own length (2 bytes)
+                        let len_header = (report.data().len() as u16 + device_descriptor::HID_REPORT_HEADER_SIZE_BYTES)
+                            .to_le_bytes();
+                        self.bus.write(&len_header).await?;
```
jerrysxie: We should buffer up the write and perform a complete i2c write, partial write are very inefficient over i2c from the slave.

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,708 @@
...
+                        Ok::<(), Error<Bus::Error>>(())
+                    })
+                    .await??;
+
+                Ok(())
+            }
+
+            Opcode::SetReport => {
+                trace!("Processing set report command");
+                let (report_type, report_id) = self.get_command_report_header(command_byte).await?;
```
jerrysxie: Are we consuming the 2-byte data register  for this command as well?

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,708 @@
...
+            match event {
+                embassy_futures::select::Either3::First(bus_request) => {
+                    trace!("HID-I2C: Processing request from host");
+                    match bus_request {
+                        Ok(request) => {
+                            self.process_request(request).await;
+                        }
+                        Err(bus_error) => {
+                            error!(
+                                "HID-I2C: Error during bus operation: {:?}",
```
jerrysxie, line 305: Are we planning to do the bus recovering here?

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,708 @@
...
+    async fn run(mut self) -> embedded_services::Never {
+        loop {
+            let event = {
+                // If we've raised the interrupt, we know it won't be dismissed again until it's serviced by the host reading
+                // the input report, so we don't need to listen for another notification.
+                let input_report_ready_future = async {
+                    if self.attn_pin.asserted() {
+                        core::future::pending().await
+                    } else {
+                        self.hid_device.wait_for_input_report().await
```
jerrysxie, line 286: This can be drop, so it has to be cancellation safe?

jerrysxie (changes requested):
I have not reviewed the examples yet. These are my initial thoughts.

hidi2c-target-service/src/service.rs
williampMSFT: Yeah, wait_for_input_report needs to be cancellation-safe, which the naive channel-based implementation is - it doesn't pop anything from the message queue, just waits until the queue is not empty.  I can add a comment in the trait definition to clarify this

hidi2c-target-service/src/service.rs
williampMSFT: I don't think so? I guess we could hoist the error handling out of here and up to run(), but the error handling in there was already getting a bit deep so I figured we could just handle all the error paths for failed requests in here. If you feel strongly that it should be hoisted out, I can do that, though

hidi2c-target-service/src/service.rs
williampMSFT: It's not totally clear to me if we're supposed to call recover() here. The docs only indicate that we have to call it after dropping a respond_to_read or respond_to_write future, but makes no claims about correct behavior when a listen(0 call fails.  @felipebalbi what was the intent for recover() aftter failed listen() calls?

hidi2c-target-service/src/attn_pin_handler.rs
williampMSFT: yeah, I saw that one but it looked like it wasn't implemented on the pin I was using so I went with this. I can add a comment to clarify

hidi2c-target-service/src/attn_pin_handler.rs
```diff
@@ -0,0 +1,45 @@
...
+        result.clear_interrupt();
+        result
+    }
+
+    /// Clear the interrupt, which is done by setting the pin high.
+    pub fn clear_interrupt(&mut self) {
+        trace!("HID-I2C: ATTN: clear interrupt");
+        self.attn_pin
+            .set_high()
+            .unwrap_or_else(|_| error!("HID-I2C: Failed to clear interrupt on attn pin"));
```
kurtjd: While most GPIO drivers I've seen are infallible and likely not going to cause runtime errors in reality, do we still want to return an error to let caller handle it if it's a possibility on their hardware?

Likewise, if we continue in the failure case and set `asserted` false it kinda puts things in a contradictory state,

hidi2c-target-service/src/attn_pin_handler.rs
```diff
@@ -0,0 +1,45 @@
...
+        self.asserted = false;
+    }
+
+    /// Assert the interrupt, which is done by pulling the pin low.
+    pub fn assert_interrupt(&mut self) {
+        trace!("HID-I2C: ATTN: assert interrupt");
+        self.attn_pin
+            .set_low()
+            .unwrap_or_else(|_| error!("HID-I2C: Failed to assert interrupt on attn pin"));
```
kurtjd: Same as above.

hidi2c-target-service/src/attn_pin_handler.rs
felipebalbi: I think that's a really good idea. Also avoids a breaking change later on.

hidi2c-target-service/src/device_descriptor.rs
```diff
@@ -0,0 +1,121 @@
...
+        const HID_I2C_PROTOCOL_VERSION: u16 = 0x0100;
+        Self {
+            w_hid_desc_length: core::mem::size_of::<DeviceDescriptor>() as u16,
+            bcd_version: HID_I2C_PROTOCOL_VERSION,
+            w_report_desc_length: hid_device.report_descriptor().as_bytes().len() as u16,
+            w_report_desc_register: crate::HidI2cRegister::ReportDescriptor as u16,
+            w_input_register: crate::HidI2cRegister::Input.into(),
+            w_max_input_length: HidDevice::InputReportMaxSize::USIZE as u16
+                + HID_REPORT_HEADER_SIZE_BYTES
+                + if hid_device.report_descriptor().report_ids_implicit() {
```
kurtjd: nit: You have this if statement twice, could probably just de-dup this and store it in a variable first.

hidi2c-target-service/src/service.rs
williampMSFT: great catch, thanks! I thought I was exercising this but it looks like the windows mouse and keyboard drivers only ever use the input/output registers rather than using the command register to send input/output reports. I'm a bit surprised that the keyboard driver in particular doesn't read the output report from the keyboard LEDs, but apparently it just clobbers whatever's there.

I have a fix for this and will add some automated tests to catch this class of bug in a follow-up now that the i2c trait is stable

hidi2c-target-service/src/service.rs
williampMSFT: Right now, the knowledge of which report IDs are valid exists in two places: the report descriptor (which we have but don't currently understand) and the HidDevice implementation (which can report failure to us, but is currently only allowed to do so by requesting the triggering of a reset).

Unfortunately, parsing the HID report descriptor to figure out which report IDs are valid and which aren't requires a fair bit of machinery.  We have most of that machinery in the follow-up PR to this to add an aggregation macro, but that's a *lot* of code and this PR was already getting a bit large for my taste, so I split them up.

For now, if an invalid report ID comes in, the HidDevice implementation can still respond with an empty report, but I think you're right that we should handle it here so there's less for implementations to get wrong.

I propose adding a //TODO here for this and circling back once we have the descriptor parsing logic in.  Will do that unless you have objections :)

hidi2c-target-service/src/attn_pin_handler.rs
williampMSFT: We can, although to be clear, this type is not part of our public interface, it's not exported and it's only used by the service so there's not really a caller to make that kind of decision. I'll change the access modifiers to pub(crate) to make that clear without needing to walk the entire import tree.

I moved the unwrap_or_elses into here to avoid duplicating the error message but I can hoist those to service.rs if we want to - thoughts?

kurtjd: Sounds good, yeah makes sense it's not meant to be called directly by the user. Though wonder if it makes sense to still somehow handle the error in a way that user code can respond to. For example, in the thermal service, if a sensor encounters an error reading from the sensor, it broadcasts a `Failure(Error)` event on its broadcast channel, so anyone listening has the chance to try and respond to that.

Might be overkill for this case, but it does feel a bit wrong to ignore a possible hardware failure and let things get into an inconsistent state.

embedded-service/src/relay/hid.rs
williampMSFT: TL;DR: I looked into doing this, but wound up tripping over `reset()` semantics - I don't think we can/should do this, but happy to chat further about options if you feel strongly that this is necessary.

The issue is that the semantics of HID reset are that you're supposed to clear any in-flight messages and reset the piece of the service that's HID-facing to a clean state.  This means that to correctly implement `reset()`, you need to mess with both the 'sender' state (to change settings on the service or whatever) and 'receiver' state (to drain the queue of pending messages).

If we go the `split()` route, this means you end up needing to put all the shared state behind a mutex anyway to implement reset(), and you're back to needing to lock that state to wait for the next report, so you're not really saving anything.  Additionally, you have to introduce an extra layer of external storage (oncelock/staticcell/etc) to accommodate this. However, if we view implementations of HidDevice as "super-receivers" that are owned by the relay service and may be implemented using a handle-to-service (potentially behind a mutex), this becomes trivial.

I think in the "hid-native" case you're talking about, the service's interface (e.g. keyboard service) would do the 'notification' pattern you're talking about, and then we'd construct a HidDevice that contained both a receiver and a reference-to-keyboard-service (potentially through a mutex), and some of that receiver's messages would be "which keys are pressed", which the HidDevice implementation can translate to HID reports.  In the specific case of the keyboard service, the actual notifications may be the same struct that the HidDevice implementation passes on to the host because keyboard is so hid-centric, but you could also e.g. send out a notification when the LED state changes that the HidDevice implementation drops because that's not part of the HID spec.  Does that make sense?

hidi2c-target-service/src/service.rs
williampMSFT: I think this is going to be cheaper to do in some places than others.  In particular, in this specific case I think with the way the i2c trait is currently written, that means we have to memcpy the entire report. For small reports that may be worth it but for large reports it may not - I think we may need to measure?  I'll see if I can take care of this in places where we don't have to do a memcpy to get the benefit though

Alternatively, maybe we could enhance the trait to take a slice of slices rather than a single slice or something to queue up multiple chunks that are coming from disparate locations in memory, or perhaps take an iterator rather than a slice?  Not sure how viable that would be or if it's supportable on all hardware though - @felipebalbi thoughts?

hidi2c-target-service/src/service.rs
felipebalbi: a failed listen didn't really initiate any bus transfer. Controller side should issue a Stop signal, I don't think we need a recover() here.

hidi2c-target-service/src/service.rs
williampMSFT: I chatted with Felipe on this one, and it seems like we have two paths:

1. memcpy to get stuff contiguous in memory.  This is doable, but may be expensive depending on the size of the report and I'd really rather not.
2. Enhance the I2C trait to have 'enhanced' read/write operations that take a slice of slices.  This lets us dodge the copy and shouldn't require too many changes here. Function would be something like
`fn respond_to_read_with_slices(&mut self, bufs: &[&[u8]]) -> Result<ReadStatus, Self::Error>`
and then we can just call

I pretty heavily favor (2) since it dodges the memcpy and we can't know how big the buffers are at relay-service-authoring-time because we allow user-defined types.  It does require changing or augmenting the I2C trait, though, which might be out of scope for this PR?  What do you think about staying with this 'split' approach for now with the plan being to switch to the sliced-IO API when it becomes available?

hidi2c-target-service/src/service.rs
williampMSFT: resolving this one and continuing the discussion in https://github.com/OpenDevicePartnership/embedded-services/pull/926#discussion_r3677104352

williampMSFT: actually, thinking a bit more about this, reads and writes are a different problem because the writes need to come from disjoint memory on the EC but the reads don't; unresolving and will look into this as part of this PR

hidi2c-target-service/src/service.rs
jerrysxie: As long as we create an issue to track this, I am okay punting this.

hidi2c-target-service/src/service.rs
williampMSFT: solved this for reads as of the latest iteration.  Per discussion in the other thread, will punt on solving this for writes until we have support from the I2C trait/HALs for disjoint writes to dodge a memcpy

hidi2c-target-service/src/service.rs
williampMSFT: Filed this task to track fixing this on the write path: https://github.com/OpenDevicePartnership/embedded-services/issues/929, which requires the I2C HAL trait allowing disjoint writes to the bus

Fixed it on the read path as part of the latest iteration because we control the original memory location for the host sending us stuff so we can just hand out slices to the buffer

hidi2c-target-service/src/service.rs
williampMSFT: yeah, I was mimicking what the old trait does but if the host misbehaves this will definitely hang the bus.  This incidentally got a lot easier to deal with when fixing the read path to be single-operation, so fixed as part of the latest iteration.

embedded-service/src/relay/hid.rs
williampMSFT: Per our chat offline, reset() means that the split doesn't buy us much so going to resolve this one as wontfix, but will continue looking into cleaning up the notifications interface between the mock-(mouse|keyboard) and the mock-(mouse|keyboard)-HidDevice implementation

hidi2c-target-service/src/service.rs
williampMSFT: The latest iteration better demonstrates this pattern in the examples - if we wanted to have the thing shareable, we'd pass an implementation of the MouseService/KeyboardService traits that is backed by a mutex over the control handle to the 'real' service, and the MouseServiceHidRelay/MouseServiceKeyboardRelay are not intended to be shared - rather, they're directly owned by the service.

Chatted with Robert offline about this next bit, but for everyone else reviewing, a few things came up in the course of this investigation that are probably worth mentioning
- Embassy's pubsubchannel doesn't support peeking
- As far as I can tell, there is no embassy_sync type with all the following properties:
  - broadcast / 1:many messaging
  - zerocopy support
  - peek support

Unfortunately, this kind of prices us into either
A) eating a copy of each input report (which is what we do in the examples currently), or
B) doing something where the shape of the message is a RefCell<LargePayloadType> or something rather than just a LargePayloadType and using a heapless::arcpool to back it or something, or
C) coming up with our own data structure that does have those properties.

I think long-term it may be worth investing in (C), we'll probably need it eventually - however, I think that may be out of scope for this PR so punting for now and eating the copy.  I believe that the way the HidDevice trait is currently structured, we should be able to introduce (C) without a breaking change to the HidDevice trait.  It would require a breaking change to the service interfaces (e.g. the mock KeyboardService / MouseService traits), but right now we just have mocks of those so there's no actual interface to break yet.  Presumably when we start building an interface that does need the zerocopy semantics, that's when we'll invest in (C).

examples/rt685s-evk/src/mocks/keyboard.rs
williampMSFT: I can do that if you feel strongly about it, but it seems like a departure from the naming convention we have for other things of the same class so I'm not sure I understand why - can you expand on that? Is it because the old examples didn't fit your mental model for what constitutes a service, and if yes, do the new examples? If not, what am I missing?

examples/rt685s-evk/src/mocks/keyboard.rs
williampMSFT: pushed this into a resources object

examples/rt685s-evk/src/mocks/keyboard.rs
williampMSFT: per our chat offline there's some complexity around reset semantics and the cost of handling multiple logical protocols that make this tricker than it sounds - resolving this one as wontfix, please feel free to reactivate if you feel strongly that we need to make some changes here

hidi2c-target-service/src/attn_pin_handler.rs
williampMSFT: hoisted the error, but not really sure how to expose that to users of the hid-i2c service in a useful way? I guess we could create a broadcast channel specifically for this purpose, but I'm not at all clear on what reasonable action could be taken in response to that signal other than printing an error and continuing on, which is what we're already doing?

kurtjd: That's fair, I can't think of a specific way to recover from or handle this. So might not be worth adding a broadcast channel just for this. I think as long as internally the whole system keeps a consistent state and doesn't get into a "bad" state if it fails to assert/clear an interrupt that should be fine.

williampMSFT: ah, yeah, the failure mode if we're unable to assert the interrupt is that the host will be unaware that there are new input reports ready for retrieval and never poll us for them when we have them.  That's all the interrupt line is - a notification to the host that there's a message ready and it should ask for it over I2C.

There's not really a way in the HID-I2C spec to communicate this kind of failure mode to the host - it assumes that the physical transport is functional, and the host only ever attempts to talk to us over I2C if it either wants to send us an output/feature report, it wants to mess with power settings, or we raise the interrupt telling it we have an input report for it.  It doesn't periodically poll over I2C for any sort of unsolicited error handling or anything like that.

Practically, this means that messages will pile up in our message queue until the queue fills up and we'll start dropping messages.  That does seem like a 'bad state', but I don't think we have a path to recovering from it so I'm not currently trying to solve it here

kurtjd: Makes sense, thanks!

hidi2c-target-service/src/service.rs
```diff
@@ -0,0 +1,748 @@
...
+            }
+        }
+    }
+
+    async fn discard_remaining_bytes_from_host(&mut self) -> Result<(), Error<Bus::Error>> {
+        let mut discard_buffer = [0u8; 16];
+        loop {
+            let result = with_timeout(
+                self.timeout_settings.data_read_timeout,
+                self.bus.respond_to_write(&mut discard_buffer),
```
jerrysxie, line 196: If we time out and `respond_to_write()` gets dropped, we probably need to call `recover()` on the bus.

jerrysxie (dismissed):

hidi2c-target-service/src/service.rs
williampMSFT: good catch, thanks!

jerrysxie (approved):

RobertZ2011 (approved):

PR #932, opened 2026-07-31
type-c-interface: Add notification traits

RobertZ2011:
Add notification traits to allow services to decouple from the notion of a channel-like object. This makes our code more flexible and makes it easier to implement certain types of custom functionality.

assisted-by:claude-opus-4.8

### DIFF
--- a/type-c-service/src/service/mod.rs
+++ b/type-c-service/src/service/mod.rs
@@ -1,19 +1,22 @@
 use core::marker::PhantomData;
 use core::ptr;
 
-use embedded_services::event::NonBlockingSender as _;
 use embedded_services::named::Named as _;
 use embedded_services::sync::Lockable;
 use embedded_services::{debug, error, info, trace};
 use embedded_usb_pd::GlobalPortId;
 use embedded_usb_pd::PdError as Error;
+use embedded_usb_pd::ado::Ado;
 use power_policy_interface::service::event::EventData as PowerPolicyEventData;
+use type_c_interface::control::dp::DpStatus;
 use type_c_interface::control::pd::PortStatus;
+use type_c_interface::port::event::VdmData;
+use type_c_interface::port::notification::NotificationHandler;
 use type_c_interface::port::pd::Pd;
-use type_c_interface::service::event::{DebugAccessoryData, EventData, PortEvent, PortEventData};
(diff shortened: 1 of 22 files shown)


type-c-service/src/service/mod.rs
```diff
@@ -75,10 +78,18 @@ impl<'port, Reg: Registration<'port>> Service<'port, Reg> {
...
+        }
+    }
+
+    async fn notify_ucsi_change_indicator(&mut self, port: &'port Reg::Port, port_id: GlobalPortId, notify_opm: bool) {
+        for notifier in self.registration.notifiers() {
+            if let Err(e) = notifier.notify_ucsi_change_indicator(port, port_id, notify_opm).await {
+                error!("Failed to notify UCSI change indicator: {:#?}", e);
             }
         }
     }
```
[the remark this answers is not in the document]
RobertZ2011: This is intentional. Ultimately, it's the responsibility of the application to decide what the appropriate behavior is for their task, and to prevent deadlocks with `with_timeout` or similar.

type-c-service/src/service/mod.rs
```diff
@@ -121,25 +127,22 @@ impl<'port, Reg: Registration<'port>> Service<'port, Reg> {
...
-                    unhandled
-                );
-                Ok(())
-            }
+            PortEventData::Alert(ado) => self.process_notify_alert(event.port, ado).await,
+            PortEventData::Vdm(vdm) => self.process_notify_vdm(event.port, vdm).await,
+            PortEventData::DiscoverModeCompleted => self.process_notify_discover_mode_completed(event.port).await,
+            PortEventData::UsbMuxErrorRecovery => self.process_notify_usb_mux_error_recovery(event.port).await,
+            PortEventData::DpStatusUpdate(status) => self.process_notify_dp_status_update(event.port, status).await,
+            _ => Ok(()),
```
jerrysxie: Do we want to keep the logging here for unhandled events instead of silently dropping them?

jerrysxie (dismissed):

type-c-service/src/service/mod.rs
RobertZ2011: Done.

kurtjd (approved):

jerrysxie (approved):

tullom (approved):

PR #933, opened 2026-08-03
Implement HID report aggregation

### LINKED ISSUE #838: Implement HID support library
We need a HID support library to enable:
- Macro to declare HID reports that yield struct definitions and TryFrom/Into implementations
- Functionality to logically concatenate a collection of N descriptors (will require parsing and renumbering)

### LINKED ISSUE #840: Implement HidDevice aggregator macro
We need a macro that can take a set of concrete HidDevice trait implementation types and yield a new HidDevice trait implementation that contains the top-level collections of each of those as its own top-level collection.

Depends on #838

williampMSFT:
Introduced a new facility for aggregating a collection of HidDevice implementations into a single HidDevice implementation, including introducing and/or renumbering report IDs as necessary: `impl_hid_aggregate_device!`.  This is the HID analogue to the `impl_odp_mctp_relay_handler!` macro.

This required extending the HidDevice trait slightly so that subdevices can express the size of their report descriptor as part of their type, which allows the aggregation macro to correctly size the combined report descriptor buffer.

Additionally, added an example that combines the mock mouse and keyboard, and removed comments in the single-device macros indicating that this functionality was not yet implemented.

Resolves #840.
Resolves #838.

### DIFF
--- a/embedded-service/src/relay/hid.rs
+++ b/embedded-service/src/relay/hid.rs
@@ -225,6 +225,11 @@ pub trait HidDevice {
     fn reset(&mut self) -> impl core::future::Future<Output = ()>;
 }
 
+/// Maximum `Push`/`Pop` global-item nesting depth supported when parsing a report descriptor
+/// (section 6.2.2.7 of the HID 1.11 spec). Descriptors that nest deeper are rejected with
+/// [`HidDescriptorError::PushPopStackOverflow`].
+const MAX_PUSH_POP_STACK_DEPTH: usize = 16;
+
 /// A HID report descriptor
 ///
 #[derive(Debug, Clone, Copy)]
@@ -273,7 +278,7 @@ impl<'a> HidReportDescriptor<'a> {
         };
 
         let mut state = GlobalItemState::default();
-        let mut stack = heapless::Vec::<GlobalItemState, 16>::new();
+        let mut stack = heapless::Vec::<GlobalItemState, MAX_PUSH_POP_STACK_DEPTH>::new();
 
         for item in DescriptorItems::new(bytes) {
             let item = item?;
@@ -431,12 +436,24 @@ impl ShortItemSize {
     }
 }
 
+/// Bit layout of a HID short-item header byte (section 6.2.2.2 of the HID 1.11 spec):
+/// bits `[1:0]` are the data size, bits `[3:2]` are the item type, and bits `[7:4]` are the tag.
+const HEADER_ITEM_TYPE_SHIFT: u8 = 2;
+/// Bit position of the tag field within a short-item header byte (see [`HEADER_ITEM_TYPE_SHIFT`]).
+const HEADER_ITEM_TAG_SHIFT: u8 = 4;
+/// Mask selecting the size or type field once shifted down (both are 2-bit fields).
+const HEADER_ITEM_FIELD_MASK: u8 = 0b11;
+
 struct HidReportDescriptorElementHeader(u8);
 impl HidReportDescriptorElementHeader {
     /// Encode a Global short-item header with the given `tag` and data size (section 6.2.2.2 of the
     /// HID 1.11 spec).
     const fn global(tag: GlobalItemTag, size: ShortItemSize) -> Self {
-        Self(((tag as u8) << 4) | ((HidItemType::Global as u8) << 2) | (size as u8))
+        Self(
+            ((tag as u8) << HEADER_ITEM_TAG_SHIFT)
+                | ((HidItemType::Global as u8) << HEADER_ITEM_TYPE_SHIFT)
+                | (size as u8),
+        )
     }
 
     /// The raw header byte, for writing this item into a descriptor byte stream.
@@ -448,13 +465,13 @@ impl HidReportDescriptorElementHeader {
     fn item_type(&self) -> HidItemType {
         // Panic safety: This can't actually panic because we mask to 2 bits and the enum covers all 2-bit values, but there's no way to express that in the type system
         #[allow(clippy::expect_used)]
-        HidItemType::try_from_primitive((self.0 >> 2) & 0b11)
+        HidItemType::try_from_primitive((self.0 >> HEADER_ITEM_TYPE_SHIFT) & HEADER_ITEM_FIELD_MASK)
             .expect("HidItemType::try_from_primitive should never fail because we mask to 2 bits")
     }
 
     /// The tag of this item, which is a 4-bit value that identifies the specific item within its type (e.g. start collection, end collection, input, output, etc)
     fn item_tag(&self) -> u8 {
-        self.0 >> 4
+        self.0 >> HEADER_ITEM_TAG_SHIFT
     }
 
     /// Whether this is a "long item" header (see section 6.2.2.3 of the HID 1.11 spec). A long item
@@ -468,7 +485,7 @@ impl HidReportDescriptorElementHeader {
     fn short_item_data_size(&self) -> ShortItemSize {
         // Panic safety: This can't actually panic because we mask to 2 bits and the enum covers all 2-bit values, but there's no way to express that in the type system
         #[allow(clippy::expect_used)]
-        ShortItemSize::try_from_primitive(self.0 & 0b11).expect(
+        ShortItemSize::try_from_primitive(self.0 & HEADER_ITEM_FIELD_MASK).expect(
             "HidReportDescriptorElementHeader::short_item_data_size should never fail because we mask to 2 bits",
         )
     }
@@ -491,6 +508,12 @@ const COLLECTION_ITEM_TAG: u8 = 0b1010;
 
 /// Header byte value that introduces a long item (section 6.2.2.3 of the HID 1.11 spec).
 const LONG_ITEM_HEADER: u8 = 0b1111_1110;
+
+/// Length, in bytes, of a short-item header (`[header]`).
+const SHORT_ITEM_HEADER_LEN: usize = 1;
+
+/// Length, in bytes, of a long-item header (`[0b1111_1110, bDataSize, bLongItemTag]`, section 6.2.2.3).
+const LONG_ITEM_HEADER_LEN: usize = 3;
 
 /// Encoded header for a one-byte Global "Report ID" item (size=1).
 const REPORT_ID_HEADER_SIZE1: HidReportDescriptorElementHeader =
@@ -600,9 +623,9 @@ impl<'a> Iterator for DescriptorItems<'a> {
                 self.done = true;
                 return Some(Err(HidDescriptorError::TruncatedItem));
             };
-            (3, data_size as usize)
+            (LONG_ITEM_HEADER_LEN, data_size as usize)
         } else {
-            (1, header.short_item_data_size().data_bytes())
+            (SHORT_ITEM_HEADER_LEN, header.short_item_data_size().data_bytes())
         };
 
         let data_start = self.pos + header_len;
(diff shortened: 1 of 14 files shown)


embedded-service/src/relay/hid.rs
```diff
@@ -246,41 +464,1497 @@ impl HidReportDescriptorElementHeader {
...
+                        writer.push(inserted_id)?;
+                        report_id_inserted = true;
+                    }
+                }
             }
 
-            if header.item_size() != 0 {
-                iter.nth(header.item_size() - 1); // skip over the data bytes for this item
+            writer.push(POP_HEADER.byte())?;
+        }
```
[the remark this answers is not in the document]
williampMSFT: I believe a descriptor must have at least one top-level Collection to be valid, so this can't happen

embedded-service/src/relay/hid.rs
```diff
@@ -195,21 +220,172 @@ pub trait HidDevice {
...
+    /// Largest input report payload, in bytes.
+    pub input: usize,
+    /// Largest output report payload, in bytes.
+    pub output: usize,
+    /// Largest feature report payload, in bytes.
+    pub feature: usize,
+}
+
+/// Reads a little-endian unsigned value from a short item's data payload (0..=4 bytes).
+fn read_item_value(data: &[u8]) -> u32 {
```
RobertZ2011: This is basically `u32::from_le_bytes`, but you wouldn't be able to detect a `data` value that is too short and you'd end up with an invalid `u32`.

embedded-service/src/relay/hid.rs
```diff
@@ -218,25 +394,67 @@ enum HidItemType {
...
         }
     }
+}
+
+struct HidReportDescriptorElementHeader(u8);
+impl HidReportDescriptorElementHeader {
+    /// Encode a Global short-item header with the given `tag` and data size (section 6.2.2.2 of the
+    /// HID 1.11 spec).
+    const fn global(tag: GlobalItemTag, size: ShortItemSize) -> Self {
+        Self(((tag as u8) << 4) | ((HidItemType::Global as u8) << 2) | (size as u8))
```
RobertZ2011: Can you do a global pass and introduce constants like `GLOBAL_TAG_BIT_LEN` for 4?

embedded-service/src/relay/hid.rs
```diff
@@ -218,25 +394,67 @@ enum HidItemType {
...
+    }
+
+    /// The raw header byte, for writing this item into a descriptor byte stream.
+    const fn byte(&self) -> u8 {
+        self.0
+    }
 
     /// The type of the item, which is one of Main, Global, Local, or Reserved.
     fn item_type(&self) -> HidItemType {
-        // This can't actually panic because we mask to 2 bits, but there's no way to express that in the type system
```
RobertZ2011, line 239: You should leave this comment in and add Panic Safety: to the beginning.

embedded-service/src/relay/hid.rs
```diff
@@ -246,41 +464,1497 @@ impl HidReportDescriptorElementHeader {
...
+    /// Whether this is a "long item" header (see section 6.2.2.3 of the HID 1.11 spec). A long item
+    /// is encoded as `[0xFE, bDataSize, bLongItemTag, data..]`, i.e. a 3-byte header followed by
+    /// `bDataSize` data bytes.
+    fn is_long_item(&self) -> bool {
+        self.0 == LONG_ITEM_HEADER
+    }
+
+    /// The number of data bytes encoded by a *short* item header. Only meaningful for short items.
+    fn short_item_data_size(&self) -> ShortItemSize {
+        #[allow(clippy::expect_used)]
```
RobertZ2011, line 487: Add a panic safety comment.

embedded-service/src/relay/hid.rs
```diff
@@ -246,41 +464,1497 @@ impl HidReportDescriptorElementHeader {
...
-            let header = HidReportDescriptorElementHeader(*header_bytes);
-            if header.item_type() == HidItemType::Global && header.item_tag() == REPORT_ID_ITEM_TAG {
-                implicit = false;
-                break;
+/// Main item tag for a Collection item (section 6.2.2.4). Not part of [`MainItemTag`], which only
+/// covers the report-data Main items.
+const COLLECTION_ITEM_TAG: u8 = 0b1010;
+
+/// Header byte value that introduces a long item (section 6.2.2.3 of the HID 1.11 spec).
+const LONG_ITEM_HEADER: u8 = 0b1111_1110;
```
RobertZ2011, line 510: nit: This constant is binary, but the comment for `is_long_item` uses the hex encoding. Not sure if we want to standardize on one over the other or only reserve hex for constants that fill entire bytes.

embedded-service/src/relay/hid.rs
```diff
@@ -246,41 +464,1497 @@ impl HidReportDescriptorElementHeader {
...
+    /// A Report ID item did not use the expected single-byte encoding (it carried zero, two, or four
+    /// data bytes). Report IDs are single-byte values, so any other encoding is treated as invalid.
+    UnsupportedReportId,
+
+    /// The combined descriptor requires more distinct report IDs than the configured capacity `N`,
+    /// or more than the 255 distinct report IDs the HID specification can represent.
+    TooManyReportIds,
+
+    /// The caller-provided output buffer was too small to hold the combined descriptor.
+    OutputBufferTooSmall,
```
RobertZ2011: Could we add desired/provided values to this enum variant? Same comment on `RemapBufferTooSmall`.

embedded-service/src/relay/hid.rs
```diff
@@ -246,41 +464,1497 @@ impl HidReportDescriptorElementHeader {
...
+    /// HID global items persist across the descriptor byte stream, so naively concatenating
+    /// descriptors would let one input's leftover global state leak into the next and corrupt its
+    /// reports. To prevent this, each input descriptor's items are wrapped in a `Push`/`Pop` pair
+    /// (HID 1.11 section 6.2.2.7). The combined descriptor is therefore larger than the sum of its
+    /// inputs by up to [`MAX_DESCRIPTOR_FRAMING_OVERHEAD`] bytes per input; the caller must size `out`
+    /// to account for this.
+    ///
+    /// Returns the combined descriptor (which borrows `out`) alongside the map.
+    ///
+    pub fn combine<'buf>(
```
RobertZ2011, line 781: Is it possible to make this function const and/or remove the const generics? It sounds like this function would generally be called at compile time to aggregate everything together. This could also contribute to code bloat because it seems like the Rust compiler isn't capable of deduplicating code that use const generics so we'd end up with monomorphized versions for each combination of `REPORT_COUNT` and `DEVICE_COUNT`. I think there's a couple of places where this could apply.

RobertZ2011 (changes requested):

embedded-service/src/relay/hid.rs
williampMSFT: Eventually I'd like to have a const variant, but I think the first use case we have for this is the passthrough case where we need to aggregate a dynamically-enumerated child device (e.g. touchpad) with other statically-enumerated devices (e.g. battery, time-alarm), and unfortunately in that case we can't know the passthrough device's report descriptor at compile time (or at least, our first client is unwilling to commit to knowing the passthrough device's report descriptor at compile time right now).

Additionally, const functions are currently pretty limited - notably you can't use trait methods in const functions right now, which locks you out of iterators/foreach loops / operator?, and slice::get() is conditionally-const so you can't use that either / have to use indexing, which can panic at runtime.

There's a tracking issue for stabilizing const traits, though, and [the most recent comment](https://github.com/rust-lang/rust-project-goals/issues/106#issuecomment-4737122434) seems to indicate they're hoping to stabilize in the next year or so, so I was thinking we might punt on doing this in a const way for now since it's not usable for our initial use case anyway and hopefully by the time we'd get value out of a const implementation, the const traits language feature will have stabilized and we'll be able to leverage those without needing to fall back to C-style loops/error handling and panicking operations.  Does that sound reasonable?  I think the alternative might be two parallel implementations to satisfy both the panic-safety constraints and the const limitation constraints, which I'd really like to avoid if possible

edit: also, re: monomorphization cost, I anticipate that in typical usage there will be exactly one monomorphization of this on any system that leverages it, but if we find that people are doing two, I think we can tweak the implementation a bit to use a separate view type or something that talks in terms of slices rather than arrays and have most of the code in there operating on slices? would have to play with it for a bit but I think it's a potential fallback if we do end up needing more than one of these

embedded-service/src/relay/hid.rs
williampMSFT: Yeah, that's kinda the idea, except you don't end up with an invalid u32 - this is a design feature of HID's descriptor format (they were trying to save as many bits as possible).

HID report descriptors are a series of "items", which are sort of like variable-length instructions to a state machine.  The format for "short items" is a one-byte header with the format
```
____ __ __
         ^--- size: number of data bits to follow.  0=0, 1=1, 2=2, 3=4 bytes. 
      ^------ type: "main", "global", "local", "reserved"
^------------ tag: what kind of instruction (e.g. a "global" instruction with tag 0b0111 is "report size")
```
followed by an optional data payload (size=0 if there is no data payload).

This function is supposed to take any of the possible slice sizes and interpret the payload as a u32, regardless of if the value is 1/2/4 bytes.  I can add some comments to try make this a bit clearer

williampMSFT: moved to a member function on DescriptorItem to make intended use clearer / harder to use on things that don't have these semantics

embedded-service/src/relay/hid.rs
williampMSFT: yeah, I went mostly with the binary representation because the [hid spec](https://www.usb.org/sites/default/files/hid1_11.pdf) mostly uses binary representation since they're doing a lot of bit-packing / partial bytes; I thought it'd be easier for someone who had both pulled up to see that they matched, but I think I copied the actual long item header from a report descriptor I was referencing, which are typically formatted as hex with comments (there are a bunch of HID report descriptor parsing tools that all seem to have standardized on mostly the same format)

I'll change the hex comments to match; thanks!

embedded-service/src/relay/hid.rs
williampMSFT: sounds good - wasn't sure if it was redundant with the expect() but that seems like a reasonable convention

embedded-service/src/relay/hid.rs
williampMSFT: done for outputbuffertoosmall; on further review, remapbuffertoosmall was redundant with the toomanyreportids error, so consolidated those

RobertZ2011 (dismissed):

kurtjd (approved):

RobertZ2011 (approved):

embedded-service/src/relay/hid.rs
williampMSFT: Filed #934 to track circling back on this once const traits land

embedded-service/src/relay/hid.rs
```diff
@@ -218,69 +390,1609 @@ enum HidItemType {
...
+                }
+            }
+
+            async fn wait_for_input_report(&mut self) {
+                // Drop safety: `wait_for_input_report` only peeks readiness and returns nothing, so
+                // dropping the futures for the children that didn't win the race cannot lose a
+                // pending report — the winning child's report stays queued until it is drained by
+                // `process_next_input_report`.
+                let _ = $crate::_macro_internal::embassy_futures::select::select_array(
+                    self.children.each_mut().map(|child| child.wait_for_input_report()),
```
jerrysxie, line 1276: If we do have an unaccounted report ID from a child, passing it through might cause a malformed HID report. If we don't want to panic, a malformed HID report might cause the OS to kill this HID driver. I guess that is acceptable.

PR #936, opened 2026-08-05
hid: fix bug in report parsing when reports are not declared contiguously

williampMSFT:
#933 had a bug: it was not correctly handling the case where a report ID gets 'revisited', but it turns out it's legal to go back to an earlier report ID and keep adding to it, e.g.
- Report ID = 1
- declare 2 bytes
- Report ID = 2
- declare 3 bytes
- Report ID = 1
- declare 2 bytes

means that report ID 1 is a 4-byte report.  This changes our parsing logic to account for this; we have to track each individual report instead of just the largest observed so far.

### DIFF
--- a/embedded-service/src/relay/hid.rs
+++ b/embedded-service/src/relay/hid.rs
@@ -268,14 +268,9 @@ impl<'a> HidReportDescriptor<'a> {
 
         let mut report_ids_implicit = true;
 
-        let mut current = ReportBits::default();
-        let mut max = ReportBits::default();
-        let mut flush = |current: &mut ReportBits| {
-            max.input = max.input.max(current.input);
-            max.output = max.output.max(current.output);
-            max.feature = max.feature.max(current.feature);
-            *current = ReportBits::default();
-        };
+        // Report fields for an ID need not be contiguous in the descriptor, so we have to track
+        // each report ID's accumulated size until the entire thing has been parsed.
+        let mut reports = [ReportBits::default(); u8::MAX as usize + 1];
 
         let mut state = GlobalItemState::default();
         let mut stack = heapless::Vec::<GlobalItemState, MAX_PUSH_POP_STACK_DEPTH>::new();
@@ -299,26 +294,20 @@ impl<'a> HidReportDescriptor<'a> {
                             return Err(HidDescriptorError::UnsupportedReportId);
                         };
                         report_ids_implicit = false;
-                        // Switching report IDs ends the current report and starts a new one.
-                        if id != state.report_id {
-                            flush(&mut current);
-                        }
                         state.report_id = id;
                     } else if tag == GlobalItemTag::Push as u8 {
                         stack
                             .push(state)
                             .map_err(|_| HidDescriptorError::PushPopStackOverflow)?;
                     } else if tag == GlobalItemTag::Pop as u8 {
-                        let restored = stack.pop().ok_or(HidDescriptorError::UnbalancedPop)?;
-                        // A Pop that restores a different report ID ends the current report.
-                        if restored.report_id != state.report_id {
-                            flush(&mut current);
-                        }
-                        state = restored;
+                        state = stack.pop().ok_or(HidDescriptorError::UnbalancedPop)?;
                     }
                 }
                 HidItemType::Main => {
                     let bits = state.report_size_bits.saturating_mul(state.report_count);
+                    let current = reports
+                        .get_mut(state.report_id as usize)
+                        .ok_or(HidDescriptorError::UnsupportedReportId)?;
                     let field = match MainItemTag::try_from_primitive(item.header.item_tag()) {
                         Ok(MainItemTag::Input) => Some(&mut current.input),
                         Ok(MainItemTag::Output) => Some(&mut current.output),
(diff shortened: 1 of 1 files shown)


kurtjd (approved):

embedded-service/src/relay/hid.rs
```diff
@@ -333,7 +322,13 @@ impl<'a> HidReportDescriptor<'a> {
...
             }
         }
-        flush(&mut current);
+
+        let mut max = ReportBits::default();
+        for report in reports {
+            max.input = max.input.max(report.input);
+            max.output = max.output.max(report.output);
+            max.feature = max.feature.max(report.feature);
+        }
```
felipebalbi, line 331: Isn't there a way to write this using max_by() or max_by_key?

embedded-service/src/relay/hid.rs
```diff
@@ -268,14 +268,9 @@ impl<'a> HidReportDescriptor<'a> {
...
-        let mut max = ReportBits::default();
-        let mut flush = |current: &mut ReportBits| {
-            max.input = max.input.max(current.input);
-            max.output = max.output.max(current.output);
-            max.feature = max.feature.max(current.feature);
-            *current = ReportBits::default();
-        };
+        // Report fields for an ID need not be contiguous in the descriptor, so we have to track
+        // each report ID's accumulated size until the entire thing has been parsed.
+        let mut reports = [ReportBits::default(); u8::MAX as usize + 1];
```
jerrysxie, line 273: So 256 * 12 = 3072 bytes. That is a lot of stack space. This works for now, but we might want to optimize with with intrusive list or something in the future.

embedded-service/src/relay/hid.rs
```diff
@@ -1739,6 +1734,52 @@ mod tests {
...
+            0x95, 0x03,       //   Report Count (3)
+            0x81, 0x02,       //   Input (3 bytes into report 2)
+            0x85, 0x01,       //   Report ID (1)
+            0x95, 0x02,       //   Report Count (2)
+            0x81, 0x02,       //   Input (+2 bytes => 4 bytes total in report 1)
+            0xC0,             // End Collection
+        ];
+        let sizes = HidReportDescriptor::new(&bytes).unwrap().max_report_sizes();
+        assert_eq!(sizes.input, 4);
+    }
```
felipebalbi, line 1758: Any chance you can find an invariant and convert this into a proptest?

williampMSFT: I'm not sure I understand the question, what do you mean?

jerrysxie (approved):

embedded-service/src/relay/hid.rs
felipebalbi: Instead of having a single case tested, add proptest as a dev dependency, then have proptest generate report bytes which you can use to check some known invariant.

embedded-service/src/relay/hid.rs
williampMSFT: Yeah, it's not great, which is why I was _really_ hoping we wouldn't have to do this, lol.  I think we do only really need like 19 bits out of the u32 (the max size of a report over hid-i2c in bytes is u16::max, but we have to count in bits) so if we need to shave stack space we might be able to do something tricksy there, or parse it three times and only look at one report type at a time or something so we pay 3x computation for 3x stack memory or something?  In typical usage this'll only be called once in a boot session, pretty early in the startup path, though, so I figured we could punt on that kind of optimization for now (or until we have a device that that overflows)

embedded-service/src/relay/hid.rs
williampMSFT: yes, but then you end up iterating over the same list three times, which I think might be worse for cache locality and whatnot? I can change it if you feel strongly about it

felipebalbi: yes and no. You can use a fold() that just tracks all three numbers. It will iterate only once. But it's nit-picking. Feel free to *not* do it.

embedded-service/src/relay/hid.rs
williampMSFT: It might be possible but that seems like a lot of complexity and it's not clear to me that there's a payoff for that investment?  This is a report descriptor, it basically describes a list of bit-packed struct definitions that haves names that are just `u8`s. We might be able to write something that generates arbitrary struct quantities/shapes/sizes, but that's a pretty similar problem to "generate arbitrary well-formed C headers and then tell me which struct in them is the biggest", I think?  Do we think we need that to solve this problem?  The question these tests is trying to ask is "are we actually treating disjoint struct definitions as a single struct definition, or are we treating them as two independent structs"?

RobertZ2011 (approved):

embedded-service/src/relay/hid.rs
felipebalbi: A single test case gives you confidence on the parser for that one test case. Proptest generates (by default) 256 different random inputs which might be able to trigger corner cases in the parser.

williampMSFT: filed #937 to track this, but I think this is out of scope for this PR

PR #940, opened 2026-08-07
power-policy-service: Introduce `DisconnectReason` enum

RobertZ2011:
Combine existing disconnect reasons into this enum and distinguishbetween renegotiation types. Also rename `ConsumerDisconnect` to `Disconnect` since this type is already used for provider disconnect flags as well.

### DIFF
--- a/embedded-service/src/power/policy/flags.rs
+++ b/embedded-service/src/power/policy/flags.rs
@@ -131,63 +131,73 @@ bitfield! {
     /// Flags for disconnect events
     #[derive(Copy, Clone, PartialEq, Eq)]
     #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-    struct ConsumerDisconnectRaw(u32);
+    struct DisconnectRaw(u32);
     impl Debug;
-    /// Renegotiation
-    ///
-    /// When set this flag indicates that the current consumer is attempting to negotiate a new power capability.
-    pub bool, renegotiation, set_renegotiation: 0;
-    /// Switching
-    ///
-    /// When set this flag indicates that the service is switching to a different PSU.
-    pub bool, switching, set_switching: 1;
+    /// Disconnect reason
+    pub u8, reason, set_reason: 3, 0;
 }
 
-/// Type safe wrapper for consumer disconnect flags
+/// Disconnect reason
+#[derive(Copy, Clone, Debug, PartialEq, Eq, IntoPrimitive, TryFromPrimitive)]
+#[num_enum(error_type(name = InvalidDisconnectReason, constructor = InvalidDisconnectReason))]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(u8)]
+#[non_exhaustive]
+pub enum DisconnectReason {
+    /// Unknown/Unspecified
+    Unknown,
+    /// Switching to a different PSU
+    Switching,
+    /// Renegotiation triggered by device
+    AutoRenegotiation,
+    /// Renegotiation triggered by code
+    ManualRenegotiation,
+}
+
+impl DisconnectReason {
+    /// Check if the reason is a renegotiation
+    pub fn is_renegotiation(&self) -> bool {
+        matches!(
+            self,
+            DisconnectReason::AutoRenegotiation | DisconnectReason::ManualRenegotiation
+        )
+    }
+}
+
+/// Conversion error for [`PsuType`]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub struct ConsumerDisconnect(ConsumerDisconnectRaw);
+pub struct InvalidDisconnectReason(pub u8);
 
-impl ConsumerDisconnect {
-    /// Create new consumer disconnect flags with no flags set
+/// Type safe wrapper for disconnect flags
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Disconnect(DisconnectRaw);
+
+impl Disconnect {
+    /// Create new disconnect flags with no flags set
     pub const fn none() -> Self {
-        Self(ConsumerDisconnectRaw(0))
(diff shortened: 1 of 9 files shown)


benonivainqueur (approved):

kurtjd (dismissed):

embedded-service/src/power/policy/flags.rs
```diff
@@ -131,63 +131,73 @@ bitfield! {
...
+pub struct InvalidDisconnectReason(pub u8);
 
-    /// Set the value of the renegotiation flag
-    pub fn set_renegotiation(&mut self, value: bool) {
-        self.0.set_renegotiation(value);
-    }
+/// Type safe wrapper for disconnect flags
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Disconnect(DisconnectRaw);
```
asasine, line 181: Can you remind me why this is using a bitfield to pack instead of a normal struct? Is it serialized onto a wire and needs a specific packed layout? I don't see serialization methods or impls so I don't think that's the case.

embedded-service/src/power/policy/flags.rs
```diff
@@ -131,63 +131,73 @@ bitfield! {
...
+/// Disconnect reason
+#[derive(Copy, Clone, Debug, PartialEq, Eq, IntoPrimitive, TryFromPrimitive)]
+#[num_enum(error_type(name = InvalidDisconnectReason, constructor = InvalidDisconnectReason))]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub struct ConsumerDisconnect(ConsumerDisconnectRaw);
+#[repr(u8)]
+#[non_exhaustive]
+pub enum DisconnectReason {
+    /// Unknown/Unspecified
+    Unknown,
```
asasine: What code paths don't specify a reason? Can we close those out and remove this smell?

embedded-service/src/power/policy/flags.rs
```diff
@@ -131,63 +131,73 @@ bitfield! {
...
 
-/// Type safe wrapper for consumer disconnect flags
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+/// Disconnect reason
+#[derive(Copy, Clone, Debug, PartialEq, Eq, IntoPrimitive, TryFromPrimitive)]
+#[num_enum(error_type(name = InvalidDisconnectReason, constructor = InvalidDisconnectReason))]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub struct ConsumerDisconnect(ConsumerDisconnectRaw);
+#[repr(u8)]
+#[non_exhaustive]
```
asasine, line 145: Why? How is this intended to be used and therefore low are callers supposed to handle the catch-all case?

embedded-service/src/power/policy/flags.rs
```diff
@@ -131,63 +131,73 @@ bitfield! {
...
 }
 
-/// Type safe wrapper for consumer disconnect flags
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+/// Disconnect reason
+#[derive(Copy, Clone, Debug, PartialEq, Eq, IntoPrimitive, TryFromPrimitive)]
+#[num_enum(error_type(name = InvalidDisconnectReason, constructor = InvalidDisconnectReason))]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub struct ConsumerDisconnect(ConsumerDisconnectRaw);
+#[repr(u8)]
```
asasine, line 144: This becomes part is the public API, which means reordering becomes a breaking change. My suggestion: don't include this unless you specifically want that. If you are only using it to cast for bitpacking, all of with seems to be private to this module, a private method with exhaustive match is more appropriate.

asasine (dismissed):

embedded-service/src/power/policy/flags.rs
RobertZ2011: This was originally to save space under the assumption of eventually using up the entire u32. It isn't really needed because the difference between copying 4 bytes and 32 bytes is unlikely to ever be an issue. But this is something that should be addressed on the main branch, fixing up the other structs in this file as well.

embedded-service/src/power/policy/flags.rs
RobertZ2011: I've added a few more variants and renamed this to `NoLongerCapable`, which I think serves as a better generic reason. We can't assume that all possible reasons for a disconnect will map cleanly to one of these enum variants so it's useful to provide one that provides just the basic "it disconnected".

embedded-service/src/power/policy/flags.rs
RobertZ2011: It would generally be a no-op. Using these reasons means there's specific logic you want to execute in addition to your generic consumer disconnected logic (notify the host, etc). A new variant is only relevant to those that care about it (either for a bug fix or a new feature). I don't see a reason to break the build for everyone else.

embedded-service/src/power/policy/flags.rs
RobertZ2011: I'm going to leave this as is for now. You're right that changing the order would be a breaking change, but there's currently test coverage that would catch this. This will get removed in main as part of a general refactoring of this file.

kurtjd (dismissed):

tullom (dismissed):

embedded-service/src/power/policy/flags.rs
```diff
@@ -131,63 +131,79 @@ bitfield! {
...
+        )
     }
+}
 
-    /// Builder method to set the renegotiation flag
-    pub fn with_renegotiation(mut self, value: bool) -> Self {
-        self.set_renegotiation(value);
-        self
-    }
+/// Conversion error for [`PsuType`]
```
jerrysxie: Comment needs to be updated.

jerrysxie (dismissed):

embedded-service/src/power/policy/flags.rs
RobertZ2011: Done.

tullom (approved):

jerrysxie (approved):

PR #941, opened 2026-08-07
type-c-service/tps6699x: Fix EPR capable flag not being set

RobertZ2011:
This flag was not being set properly. Also add two more useful fields to `PortStatus`.

### DIFF
--- a/embedded-service/src/type_c/controller.rs
+++ b/embedded-service/src/type_c/controller.rs
@@ -42,8 +42,12 @@ pub struct PortStatus {
     pub plug_orientation: PlugOrientation,
     /// power role
     pub power_role: PowerRole,
+    /// Port partner supports dual-data roles
+    pub dual_data: bool,
     /// data role
     pub data_role: DataRole,
+    /// Port partner is USB comms capable
+    pub usb_comms_capable: bool,
     /// Active alt-modes
     pub alt_mode: AltMode,
     /// Power path status
@@ -65,7 +69,9 @@ impl PortStatus {
             dual_power: false,
             plug_orientation: PlugOrientation::CC1,
             power_role: PowerRole::Sink,
+            dual_data: false,
             data_role: DataRole::Dfp,
+            usb_comms_capable: false,
             alt_mode: AltMode::none(),
             power_path: PowerPathStatus::none(),
             epr: false,
--- a/embedded-service/src/type_c/controller.rs
+++ b/embedded-service/src/type_c/controller.rs
@@ -48,7 +48,7 @@ pub struct PortStatus {
     pub alt_mode: AltMode,
     /// Power path status
     pub power_path: PowerPathStatus,
-    /// EPR mode active
+    /// Port partner is EPR capable
     pub epr: bool,
     /// Port partner is unconstrained
     pub unconstrained_power: bool,
--- a/embedded-service/src/type_c/controller.rs
+++ b/embedded-service/src/type_c/controller.rs
@@ -5,9 +5,11 @@ use core::sync::atomic::{AtomicBool, Ordering};
 
 use embassy_sync::signal::Signal;
 use embassy_time::{Duration, with_timeout};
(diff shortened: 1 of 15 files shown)


type-c-service/src/driver/tps6699x.rs
```diff
@@ -366,13 +366,19 @@ impl<M: RawMutex, B: I2c> Controller for Tps6699x<'_, M, B> {
...
                     let pdo = sink::Pdo::try_from(pdo_raw).map_err(|_| Error::from(PdError::InvalidParams))?;
                     let rdo = Rdo::for_pdo(rdo_raw, pdo).ok_or(Error::Pd(PdError::InvalidParams))?;
                     debug!("PDO: {:#?}", pdo);
                     debug!("RDO: {:#?}", rdo);
                     port_status.available_sink_contract = Contract::from_sink(pdo, rdo).try_into().ok();
-                    port_status.dual_power = source_pdos[0].dual_role_power();
-                    port_status.unconstrained_power = source_pdos[0].unconstrained_power();
+                    port_status.dual_power = fixed_source_pdo.dual_role_power;
+                    port_status.unconstrained_power = fixed_source_pdo.unconstrained_power;
+                    port_status.epr = fixed_source_pdo.epr_capable;
```
asasine: Use an exhaustive pattern to not forget fields by accident.

```rust
let Pdo { a, b, c } = pdo; // note: no `..`. Use `field: _` to intentionally discard one.
```

asasine (dismissed):

kurtjd (dismissed):

type-c-service/src/driver/tps6699x.rs
RobertZ2011: Done, I've also added `dual_data` and `usb_comms_capable` to `PortStatus` to support more functionality for downstream users. The other ignored fields are either irrelevant (current/voltage is already pulled from the PDO), or doesn't strike me as particularly useful.

embedded-service/src/type_c/controller.rs
```diff
@@ -48,7 +48,7 @@ pub struct PortStatus {
     pub alt_mode: AltMode,
     /// Power path status
     pub power_path: PowerPathStatus,
-    /// EPR mode active
+    /// Port partner is EPR capable
     pub epr: bool,
```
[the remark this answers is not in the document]
RobertZ2011: Renamed `is_epr` to `epr_capable` to match.

benonivainqueur (approved):

type-c-service/src/driver/tps6699x.rs
asasine: What about putting the whole PDO in?

asasine (dismissed):
The whole PDO would avoid ambiguity with what each field refers to, and would save us from needing to revisit this again for yet-another-field.

kurtjd (dismissed):

embedded-service/src/type_c/controller.rs
asasine, line 181: What does a "default" port status semantically represent?

RobertZ2011:
Updated the available contract data structures to contain full PD information when available. This includes:
* PDO and RDO for the negotiated contract
* Flags from fixed 5V PDO, these apply to all PDOs but they're only set in this PDO.

type-c-service/src/driver/tps6699x.rs
RobertZ2011: Done, add more details in a separate comment.

embedded-service/src/type_c/controller.rs
RobertZ2011: It represents a port with nothing physically attached. I think there's more that can be done here to wrap this state in an `Option` since these fields only make sense if something is connected. But I'm planning on doing that on main to keep these changes a little more focused.

jerrysxie (approved):

felipebalbi (approved):

tullom (approved):

PR #942, opened 2026-08-07
keyboard-service: migrate to HidDevice

williampMSFT:
Migrate the keyboard-service to leverage the new HidDevice trait, which will let it be used with the new hidi2c-target-service.

As part of this migration, switched to the RunnableService pattern for uniform resource allocation and added a simple test.

This was the last direct dependency on the legacy hid-service crate and its associated embedded-services::hid module, so marked those as deprecated.  We don't have a replacement for the i2c controller piece yet so we can't remove the crate entirely, but everything else should be doable with the hidi2c-target-service crate and HidDevice trait.

This change does not attempt to meaningfully extend the scope of the keyboard interface, which means it still has some notable gaps around support for media keys/backlights/N-key rollover/etc, so this change does not create a new stable interface crate - a skeleton is in the keyboard-service for now and we can revisit making a more complete interface later.


kurtjd (dismissed):

keyboard-service/src/gpio_kb.rs
```diff
@@ -257,11 +191,11 @@ fn has_ghost<const NROWS: usize, const NCOLS: usize>(pressed: &[[bool; NROWS]; N
     // Chose u128 as it's the largest primitive and it's very unlikely a keyboard will have more than 128 rows
     let mut pressed_bits = [0u128; NCOLS];
     let mut count = 0;
-    for (c, row) in pressed.iter().enumerate() {
+    for (row, pressed_bits_col) in pressed.iter().zip(pressed_bits.iter_mut()) {
         for (r, &key) in row.iter().enumerate() {
```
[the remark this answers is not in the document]
jerrysxie: It seems like Copilot is correct that we are getting a column here instead of a row.

keyboard-service/src/relay.rs
```diff
@@ -0,0 +1,202 @@
...
+            HidDevicePowerState::On => KeyboardPowerState::On,
+            HidDevicePowerState::Sleep => KeyboardPowerState::Sleep,
+            HidDevicePowerState::Off => KeyboardPowerState::Off,
+        });
+        Ok(())
+    }
+
+    async fn reset(&mut self) {
+        self.pending_input_report = None;
+        self.subscriber.clear();
```
jerrysxie, line 201: Should `self.last_report` be cleared as well?

keyboard-service/src/gpio_kb.rs
```diff
@@ -298,258 +232,489 @@ fn has_ghost<const NROWS: usize, const NCOLS: usize>(pressed: &[[bool; NROWS]; N
...
+            if !self.shared.is_powered_on() {
+                return ScanOutcome::Unpowered;
+            }
+
+            match self.kb.matrix.get_with_delay(&mut self.kb.delay) {
+                Ok(pressed) => {
+                    // If ghosting detected, report a rollover error.
+                    if self.kb.deghost && has_ghost(&pressed) {
+                        warn!("Key ghosting detected");
+                        return ScanOutcome::Report(KeyboardInputReport::error(ERROR_ROLL_OVER));
```
jerrysxie: There is no delay in the error cases in this function. Would it be possible to busy spin and keep getting errors?

keyboard-service/src/relay.rs
```diff
@@ -0,0 +1,202 @@
...
+                let flags = LedFlags::from_bits_retain(report.data().first().copied().unwrap_or(0));
+                if self.service.set_leds(flags).await.is_err() {
+                    error!("Failed to set keyboard LEDs");
+                }
+            }
+            SetHidReport::Output(_) | SetHidReport::Feature(_) => {}
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
williampMSFT: by design, we're supposed to drop unexpected reports

keyboard-service/src/relay.rs
williampMSFT: I went back and forth on this one a bit, and I'm not totally sure.  If the host commands a reset, that doesn't change which keys are currently physically depressed by the user, and self.last_report is essentially our understanding of which keys are currently depressed.  There might be a case for re-queueing the report so the host immediately gets a report after the reset, but I don't think we were doing that before so I thought I'd leave it?  I think in practice this only matters until the first time the user presses or releases a key after the reset, though, so perhaps it's cleaner to just clear the current state - I can do that.

keyboard-service/src/gpio_kb.rs
williampMSFT: yeah, I think we already had this problem, but added another return disposition so we can handle this kind of failure differently

kurtjd (approved):

jerrysxie (approved):

RobertZ2011 (approved):

PR #946, opened 2026-08-18
type-c-service: Add Pd::reports_unconstrained_power and port unconstrained test from stable to main

### LINKED ISSUE #924: Port `unconstrained` integration test to main from v1 branch

tullom:
Add a new method to get whether a port itself reports unconstrained power to the port::Pd trait and port the type-C unconstrained.rs integration test from stable-v0.1.y to main

`PortStatus::unconstrained_power` is the bit the partner reports in its PDO, so it can't answer whether a port considers itself an unconstrained source. that also depends on the port's `UnconstrainedSink` policy. Add a method to the port's Pd trait that allows itself to report unconstrained power.

The type-C service maps the system unconstrained state from the power policy onto the unconstrained bit each port advertises to its partner.

The test drives port status changes end-to-end and asserts the resulting `set_unconstrained_power` calls on each controller mock.

A second test covers a threshold-flagged sink, where the system is unconstrained due to being over a power threshold.

Assisted-by: GitHub Copilot:claude-opus-5

Resolves #924

### DIFF
--- a/type-c-service/tests/unconstrained.rs
+++ b/type-c-service/tests/unconstrained.rs
@@ -1,0 +1,323 @@
+//! Test the unconstrained power logic of the type-C service
+#![allow(dead_code)]
+#![allow(clippy::indexing_slicing)]
+#![allow(clippy::panic)]
+#![allow(clippy::unwrap_used)]
+
+use embassy_time::{Duration, Timer, with_timeout};
+use embedded_services::named::Named;
+use embedded_usb_pd::{PowerRole, type_c::ConnectionState};
+use log::info;
+use power_policy_interface::{
+    capability::PowerCapability,
+    service::{UnconstrainedState, event::Event as PowerPolicyEvent},
+};
+use type_c_interface::{
+    control::pd::PortStatus,
+    port::event::{PortEvent, PortStatusEventBitfield},
+};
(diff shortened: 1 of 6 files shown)


kurtjd (dismissed):

type-c-service/tests/unconstrained.rs
```diff
@@ -0,0 +1,323 @@
...
+        scenario.detach(1).await;
+        scenario
+            .expect_unconstrained(UnconstrainedState::new(false, 0), [false, false, false])
+            .await;
+    }
+}
+
+/// Test the unconstrained logic when the system is unconstrained but no port reports it.
+///
+/// A sink flagged unconstrained through [`UnconstrainedSink::PowerThresholdMilliwatts`] doesn't
```
RobertZ2011: We should fix this behavior by introducing a function to get the port's reported unconstrained status. The boolean in `PortStatus` should just be what's reported in the PDO. I'm fine if you want to do that as a separate PR.

RobertZ2011 (changes requested):

type-c-service/tests/unconstrained.rs
tullom: Updated this PR and added a new commit adding this functionality.

RobertZ2011 (approved):

jerrysxie (approved):

PR #948, opened 2026-08-21
power-policy-interface: Port disconnect reasons from v0.1

RobertZ2011:
Port changes from v0.1 that introduce an enum for disconnect reasons instead of separate flags. Also refactor existing power policy flag structs to use normal Rust structs instead of bitfield representations.

### DIFF
--- a/power-policy-interface/src/capability.rs
+++ b/power-policy-interface/src/capability.rs
@@ -1,6 +1,4 @@
 //! Power capability definitions and related flags
-use bitfield::bitfield;
-use num_enum::{IntoPrimitive, TryFromPrimitive};
 
 /// Amount of power that a device can provider or consume
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -45,7 +43,7 @@ impl From<PowerCapability> for ConsumerPowerCapability {
     fn from(capability: PowerCapability) -> Self {
         Self {
             capability,
-            flags: ConsumerFlags::none(),
+            flags: ConsumerFlags::default(),
         }
     }
 }
@@ -64,7 +62,7 @@ impl From<PowerCapability> for ProviderPowerCapability {
     fn from(capability: PowerCapability) -> Self {
         Self {
             capability,
-            flags: ProviderFlags::none(),
+            flags: ProviderFlags::default(),
         }
     }
 }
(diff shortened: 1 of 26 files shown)


power-policy-service/src/service/consumer.rs
```diff
@@ -247,11 +255,11 @@ impl<'device, Reg: Registration<'device>, Customization: customization::Customiz
 
     /// Determines and connects the best external power
     ///
-    /// `disconnect_flags` describes the reason for a disconnect and is applied to the
+    /// `disconnect` describes the reason for a disconnect and is applied to the
     /// [`ServiceEvent::ConsumerDisconnected`] event when the current consumer is removed and not
     /// replaced by another one. When switching between consumers the flags are derived from the
     /// switch itself (see [`Self::connect_new_consumer`]).
```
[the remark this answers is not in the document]
RobertZ2011: Updated comment.

type-c-service/src/controller/power.rs
```diff
@@ -141,6 +143,35 @@ impl<
...
+        ) {
+            return Ok(());
+        }
+
+        info!("({}): PD hard reset, tearing down power contract", self.name);
+        if let Err(e) = self.psu_state.disconnect(true) {
+            error!("({}): Error updating PSU state after hard reset: {:?}", self.name, e);
+        }
+        if let Err(e) = self
+            .power_policy_notifier
```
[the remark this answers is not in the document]
RobertZ2011: This is intended. A hard reset automatically disables the sink path with currently supported PD controllers.

kurtjd (dismissed):

type-c-service/src/controller/power.rs
```diff
@@ -141,6 +143,35 @@ impl<
         Ok(())
     }
 
+    /// Tear down the active power contract after a PD hard reset.
+    pub(super) async fn process_hard_reset(&mut self) -> Result<(), PdError> {
+        if !matches!(
+            self.psu_state.psu_state,
+            PsuState::ConnectedConsumer(_) | PsuState::ConnectedProvider(_)
```
jerrysxie, line 155: Does this reset gets ignored when a sink contract is waiting for sink ready which will be in idle? What would happen to that sink contract in progress after this this reset is ignored? Is that contract now invalid?

jerrysxie (commented):
Mixing the removal of bitfields and the porting make the PR tough to review: it was hard to separate the mechanical bitfield removal change versus the logic changes. Thankfully they were 2 different commits.

type-c-service/src/controller/power.rs
RobertZ2011: I added logic to clear the deadline and test coverage. This case is unique because it doesn't necessarily change the power role or connection state which is what clears the deadline in other circumstances like detach or power role swap. Existing tests should already catch this in those other circumstances.

kurtjd (dismissed):

type-c-service/tests/power.rs
```diff
@@ -1049,6 +1105,139 @@ impl Test for TestProviderToConsumerRoleSwap {
...
+            .await
+            .unwrap();
+
+        // We should timeout because the hard reset cancels the sink ready deadline.
+        assert!(matches!(
+            with_timeout(Duration::from_secs(3), port0.event_receiver.wait_event()).await,
+            Err(TimeoutError)
+        ));
+    }
+}
```
jerrysxie, line 1316: I asked the code review agent to poke at the hard-reset recovery case a bit more and it produces 2 failing tests. Does the sequence of events in these test reflect real world sequence of events?

```
/// Test that a consumer can renegotiate the same contract after a PD hard reset when the
/// controller does not emit a sink ready event.
struct TestConsumerRecontractAfterHardReset;

impl Test for TestConsumerRecontractAfterHardReset {
    async fn run<'port, 'ch>(
        &mut self,
        _type_c_receiver: TypeCServiceReceiver<'port, 'ch>,
        power_policy_receiver: PowerPolicyServiceReceiver<'port, 'ch>,
        mut port0: TestPort<'port, 'ch>,
        _port1: TestPort<'port, 'ch>,
        _port2: TestPort<'port, 'ch>,
    ) {
        let connected_status = PortStatus {
            available_sink_contract: Some(POWER_CAPABILITY_5V_1A5),
            connection_state: Some(ConnectionState::Attached),
            power_role: PowerRole::Sink,
            ..Default::default()
        };
        {
            let mut mock0 = port0.mock.lock().await;
            // Queue the initial connection, hard-reset status, same-capability recontract, and
            // timer-driven sink-ready poll. The sink path is enabled for both connections.
            mock0.next_result_get_port_status.push_back(Ok(connected_status));
            mock0.next_result_get_port_status.push_back(Ok(connected_status));
            mock0.next_result_get_port_status.push_back(Ok(connected_status));
            mock0.next_result_get_port_status.push_back(Ok(connected_status));
            mock0.next_result_enable_sink_path.push_back(Ok(()));
            mock0.next_result_enable_sink_path.push_back(Ok(()));
        }

        // Establish the original consumer contract with an explicit sink-ready event.
        let mut port_event = PortStatusEventBitfield::none();
        port_event.set_plug_inserted_or_removed(true);
        port_event.set_new_power_contract_as_consumer(true);
        port_event.set_sink_ready(true);
        port0
            .port
            .lock()
            .await
            .process_event(Event::PortEvent(PortEvent::StatusChanged(port_event)))
            .await
            .unwrap();
        assert!(matches!(
            with_timeout(DEFAULT_PER_CALL_TIMEOUT, power_policy_receiver.receive()).await,
            Ok(PowerPolicyEvent::ConsumerConnected(_, _))
        ));

        // Tear down the active contract while the controller continues to report its capability.
        let mut port_event = PortStatusEventBitfield::none();
        port_event.set_pd_hard_reset(true);
        port0
            .port
            .lock()
            .await
            .process_event(Event::PortEvent(PortEvent::StatusChanged(port_event)))
            .await
            .unwrap();
        assert!(matches!(
            with_timeout(DEFAULT_PER_CALL_TIMEOUT, power_policy_receiver.receive()).await,
            Ok(PowerPolicyEvent::ConsumerDisconnected(
                _,
                DisconnectFlags {
                    reason: Some(DisconnectReason::Reset)
                }
            ))
        ));

        // Announce the same contract without sink ready. Recovery now depends on the software
        // deadline because some controllers do not emit a separate sink-ready event.
        let mut port_event = PortStatusEventBitfield::none();
        port_event.set_new_power_contract_as_consumer(true);
        port0
            .port
            .lock()
            .await
            .process_event(Event::PortEvent(PortEvent::StatusChanged(port_event)))
            .await
            .unwrap();

        assert!(
            port0.shared_state.lock().await.sink_ready_deadline().is_some(),
            "same-capability contract did not arm the sink ready deadline after hard reset"
        );

        // Drive the synthetic sink-ready event and verify that power policy reconnects the port.
        let Ok(event) = with_timeout(Duration::from_secs(3), port0.event_receiver.wait_event()).await else {
            panic!("sink ready deadline did not expire");
        };
        port0.port.lock().await.process_event(event).await.unwrap();

        assert!(matches!(
            with_timeout(DEFAULT_PER_CALL_TIMEOUT, power_policy_receiver.receive()).await,
            Ok(PowerPolicyEvent::ConsumerConnected(_, _))
        ));
    }
}

/// Test that a provider can renegotiate the same contract after a PD hard reset.
struct TestProviderRecontractAfterHardReset;

impl Test for TestProviderRecontractAfterHardReset {
    async fn run<'port, 'ch>(
        &mut self,
        _type_c_receiver: TypeCServiceReceiver<'port, 'ch>,
        power_policy_receiver: PowerPolicyServiceReceiver<'port, 'ch>,
        port0: TestPort<'port, 'ch>,
        _port1: TestPort<'port, 'ch>,
        _port2: TestPort<'port, 'ch>,
    ) {
        let connected_status = PortStatus {
            available_source_contract: Some(POWER_CAPABILITY_5V_1A5),
            connection_state: Some(ConnectionState::Attached),
            power_role: PowerRole::Source,
            ..Default::default()
        };
        {
            let mut mock0 = port0.mock.lock().await;
            // Queue the initial connection, hard-reset status, and same-capability recontract.
            mock0.next_result_get_port_status.push_back(Ok(connected_status));
            mock0.next_result_get_port_status.push_back(Ok(connected_status));
            mock0.next_result_get_port_status.push_back(Ok(connected_status));
        }

        // Establish the original provider contract.
        let mut port_event = PortStatusEventBitfield::none();
        port_event.set_plug_inserted_or_removed(true);
        port_event.set_new_power_contract_as_provider(true);
        port0
            .port
            .lock()
            .await
            .process_event(Event::PortEvent(PortEvent::StatusChanged(port_event)))
            .await
            .unwrap();
        assert!(matches!(
            with_timeout(DEFAULT_PER_CALL_TIMEOUT, power_policy_receiver.receive()).await,
            Ok(PowerPolicyEvent::ProviderConnected(_, _))
        ));

        // Tear down the provider while the controller continues to report its capability.
        let mut port_event = PortStatusEventBitfield::none();
        port_event.set_pd_hard_reset(true);
        port0
            .port
            .lock()
            .await
            .process_event(Event::PortEvent(PortEvent::StatusChanged(port_event)))
            .await
            .unwrap();
        assert!(matches!(
            with_timeout(DEFAULT_PER_CALL_TIMEOUT, power_policy_receiver.receive()).await,
            Ok(PowerPolicyEvent::ProviderDisconnected(_))
        ));

        // Reannounce the same capability and require it to be published as a new contract.
        let mut port_event = PortStatusEventBitfield::none();
        port_event.set_new_power_contract_as_provider(true);
        port0
            .port
            .lock()
            .await
            .process_event(Event::PortEvent(PortEvent::StatusChanged(port_event)))
            .await
            .unwrap();

        assert!(
            matches!(
                with_timeout(DEFAULT_PER_CALL_TIMEOUT, power_policy_receiver.receive()).await,
                Ok(PowerPolicyEvent::ProviderConnected(_, _))
            ),
            "same-capability provider contract was not published after hard reset"
        );
    }
}
```

power-policy-interface/src/capability.rs
```diff
@@ -80,276 +78,73 @@ pub enum PowerCapabilityFlags {
...
-    /// Switching
-    ///
-    /// When set this flag indicates that the service is switching to a different PSU.
-    pub bool, switching, set_switching: 1;
-}
-
-/// Type safe wrapper for consumer disconnect flags
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum DisconnectReason {
```
jerrysxie, line 122: Just for clarification, the v0.1 version has a NoLongerCapable variant, and that is not needed anymore?

```
pub enum DisconnectReason {
    /// The device is no longer capable of providing or consuming power,
    /// no further information is available
    NoLongerCapable,
    /// The device has been physically detached
    Detached,
    /// Switching to a different PSU
    Switching,
    /// Renegotiation triggered by device
    AutoRenegotiation,
    /// Renegotiation triggered by code
    ManualRenegotiation,
    /// The device has changed its role
    RoleSwap,
    /// The device experienced a reset
    Reset,
}
```

tullom (dismissed):

type-c-service/tests/power.rs
RobertZ2011: The first test isn't applicable because the PD controller issues a sink ready interrupt after a hard reset. I've added a comment to the hard-reset path to note this. I've added logic to clear out the previous contracts during a hard reset to fix the bug exposed by the second test. The various boolean flags are going to be pulled into the contract fields by another PR I have so they don't need to be cleared. A hard reset doesn't affect the connection state so that field doesn't need to be cleared either.

power-policy-interface/src/capability.rs
RobertZ2011: I took it out in favor of using `Option<DisconnectReason>`. `None` fulfills the same role of indicating that the disconnect happened, without giving more specific information.

kurtjd (approved):

type-c-service/src/controller/power.rs
```diff
@@ -145,7 +145,11 @@ impl<
 
     /// Tear down the active power contract after a PD hard reset.
     pub(super) async fn process_hard_reset(&mut self) -> Result<(), PdError> {
+        // The PD controller will issue its own sink ready interrupt after the hard reset, so we clear the deadline here.
         self.shared_state.lock().await.sink_ready_deadline = None;
+        self.status.available_sink_contract = None;
+        self.status.available_source_contract = None;
```
jerrysxie, line 151: This clears out the contacts in `self.status`.

However, following through the logic in `process_port_status_changed()`:

See <-- marker below
```
 async fn process_port_status_changed(
        &mut self,
        status_event: PortStatusEventBitfield,
    ) -> Result<ServicePortEventData, PdError> {
        let new_status = self.controller.lock().await.get_port_status(self.port).await?; // <-- we cached the status here
        debug!("({}) status: {:#?}", self.name, new_status);
        debug!("({}) status events: {:#?}", self.name, status_event);

        if status_event.plug_inserted_or_removed() {
            self.process_plug_event(&new_status).await?;
        }

        if status_event.pd_hard_reset() {
            self.process_hard_reset().await?;  // <-- self.status gets cleared in this call, but what about `new_status`? Can it have the stale contracts and will proceed to be processed after this event? Or does a status with hard-reset flag set will be ensure that `new_status` will not have any contract set.
        }

        // Tear down the previous contract on a power role swap before establishing the new one
        if status_event.power_swap_completed() {
            self.process_power_role_swap(&new_status).await?;
        }

        // Only notify power policy of a contract after Sink Ready event (always after explicit or implicit contract)
        if status_event.sink_ready() {
            self.process_new_consumer_contract(&new_status).await?;
        }

        if new_status.is_connected() && new_status.available_source_contract != self.status.available_source_contract {
            self.process_new_provider_contract(&new_status).await?;
        }

        self.check_sink_ready_timeout(
            &new_status,
            status_event.new_power_contract_as_consumer(),
            status_event.sink_ready(),
        )
        .await?;

        let status_changed = StatusChangedData {
            status_event,
            previous_status: self.status,
            current_status: new_status,
        };
        self.status = new_status;
        if let Err(e) = self
            .port_notifier
            .notify_status_changed(
                status_changed.status_event,
                status_changed.previous_status,
                status_changed.current_status,
            )
            .await
        {
            error!("Failed to send port status type-C event: {:#?}", e);
        }
        Ok(ServicePortEventData::StatusChanged(status_changed))
    }

```

jerrysxie: Chatted with @RobertZ2011 offline:

new_status can contain contracts, but those contracts won't be stale. Immediately after a hard-reset the port returns to a default 5V contract. That's what will appear in new_status, but we won't start charging from that port until we get a sink ready event from the port.


Please add a comment to explain this case.

jerrysxie (approved):

tullom (approved):

PR #949, opened 2026-08-21
type-c-service: Pull `PortStatus` changes from v0.1

RobertZ2011:
Pull in changes from PR #941.

Assisted-by: GitHub Copilot:GPT-5.6 Sol

### DIFF
--- a/type-c-interface/src/control/pd.rs
+++ b/type-c-interface/src/control/pd.rs
@@ -3,21 +3,165 @@
 use embedded_usb_pd::{
     DataRole, PlugOrientation, PowerRole,
     pdinfo::{AltMode, PowerPathStatus},
+    pdo::{self, sink::FrsRequiredCurrent},
     type_c::ConnectionState,
 };
+use power_policy_interface::capability::PowerCapability;
+
+/// Information about the negotiated source contract from PD
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct PdSourceInfo {
+    /// Received 5V sink PDO data from port partner
+    ///
+    /// Contains various flags that aren't present in other PDOs
+    pub rx_fixed_5v_data: Option<pdo::sink::FixedData>,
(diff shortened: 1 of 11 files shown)


type-c-interface/src/port/pd.rs
```diff
@@ -31,7 +31,7 @@ pub trait Pd: Named {
     /// Returns whether this port reports unconstrained power to the system.
     ///
     /// This is the port's own determination and can differ from
-    /// [`PortStatus::unconstrained_power`], which is what the partner reports in its PDO.
+    /// [`PortStatus`], which is what the partner reports in its PDO.
     fn reports_unconstrained_power(&self) -> bool;
```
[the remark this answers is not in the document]
RobertZ2011: This is intentional since there are multiple unconstrained status values.

RobertZ2011:
examples/rt685s-evk failure is expected as the tps6699x crate will need to be updated once these changes land.

tullom (approved):

type-c-interface/src/control/pd.rs
```diff
@@ -3,21 +3,165 @@
...
+
+    /// Returns true if the port partner has unchunked extended messages support
+    pub fn unchunked_extended_messages_support(&self) -> bool {
+        self.pd
+            .map(|pd| pd.rx_fixed_5v_data.unchunked_extended_messages_support)
+            .unwrap_or(false)
+    }
+
+    /// Returns true if the port partner is EPR capable
+    pub fn epr_capable(&self) -> bool {
```
jerrysxie, line 150: Just a clarification, does this mean epr capable or epr active? The old version is epr active.

jerrysxie (approved):
The change to a public struct breaks API, please announce on Zulip.

type-c-interface/src/control/pd.rs
RobertZ2011: This is if the port advertises itself as EPR capable. Checking if the current contract is an EPR contract can be done by seeing if the voltage exceeds 20V.

PR #950, opened 2026-08-21
Add thermal service unit test

jerrysxie:
Add new unit test to thermal service with the assistance of AI, discover 2 quirks in the thermal service behavior that needs some clarifications

| File | Before (line) | After (line) | Δ |
|---|---|---|---|
| `sensor.rs` | 0/138 (0%) | 133/138 (96.4%) | +96.4 |
| `fan.rs` | 0/225 (0%) | 215/225 (95.6%) | +95.6 |
| `utils.rs` | 0/51 (0%) | 51/51 (100%) | +100.0 |
| `lib.rs` | 0/16 (0%) | 0/16 (0%) | 0.0 |
| **TOTAL** | **0/430 (0%)** | **399/430 (92.79%)** | **+92.79** |

### DIFF
--- a/thermal-service/tests/fan.rs
+++ b/thermal-service/tests/fan.rs
@@ -32,6 +32,7 @@ impl embedded_fans_async::Error for TestError {
 struct FanState {
     rpm: u16,
     fail_commands: bool,
+    fail_rpm_reads: bool,
     requested_rpms: Vec<u16>,
     rpm_readings: VecDeque<u16>,
 }
@@ -70,15 +71,14 @@ impl Fan for TestFan {
 
 impl RpmSense for TestFan {
     async fn rpm(&mut self) -> Result<u16, Self::Error> {
-        self.state
-            .lock()
-            .map(|mut state| {
-                if let Some(rpm) = state.rpm_readings.pop_front() {
-                    state.rpm = rpm;
-                }
-                state.rpm
-            })
-            .map_err(|_| TestError)
+        let mut state = self.state.lock().map_err(|_| TestError)?;
+        if state.fail_rpm_reads {
+            return Err(TestError);
+        }
+        if let Some(rpm) = state.rpm_readings.pop_front() {
+            state.rpm = rpm;
+        }
+        Ok(state.rpm)
     }
 }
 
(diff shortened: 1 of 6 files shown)


thermal-service/tests/sensor.rs
```diff
@@ -0,0 +1,630 @@
...
+    assert_eq!(
+        with_timeout(Duration::from_millis(500), service.temperature_immediate())
+            .await
+            .unwrap(),
+        Err(sensor::Error::RetryExhausted)
+    );
+}
+
+#[tokio::test]
+async fn immediate_temperature_ignores_offset() {
```
jerrysxie: @kurtjd Is this the intended behavior?

thermal-service/tests/fan.rs
```diff
@@ -0,0 +1,695 @@
...
+    service.set_rpm(3_250).await.unwrap();
+    let assertion = async {
+        embassy_time::Timer::after_millis(5).await;
+        assert_eq!(driver_state.lock().unwrap().requested_rpms, [3_250]);
+
+        service.enable_auto_control().await.unwrap();
+        loop {
+            let requested_rpms = driver_state.lock().unwrap().requested_rpms.clone();
+            if requested_rpms.len() >= 4 {
+                assert_eq!(requested_rpms, [3_250, 0, 1_500, 6_000]);
```
jerrysxie, line 230: @kurtjd enable auto control will force RPM to 0 before ramping up? Is that intended?

jerrysxie (commented):
@kurtjd Some of these tests are a bit low value because the AI agent was focused on grcov metric. Feel free to suggest removal of any test that you deem as unnecessary.

thermal-service/tests/fan.rs
```diff
@@ -0,0 +1,695 @@
...
+    let driver = TestFan {
+        state: Arc::clone(&driver_state),
+    };
+    let mut resources = Resources::<TestFan, 4>::default();
+    let mut event_senders: [NoopSender; 0] = [];
+    let (service, runner) = Service::new(
+        &mut resources,
+        InitParams {
+            driver,
+            // A ten second update period would let only the first transition through before the
```
RobertZ2011, line 545: This comment doesn't seem to match `update_period` being set to 10.

thermal-service/tests/fan.rs
```diff
@@ -0,0 +1,695 @@
...
+            ..Default::default()
+        })),
+    };
+    let mut resources = Resources::<TestFan, 4>::default();
+    let mut event_senders: [NoopSender; 0] = [];
+    let (service, runner) = Service::new(
+        &mut resources,
+        InitParams {
+            driver,
+            // A ten second sampling period would let only the first reading through before the
```
RobertZ2011, line 594: Same comment as above.

thermal-service/tests/sensor.rs
```diff
@@ -0,0 +1,630 @@
...
+    let driver = ScriptedSensor {
+        readings: VecDeque::from([Ok(20.0), Ok(21.0)]),
+    };
+    let mut resources = Resources::<ScriptedSensor, 4>::default();
+    let mut event_senders: [NoopSender; 0] = [];
+    let (service, runner) = Service::new(
+        &mut resources,
+        InitParams {
+            driver,
+            // A one second sample period would let only the first reading through before the
```
RobertZ2011, line 459: Same comment.

RobertZ2011 (changes requested):

thermal-service/tests/fan.rs
jerrysxie: If setting the update period to 10 second initially to ensure only 1 update happens until we set the update period to 1 ms, then we check whether the requested RPMs has been set at least 2 times to the max RPM.

Because if the update period was not updated and stays 10 seconds, then the request RPM would not have been updated twice. It is a bit of confusing way to ensure that the update period actually updated, open for suggestions.

thermal-service/tests/fan.rs
jerrysxie: These tests follows the same pattern, set a big initial period, then update it to a smaller value. After setting it to a smaller period, it is expecting to see more updates/samples which the initial long period should not allow.

thermal-service/tests/sensor.rs
kurtjd: I think this is an oversight. I don't recall any particular reason why I would have intentionally NOT added the offset to `temperature_immediate`. So likely should fix so `temperature_immediate` returns value with offset applied, then update this test. I can do that in a follow up PR though, so non-blockng.

thermal-service/tests/fan.rs
kurtjd: Yes. I added this behavior because it forces the fan state machine back into a known good point to let the state machine resume normal operation. Otherwise, I was seeing bugs related to the fact that when the fan is put into manual control (e.g. from a manual `set_rpm`), the fan's `state` retains the value at whatever it was at that moment. If we don't reset it, but the user enables auto control again, the fan's `state` is no longer in sync with the fan's RPM so the state machine breaks.

This does potentially add some brief jitter where the fan has to drop to 0 momentarily, so probably worth investigating an alternative approach in the future, but for this PR at least I would say this is expected behavior.

kurtjd (dismissed):

thermal-service/tests/sensor.rs
jerrysxie: No worries, I added a commit to fix the immediate temperature issue.

thermal-service/tests/fan.rs
jerrysxie: Sounds good, we don't need to fix it for now. On some fan, this might create undesirable acoustic. We will just keep this in mind and fix it in the future when we need to optimize for acoustic in practice.

thermal-service/tests/sensor.rs
kurtjd: Thanks!

kurtjd (approved):

williampMSFT (approved):

RobertZ2011 (approved):

PR #951, opened 2026-08-27
test(hidi2c): add unit tests for target service

jerrysxie:
Add unt test for hidi2c target service

| file | before | after | Δ | covered (after) |
|------|--------|-------|-----|-----------------|
| attn_pin_handler.rs | 0.00% | 100.00% | +100.00 | 23 / 23 |
| `device_descriptor.rs` | 0.00% | 62.34% | +62.34 | 96 / 154 |
| error.rs | 0.00% | 25.00% | +25.00 | 3 / 12 |
| `service.rs` | 0.00% | 62.12% | +62.12 | 492 / 792 |
| `test_support.rs` | — | 81.82% | (new, test-only) | 54 / 66 |
| **Total** | **0.00%** | **63.80%** | **+63.80** | 668 / 1047 |

### DIFF
--- a/hidi2c-target-service/src/device_descriptor.rs
+++ b/hidi2c-target-service/src/device_descriptor.rs
@@ -161,3 +161,128 @@ impl DeviceDescriptor {
         })
     }
 }
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
+mod tests {
+    use super::*;
+    use crate::test_support::{descriptor_device, hardware_version_info};
+
+    const IMPLICIT_DESCRIPTOR: &[u8] = &[
+        0x05, 0x01, // Usage Page (Generic Desktop)
+        0x09, 0x02, // Usage (Mouse)
+        0xa1, 0x01, // Collection (Application)
+        0x75, 0x08, // Report Size (8)
+        0x95, 0x01, // Report Count (1)
+        0x81, 0x02, // Input
+        0x91, 0x02, // Output
+        0xb1, 0x02, // Feature
+        0xc0, // End Collection
+    ];
+
+    const EXPLICIT_DESCRIPTOR: &[u8] = &[
+        0x05, 0x01, // Usage Page (Generic Desktop)
+        0x09, 0x02, // Usage (Mouse)
+        0xa1, 0x01, // Collection (Application)
+        0x85, 0x01, // Report ID (1)
+        0x75, 0x08, // Report Size (8)
+        0x95, 0x01, // Report Count (1)
+        0x81, 0x02, // Input
+        0x91, 0x02, // Output
+        0xb1, 0x02, // Feature
+        0xc0, // End Collection
+    ];
+
+    const TWO_BYTE_INPUT_DESCRIPTOR: &[u8] = &[
+        0x05, 0x01, // Usage Page (Generic Desktop)
+        0x09, 0x02, // Usage (Mouse)
+        0xa1, 0x01, // Collection (Application)
+        0x75, 0x08, // Report Size (8)
+        0x95, 0x02, // Report Count (2)
+        0x81, 0x02, // Input
+        0xc0, // End Collection
+    ];
+
+    const TWO_BYTE_OUTPUT_DESCRIPTOR: &[u8] = &[
+        0x05, 0x01, // Usage Page (Generic Desktop)
+        0x09, 0x02, // Usage (Mouse)
+        0xa1, 0x01, // Collection (Application)
+        0x75, 0x08, // Report Size (8)
+        0x95, 0x02, // Report Count (2)
+        0x91, 0x02, // Output
+        0xc0, // End Collection
+    ];
+
+    const TWO_BYTE_FEATURE_DESCRIPTOR: &[u8] = &[
+        0x05, 0x01, // Usage Page (Generic Desktop)
+        0x09, 0x02, // Usage (Mouse)
+        0xa1, 0x01, // Collection (Application)
+        0x75, 0x08, // Report Size (8)
+        0x95, 0x02, // Report Count (2)
+        0xb1, 0x02, // Feature
+        0xc0, // End Collection
+    ];
+
+    #[tokio::test]
+    async fn descriptor_uses_implicit_report_framing() {
+        let descriptor =
+            DeviceDescriptor::new(&descriptor_device(IMPLICIT_DESCRIPTOR), hardware_version_info()).unwrap();
+
+        assert_eq!(
+            descriptor.w_hid_desc_length,
+            core::mem::size_of::<DeviceDescriptor>() as u16
+        );
+        assert_eq!(descriptor.bcd_version, 0x0100);
+        assert_eq!(descriptor.w_report_desc_length, IMPLICIT_DESCRIPTOR.len() as u16);
+        assert_eq!(descriptor.w_max_input_length, 3);
+        assert_eq!(descriptor.w_max_output_length, 3);
+        assert_eq!(descriptor.w_vendor_id, 0x1234);
+        assert_eq!(descriptor.w_product_id, 0x5678);
+        assert_eq!(descriptor.w_version_id, 0x0100);
+    }
+
+    #[tokio::test]
+    async fn descriptor_accounts_for_explicit_report_id() {
+        let descriptor =
+            DeviceDescriptor::new(&descriptor_device(EXPLICIT_DESCRIPTOR), hardware_version_info()).unwrap();
+
+        assert_eq!(descriptor.w_max_input_length, 4);
+        assert_eq!(descriptor.w_max_output_length, 4);
+    }
+
+    #[tokio::test]
+    async fn descriptor_rejects_oversized_input_report() {
+        let result = DeviceDescriptor::new(&descriptor_device(TWO_BYTE_INPUT_DESCRIPTOR), hardware_version_info());
+
+        assert_eq!(
+            result,
+            Err(DeviceDescriptorError::InputReportTooLarge { actual: 2, max: 1 })
+        );
+    }
+
+    #[tokio::test]
+    async fn descriptor_rejects_oversized_output_report() {
+        let result = DeviceDescriptor::new(&descriptor_device(TWO_BYTE_OUTPUT_DESCRIPTOR), hardware_version_info());
+
+        assert_eq!(
+            result,
+            Err(DeviceDescriptorError::OutputReportTooLarge { actual: 2, max: 1 })
+        );
+    }
+
+    #[tokio::test]
+    async fn descriptor_rejects_oversized_feature_report() {
+        let result = DeviceDescriptor::new(&descriptor_device(TWO_BYTE_FEATURE_DESCRIPTOR), hardware_version_info());
+
+        assert_eq!(
+            result,
+            Err(DeviceDescriptorError::FeatureReportTooLarge { actual: 2, max: 1 })
+        );
+    }
+
+    #[tokio::test]
+    async fn vendor_id_rejects_zero() {
+        assert!(VendorId::new(0).is_none());
+        assert_eq!(VendorId::new(1).unwrap().value(), 1);
+    }
+}
--- a/hidi2c-target-service/src/service.rs
+++ b/hidi2c-target-service/src/service.rs
@@ -5,7 +5,7 @@ use embassy_time::{Duration, with_timeout};
 use embedded_mcu_hal::i2c::target::asynch::I2c as I2cTargetAsync;
 use embedded_mcu_hal::i2c::target::{ReadStatus, Request, WriteStatus};
 use embedded_services::relay::hid;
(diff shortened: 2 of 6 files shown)


felipebalbi:
Failing CI.

hidi2c-target-service/src/service.rs
```diff
@@ -764,3 +764,411 @@ impl Default for TimeoutSettings {
...
+
+    impl embedded_hal::digital::OutputPin for NoopPin {
+        fn set_low(&mut self) -> Result<(), Self::Error> {
+            Ok(())
+        }
+
+        fn set_high(&mut self) -> Result<(), Self::Error> {
+            Ok(())
+        }
+    }
```
felipebalbi, line 818: `embedded-hal-mock` is a much better option here.

The concrete problem: `NoopPin::Error = Infallible`, and both `set_low`/`set_high` are no-ops that return `Ok(())`. That makes two things impossible to test:

1. The five `unwrap_or_else(|_| error!(...))` arms in `service.rs` (315, 492, 506, 536, 681) are unreachable. There is no way to verify this code behaves when the pin errors.
2. `reset_asserts_interrupt_and_first_read_acknowledges_completion` asserts `runner.attn_pin.asserted()`, which only reads `AttnPinHandler`'s own bookkeeping. **That test passes even if the GPIO is never driven at all.**

`embedded_hal_mock::eh1::digital::Mock` fixes both: a `Transaction` list plus `.done()` proves the pin was actually driven, and it can be made to return `Err` on demand.

Scoping note so you don't go hunting: `embedded-hal-mock` only covers `embedded_hal::i2c::I2c` (controller mode). It has nothing for `embedded_mcu_hal::i2c::target::asynch::I2c`, so `NoopBus`/`ScriptedBus` still have to be hand-rolled — but they should stop using `Infallible` for the same reason.

Two practical notes: the new dev-dependency will need to clear the `deny` and `vet-dependencies` jobs, and it is worth considering whether a reusable target-mode I²C mock belongs in its own crate alongside `type-c-interface-test-mocks` / `power-policy-interface-test-mocks` rather than inside this crate's `#[cfg(test)]`.

hidi2c-target-service/src/service.rs
felipebalbi, line 657: This predates the PR, but it is the root cause of the failing tests, so it needs settling here. @williampMSFT could you confirm?

For a device with explicit report IDs and an N-byte payload, a spec-conformant SET_REPORT frame has `len = 2 + 1 + N`, and the data-register payload is `[report_id, payload…]` (N+1 bytes).

This code computes `report_size = len - 2 = N + 1`, sets `data_start_index = 1`, and asks for `data.get(1..N+2)` on an N+1-byte slice → `None` → `Err(InvalidSize)`.

So SET_REPORT appears to be unconditionally broken for any device that uses report IDs. It simply never had a test. Am I missing something?

Note that the `Output`-register path at service.rs:438-446 computes `header_size_bytes = 2 + (1 if explicit)` and gets this right, so the two paths disagree with each other.

History worth flagging, because it explains the four red tests:

- `68d69e49` fixed exactly this — subtracted `HID_REPORT_HEADER_SIZE_BYTES + report_id_size`, validated the echoed report ID against the command header, and added `set_report_rejects_mismatched_wire_report_id`.
- `8a308e42` reverted it on the premise that report IDs "are not repeated in the data payload", deleted that test, and rewrote the remaining expectations to match.

Please don't resolve this by adjusting the test byte arrays — that is what produced the current state. Let's settle what the spec says first, then make the code and the tests follow it.

hidi2c-target-service/src/test_support.rs
```diff
@@ -0,0 +1,160 @@
...
+{
+    descriptor: HidReportDescriptor<'static>,
+    pub power_state: Option<HidDevicePowerState>,
+    pub report_id: Option<ReportId>,
+    pub report_data: [u8; 8],
+    pub report_len: usize,
+    pub feature_report: bool,
+    pub reset_count: usize,
+    _phantom: PhantomData<(In, Out, Feat)>,
+}
```
felipebalbi, line 46: The HID mock is shaped so the main path cannot execute.

`has_pending_input_report()` (line 115) always returns `false`, so `service.rs:510-538` — the entire input-report framing block, including the implicit/explicit report-ID branch — is dead in every test in this PR. `wait_for_input_report()` (line 111) is `pending()` forever, so `run()`'s second `select3` arm never fires.

Separately, `process_get_report`, `set_report`, `set_power_state` and `reset` all return `Ok` unconditionally, so `Error::Device(_)` propagation is only ever reachable via the `Reset` opcode that `process_command` synthesises itself. Even `set_report`'s own failure branch (`ok_or(HidError::TriggerReset)`, line 105) is unreachable, since `report_len` is bounded by the caller.

The mock needs to be able to say both "yes, I have a report" and "no, that failed" before any of this becomes testable.

hidi2c-target-service/src/device_descriptor.rs
felipebalbi, line 149: Re `w_report_desc_length: descriptor.as_bytes().len() as u16`: nothing exercises this arithmetic.

`as u16` silently truncates, so a report descriptor of 65536+ bytes advertises a length of 0. No test, no guard.

felipebalbi, line 152: Re `w_max_input_length: input_max as u16 + report_length_header_size` (and the same shape on line 154): nothing exercises this arithmetic either.

Two distinct failure modes. `input_max as u16` truncates, and the `u16 + u16` that follows panics on overflow in debug and wraps in release. A `HidDevice` declaring `InputReportMaxSize = U65535` is enough to reach it. Untested in both directions.

hidi2c-target-service/src/service.rs
felipebalbi, line 519: Re `report.data().len() as u16 + HID_REPORT_HEADER_SIZE_BYTES + ...`: untested, and it has two distinct failure modes.

`as u16` truncates, so a report longer than 65535 bytes produces a nonsense length header. And the `u16 + u16` chain that follows panics on overflow in debug and wraps in release — `report.data().len() == 65535` is enough.

Nothing in the suite can reach it regardless: the mock's `has_pending_input_report()` returns `false`, so this entire block is dead code in every test.

felipebalbi, line 627: Same problem here — `report.data().len() as u16 + HID_REPORT_HEADER_SIZE_BYTES` truncates and can overflow.

This is also the line `8a308e42` rewrote. It previously added `HID_REPORT_ID_SIZE_BYTES` and emitted the report ID in the header for explicit-ID devices, which is what `reply_with_input_report` (510-526) still does and what `device_descriptor.rs:152` still advertises. Those three currently disagree; whichever is right, all three should agree and a test should pin it.

felipebalbi, line 582: This function welds parsing to I/O, and that is the root cause of most of what is wrong with the tests.

Concretely: to assert that `0x23` means "Output report, ID 3" you currently need a `TimeoutBus`, a `NoopPin`, a `RecordingHidDevice`, a tokio runtime, and `critical-section/std` + `embassy-time/std` with a global timer queue. That is roughly 260 lines of mock and three dev-dependencies to check byte decoding.

Compare `HidI2cReportCommandHeader::try_from_command_byte` — a pure `u8 -> Result<Header, ProtocolError>`. Its three tests need no bus, no pin, no device and no runtime, and they are the only cheap tests in this PR.

Extracting the frame parse into something like:

```rust
fn parse(frame: &[u8], framing: ReportFraming) -> Result<SetReportCommand<'_>, ProtocolError>
```

would make the four currently-failing tests sync and mock-free, and would let proptest assert "never panics on arbitrary input" — precisely the property nobody can verify today.

Two type-level consequences worth taking at the same time:

- Splitting `HidI2cReportType` into `GetReportType` / `SetReportType` deletes the `TryFrom<HidI2cReportType> for GetHidReportType` impl (60-70), the `Input => return Err(InvalidReportType)` arm (660-663), and both tests guarding them. Neither case stays representable.
- `report_ids_implicit: bool` drives a header-size calculation at 432, 438, 510 and 650 here, plus `device_descriptor.rs:140` — five hand-written copies of "2, or 3 if explicit", and two of them disagree. That disagreement *is* the SET_REPORT bug above. A `ReportFraming` type owning `header_len()` makes the arithmetic exist once.

Context for the reasoning: https://balbi.sh/posts/making-smaller-things/

felipebalbi (changes requested):
## Blocking: CI is red, and the cause is a reverted fix

`ubuntu / stable / test` fails — 18 passed, 4 failed:

- `set_output_report_forwards_payload` — panics at service.rs:961
- `set_feature_report_accepts_extended_report_id` — panics at service.rs:988
- `get_feature_report_excludes_explicit_report_id_from_payload` — panics at service.rs:1048
- `get_report_rejects_output_report_type` — assertion failure at service.rs:1028

Please don't fix these by adjusting the byte arrays. The history matters:

- `68d69e49` added the tests **and** fixed a real SET_REPORT bug — consumed the echoed report-ID byte, validated it against the command header, subtracted `HID_REPORT_HEADER_SIZE_BYTES + report_id_size`, and added `set_report_rejects_mismatched_wire_report_id`.
- `8a308e42` reverted all of that on the premise that report IDs "are not repeated in the data payload", deleted that test, and rewrote the remaining expectations to match — incorrectly, which is why four are red.

Net production diff against `main` is zero, so the bug is still there. Details in my inline comment on the SET_REPORT length arithmetic; I would like to settle what the spec actually says before any code moves.

Related: the crate now disagrees with itself in three places on GET_REPORT framing. `process_command` emits a bare 2-byte header; `reply_with_input_report` (service.rs:510-526) emits `[len_lo, len_hi, report_id]`; `device_descriptor.rs:152` advertises `w_max_input_length = input_max + 2 + 1`. Whichever is right, all three should agree, and a test should pin it.

Two process asks: a production behaviour change does not belong in a `test(...)` PR, and please cite the spec section rather than an AI paraphrase of it.

## Coverage gaps

No coverage on `process_request`, `process_register_access`, or `process_input_report_read`. These are the same shape as `reply_with_input_report`, which this PR *does* test directly via a real `Runner` — so the harness already exists. `Runner::run()` and `listen_indefinitely` I will grant are awkward, and `Service::reset()` is only meaningful alongside `run()`.

Also untested: invalid opcode byte, invalid register address, `I2cPowerState::try_from` failure, GET_REPORT with Input type, SET_REPORT with Input type, and the `RepeatedStart` skip loop in `listen_for_response` — that last one is *unreachable*, since `ScriptedBus::listen` is hardcoded `pending()`.

More broadly, the failure paths are structurally untestable rather than merely untested. Both bus mocks and the pin mock use `Infallible`, so every `Error::Bus(_)` arm and all five `unwrap_or_else(|_| error!(...))` arms cannot be reached by construction. Inline comments on both.

## The harness is heavier than the code under test

9 of the 22 tests are `#[tokio::test] async fn` with zero `.await` — all six descriptor tests and all three header tests. Those exercise `DeviceDescriptor::new` and `HidI2cReportCommandHeader::try_from_command_byte`, which are already sync and already pure. The small thing was right there; the harness reached past it.

That is the diagnosis for the rest of the file too: roughly 260 lines of mock (test_support.rs plus NoopBus/ScriptedBus/NoopPin) and three new dev-dependencies — tokio, critical-section/std, embassy-time/std — to assert that `0x23` means "Output report, ID 3". The three header tests need none of it, and they are the only cheap tests in the PR. Extracting the frame parsing out of `process_command` makes the four failing tests sync and mock-free, and opens the door to a proptest no-panic property over arbitrary input. Inline comment on that.

Note the new dev-deps will need to clear the `deny` and `vet-dependencies` jobs.

hidi2c-target-service/src/service.rs
williampMSFT: Good catch, I'll send out a fix shortly.  I think I haven't hit this in my time-alarm testing yet because I'm currently working on hooking up the host side of an input report that's one byte smaller than the largest one in the report descriptor, so this is timely :)

There's another unrelated problem in SetReport path that I found while working on the time-alarm stuff that may also impact this PR that I'll put in as part of that too - similar to the fix in 95f81fb9fa8181a9619808b0e8c0e470ea55a723. The spec is a bit ambiguous on whether or not you need to repeat the report ID twice in the command:get|set paths, but it turns out you do for both, which strikes me as a bit odd given the lengths they went to in the same packet to dodge a single byte in the case where the report ID is <=15, but  oh well.

felipebalbi: I have a couple fixes which I'm almost ready to send. But I was going to send it as a PR to @jerrysxie's PR. I don't mind dropping the fixes, though.

williampMSFT: took a look at your PR and it looks like it incidentally fixes the SetReport thing too, so may as well just go with yours.  Thanks!

felipebalbi:
FYI, pretty much all of my comments are fixed in a PR I added against this PR. For those interested, here it is: https://github.com/jerrysxie/embedded-services/pull/1/

PR #952, opened 2026-08-28
type-c-service: Port UCSI test from stable to main

### LINKED ISSUE #923: Port `ucsi` integration test to main from v1 branch

tullom:
Port the type-C `ucsi.rs` integration test from stable-v0.1.y to main.

On main, UCSI commands are executed directly on the service through `Service::process_ucsi_command()` rather than the old global `type_c::external::execute_ucsi_command()`, so the test harness is extended to hand each test a reference to the running service. Commands are driven through the service and connector changes are observed as `EventData::UsciChangeIndicator` broadcasts on the type-C service receiver.

Changes:

- Extend the `Test` trait and `run_test` to pass the type-C service to each test, allowing tests to invoke request/response APIs directly (existing tests updated to match the new signature).
- Add the `ucsi` integration test covering PPM reset, notification enable, and per-port `GetConnectorCapability`, connect, `GetConnectorStatus`, and `AckCcCi` across all three ports.

Resolves #923

### DIFF
--- a/type-c-service/tests/ucsi.rs
+++ b/type-c-service/tests/ucsi.rs
@@ -1,0 +1,346 @@
+//! Integration test for UCSI
+#![allow(dead_code)]
+#![allow(clippy::unwrap_used)]
+#![allow(clippy::expect_used)]
+#![allow(clippy::panic)]
+
+use embassy_time::with_timeout;
+use embedded_usb_pd::type_c::ConnectionState;
+use embedded_usb_pd::ucsi::cci::GlobalCci;
+use embedded_usb_pd::ucsi::lpm::get_connector_capability::{
+    OperationModeFlags, ResponseData as UcsiConnectorCapability,
+};
+use embedded_usb_pd::ucsi::lpm::get_connector_status::{
+    BatteryChargingCapabilityStatus, ConnectedStatus, ConnectorStatusChange,
+};
+use embedded_usb_pd::ucsi::lpm::{self, ResponseData as LpmResponseData};
+use embedded_usb_pd::ucsi::ppm::{
+    self, ack_cc_ci::Ack, get_capability::ResponseData as PpmCapabilities, set_notification_enable::NotificationEnable,
+};
+use embedded_usb_pd::ucsi::{GlobalCommand, ResponseData as UcsiResponseData};
+use embedded_usb_pd::{GlobalPortId, PdError, PowerRole};
+use log::info;
+use type_c_interface::control::pd::PortStatus;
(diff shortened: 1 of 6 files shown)


type-c-service/tests/ucsi.rs
```diff
@@ -0,0 +1,346 @@
...
+    DEFAULT_PER_CALL_TIMEOUT, DEFAULT_TEST_DURATION, PowerPolicyServiceReceiver, Test, TestPort, TypeCServiceMutexType,
+    TypeCServiceReceiver,
+};
+
+mod common;
+
+/// Execute a UCSI command against the service and assert the resulting response fields.
+///
+/// The response type is crate-private, so fields are compared individually rather than as a whole.
+async fn execute_and_assert(
```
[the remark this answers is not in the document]
tullom: Good catch, reworded.

RobertZ2011 (approved):

type-c-service/tests/ucsi.rs
```diff
@@ -0,0 +1,346 @@
...
+            })),
+            true,
+            *GlobalCci::default().set_ack_command(true),
+            Ok(None),
+        )
+        .await;
+
+        test_lpm(service, &type_c_receiver, &port0, GlobalPortId(0)).await;
+        test_lpm(service, &type_c_receiver, &port1, GlobalPortId(1)).await;
+        test_lpm(service, &type_c_receiver, &port2, GlobalPortId(2)).await;
```
jerrysxie, line 310: This is slightly different from the test in stable. In stable, we connected multiple ports before acknowledge the first connector, so we are exercising the path where pending ports has multiple entries. Here it is going through each port sequentially, testing the sequence for each port in isolation.

@RobertZ2011 Does this change loss test coverage for the multiple pending port case?

tullom: Err, im pretty sure the test in stable also goes through each port sequentially, i dont see the multiple pending ports path tested in stable.

https://github.com/OpenDevicePartnership/embedded-services/blob/65f7e22412884b365343286dd8ca9bf6166906a9/type-c-service/tests/ucsi.rs#L299-L301

I dont think its a bad idea to add that test but i think its better off going into a new PR since this PR is just to bring the test up to parity with the stable test.

jerrysxie: Oh, my bad, I forgot to update my stable branch. I was wondering why the tests look so different between the stable and main.

jerrysxie: Now the diff looks much more straight-forward.

jerrysxie (approved):

